PR #7, opened 2023-10-05
feat: Integrations

HazAT:
fixes: https://github.com/getsentry/spotlight/issues/9


package.json
```diff
@@ -6,8 +6,8 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc && vite build && yarn build:sidecar",
-    "build:sidecar": "tsc --module nodenext --moduleResolution nodenext --esModuleInterop false --target esnext src/node/sidecar.ts --outDir dist/",
-    "watch:build": "vite build --watch",
+    "build:sidecar": "tsc --module nodenext --moduleResolution nodenext --esModuleInterop false --target esnext src/node/sidecar.ts --outDir node/",
+    "build:watch": "vite build --watch",
```
dcramer, line 10: fwiw the reason we do these under watch: is because you often want to run "build:*" and putting a diff env-build style in there complicates things in what inevtiably always becomes a monorepo

src/components/Overview.tsx
```diff
@@ -59,7 +59,25 @@ export default function Overview() {
...
+  
+  integrations.forEach((integration) => {
+    if (integration.tabs) {
+      integration.tabs.forEach((tab) => {
+        tabs.push(
+          {
+            name: tab.name,
+            active: activeTab === tab.name,
+            onSelect: () => {
+              setEventId(null);
```
dcramer, line 71: def gotta figure out a better way to handle routing params.

idk if react-router could work _without_ actually persisting the route in the url or something

or we could just pass around route names/params in some home grown shenanigan

HazAT: Just hacking stuff to make it work now

Lms24:
I'm going to merge this PR to iterate with smaller changes later. Based on Daniel's work, I made a few more changes:

* simplified and adjusted the integration API a little. IMO there's no need for categorized hooks like in Astro - let's just just go with a flat integration API with simple hooks (`setup`, `processEvent` for now).
* adjusted typing to be typesafe within an integration. This is not perfect yet and ideally we find a nicer way to distinguish between content types that increases this type safety further.
* naming of things is WIP and definitely TBD. 
* Console integration now "works" for browser console.log. Not pretty and just for dev purposes but it's something...

Still a lot of todos:
* [ ] Convert Sentry tabs to integrations
* [ ] Integrate Spotlight with Astro's toolbar
* [ ] Convert repo into monorepo. To start with I think we need three packages:
  * `@sentry/spotlight` holding the core spotlight components
  * `@sentry/spotlight-astro` with the Astro integration
  * website (the spotlight website)

PR #14, opened 2023-11-01
Initial work on pnpm monorepo

dcramer:
This breaks the project up into four current packages:

- spotlight (the main library)
- sidecar (the node sidecar proxy)
- website
- tsconfig (shared tsconfig)

There are a number of changes I included that are not monorepo specific, such as lint-staged and standardized prettier config including e.g. tailwind plugin.


dcramer:
A few unrelated thoughts while I was doing this:

- Can we rename the project from sentry-spotlight to spotlight?
- Should (can?) we claim `@spotlight` on NPM?
- Some config was hardcoded (e.g. tsc for sidecar) - I fixed a bit but theres probably more, which may related to the current compile errors.

Lms24 (commented):
Nice! I think eventually we want to extract another package from the website that contains the Astro integration for spotlight. But we can/should tackle this later, this is fine for now. 

If we want to go with the `@spotlight` namespace (seems like there's currently no package with this namespace on npm), I propose we go with:

* `@spotlight/core ` for the main package
* `@spotlight/sidecar` for the sidecar
* `@spotlight/astro` for the Astro integration
* (`@spotlight/website` but not really relevant I guess)

Lms24:
I applied a [workaround](https://github.com/pnpm/pnpm/issues/7279#issuecomment-1793428443) to the website package to get prod builds of the website working by setting the `node-linker=hoisted` strategy in `.npmrc`. This fixes the failing build with pnpm (worked fine with yarn/npm). Let's just be aware that this is a blocker for pnpm users wanting to use the Astro SDK. The workaround here is only meant to unblock us in creating a monorepo. We need to find a better solution.

PR #20, opened 2023-11-13
build: Update package names, version & private

mydea:
Move packages under `@spotlightjs/` name, update `private` and `version` fields.


packages/spotlight/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@sentry/spotlight",
-  "private": true,
-  "version": "0.0.1",
+  "name": "@spotlightjs/spotlight",
```
mydea, line 2: Ah, so the idea is this should be `@spotlightjs/core`? 🤔 not 100% sure, this is what we expect users to install in their app, right? or is the idea that generally this will be abstracted away by other (meta) packages?

packages/sidecar/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@sentry/spotlight-sidecar",
-  "private": true,
-  "version": "0.0.1",
+  "name": "@spotlightjs/sidecar",
+  "private": false,
```
Lms24, line 3:
```suggestion
```

l: I think we can remove `private` it we don't set it to `true`

Lms24 (approved):

packages/spotlight/package.json
Lms24: Yes, I was thinking of `@spotlightjs/core` because at least for Astro `@spotlightjs/astro` will depend on it. However, if spotlight is installed in a framework we don't have a dedicated package for, the idea is that users will install this package. I have no strong opinion about core vs. spotlight. If you prefer `@spotlightjs/spotlight` I think it's also file (albeit a little repetitive 😅 )

Lms24: Also open to other suggestions

mydea: let's go with core, it prob. doesn't matter to much anyhow 👍

PR #21, opened 2023-11-14
build: Setup changesets

mydea:
This sets up changesets to be added.

Workflow:

* With each PR that has a user facing change, run `pnpm changeset:add` locally. This will guide you through the process to add a changeset.
* Later, we'll add a GH action to consume the changesets and bump versions etc. accordingly.
* We'll also add a GHA flow that will warn you in PRs when you forgot to add a changeset.


Lms24 (approved):
I'm so looking forward to seeing how this is gonna work out! Let's give it a try!

PR #22, opened 2023-11-14
build: Add Github Action to prepare a publish

mydea:
This is on top of https://github.com/getsentry/spotlight/pull/21

This adds a workflow that can be triggered, which will open a PR with all the changesets consumed & versions bumped accordingly.

For now, merging this PR will not do anything. But in a follow up to this, I'll add another action that is triggered when this PR is merged into main, which will actually run the npm publish automatically.

So the flow would be, overall:

1. Include changesets in each PR
2. When we want to cut a release, trigger the `Prepare Publish` GHA workflow
3. This will create a PR that needs to be approved & merged
4. When the PR is merged, the npm publish will happen automatically

### DIFF
--- a/.github/workflows/prepare-publish.yml
+++ b/.github/workflows/prepare-publish.yml
@@ -1,0 +1,30 @@
+name: Prepare Publish
+
+on:
+  workflow_dispatch:
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  publish:
+    name: Prepare Publish
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


Lms24 (approved):
I have a question: AFAIK, when adding a changeset for a PR, we need to determine the version impact (major, minor, patch). Will this (or the follow up) action determine the correct release value automatically?

mydea:
yeah, so this action will pick up all "pending" changesets, and bump all packages accordingly. So for example, if you have these changesets pending:

* `@spotlightjs/core`: patch, patch, patch
* `@spotlightjs/sidecar`: patch, minor, patch
* `@spotlightjs/astro`: -

It would bump core patch-level, sidecar minor-level and astro not at all.

PR #23, opened 2023-11-14
build: Use github changelog generator

mydea:
This uses the github changelog generator, which is supposed to include commit/pr/user links. Let's see if that works out of the box - it needs a github token, but I guess if we run this on CI we can just use the GITHUB_TOKEN and it should work 🤷


Lms24 (approved):
Interesting, let's try it

PR #24, opened 2023-11-14
build: update build script

mydea:
TBH not sure if this will actually work, but _seems_ more correct to me - this is failing right now..?


mydea:
@Lms24 can I try to run this manually from this branch? I guess this is safe to do...?

Lms24:
@mydea yup sure, go for it!

Lms24 (approved):

mydea:
OK so the build succeeded, although manual publishing does not work due to environment protection rules. but should be good to go 🤞

PR #29, opened 2023-11-14
build: Publish on push to main

mydea:
This PR ensures that if the PRs generated by changeset are merged, we automatically publish to NPM.

I check this in three ways:

* Push to main only
* Push changing a `CHANGELOG.md` file
* The pushed commit has the message `meta(changelog): Update package versions` - this has to be kept in sync with the commit messages generated in the prepare-publish workflow.

Not sure if that is "bullet proof" enough, or if we need/can have more checks there 🤔 I think for now it should suffice, at least, and have very little room for misuse (worst case something that is on main is published "too early", as you can only publish off main).

I added a `NPM_TOKEN` secret to this repository.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,42 @@
+name: Publish
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '**/CHANGELOG.md'
+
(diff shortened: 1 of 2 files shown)


Lms24 (approved):
Nice, looks great! Yeah, let's try this check and revisit if necessary 👍

Issue #38, opened 2023-11-14
Extract and publish @spotlightjs/astro package


Lms24:
We want to extract the Astro integration from the website package and make it a proper [Astro integration package](https://docs.astro.build/en/reference/integrations-reference/#allow-installation-with-astro-add). 

For the package to work properly with the `npx astro add` command, we need to keep a few things in mind:
* The [integration factory function](https://github.com/getsentry/spotlight/blob/0bddae87339d2651915a5019f95c11661a943b7a/packages/website/src/spotlight/index.ts#L10) needs to be the default export of the package
* `keywords` need to be set [according to the integrations guide](https://docs.astro.build/en/reference/publish-to-npm/#integrations-library)
* Other fields mentioned in the guide should also be set
* AFAICT, we technically don't need to transpile to JS and we should be good to simply publish the TS source files. (Astro will bundle the integrations in its build). Let's see if this actually works.   

Note: We can also first simply extract the package and worry about the package.json, keywords, etc later.

PR #40, opened 2023-11-15
feat(astro): Add `@spotlightjs/astro` package

mydea:
This adds a new `@spotlightjs/astro` package, extracting things out from the website.

I tried running it locally and it seemed to work, at least 😅 
While at it I also updated the package.json of all packages to have a description & use the same volta config.

Closes https://github.com/getsentry/spotlight/issues/38

(Note: this is also an example for having multiple changesets!)


packages/astro/src/overlay/index.ts
```diff
@@ -0,0 +1,30 @@
+import type { DevOverlayPlugin } from 'astro';
+
+import * as Spotlight from '@spotlightjs/core';
+
+const sentryLogo = `<svg class="css-lfbo6j e10nushx4" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 50 44" width="400" height="352"
```
mydea, line 5: I simply inlined this because TS complained, and I figured it's not worth it to figure out how to fix it xD

packages/astro/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "preview": "astro preview",
+    "astro": "astro"
+  },
+  "main": "./src/index.tx",
+  "exports": {
+    ".": "./src/index.ts"
+  },
+  "dependencies": {
+    "@spotlightjs/core": "0.0.2",
+    "@spotlightjs/sidecar": "0.0.2",
```
Lms24: shouldn't these two be set to
```suggestion
    "@spotlightjs/core": "workspace:*",
    "@spotlightjs/sidecar": "workspace:*",
```
?

packages/astro/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@spotlightjs/astro",
+  "description": "Add spotlight to your Astro toolbar, for better debugging.",
+  "version": "0.0.0",
+  "type": "module",
+  "keywors": [
+    "astro",
+    "withastro",
+    "astro-component",
+    "analytics"
```
Lms24, line 10: Let's remove the analytics keyword for now. I was tempted to add it in https://github.com/getsentry/sentry-javascript/pull/9193#discussion_r1356768133 but held off from it because the Astro folks said they'd add a better category. I think something like "devtools" would fit better overall. 
```suggestion
```

packages/astro/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "preview": "astro preview",
+    "astro": "astro"
+  },
+  "main": "./src/index.tx",
+  "exports": {
+    ".": "./src/index.ts"
+  },
+  "dependencies": {
+    "@spotlightjs/core": "0.0.3",
+    "@spotlightjs/sidecar": "0.0.2",
```
Lms24, line 25: Shouldn't these be 

```suggestion
    "@spotlightjs/core": "workspace:*",
    "@spotlightjs/sidecar": "workspace:*",
```

?

Lms24 (approved):
Nice, thank you!

packages/astro/package.json
mydea: hmm not sure, as we publish this it should probably have the proper versions? I think pnpm should handle this properly, and changeset should also update these versions whenever we bump something (hopefully - let's see how that works out!)

Lms24: > I think pnpm should handle this properly

Ahh, You mean in the sense that for local development it will take the workspace version instead of downloading the published versions? I [thought it was the other way around](https://pnpm.io/workspaces#publishing-workspace-packages), that when packing the tarball,`workspace:*` will be replaced with the package's version. But from reading [these docs](https://pnpm.io/workspaces#workspace-protocol-workspace) I now get the impression that it works both ways 😅 

So it's probably fine as is - let's give it a try!

Issue #76, opened 2023-11-21
Docs: Document SDK requirements


dcramer:
How does spotlight currently identify if it should be relaying to the proxy? Does it just try no matter what? Do I need specific SDK versions?

PR #92, opened 2023-11-21
feat: Add Sidecar `connect` function

HazAT:
To hook into Sentry's Node SDK this is what you have to do:

```js
import * as Sentry from '@sentry/node';
import { connect } from '@spotlightjs/sidecar/connect'

Sentry.init({debug: true});
connect(Sentry.getCurrentHub()); // After Sentry init, call connect and pass the hub
...
Sentry.captureMessage('Something went wrong');
```

ref: https://github.com/getsentry/spotlight/issues/76

### DIFF
--- a/packages/sidecar/package.json
+++ b/packages/sidecar/package.json
@@ -25,6 +25,9 @@
     "./vite-plugin": {
       "import": "./src/vite-plugin.js"
     },
+    "./connect": {
+      "import": "./src/connect.js"
+    },
     "./run": {
       "import": "./src/run.js"
     }
--- a/packages/sidecar/src/connect.js
+++ b/packages/sidecar/src/connect.js
@@ -1,0 +1,34 @@
+function serializeEnvelope(envelope) {
+  const [envHeaders, items] = envelope;
+
+  // Initially we construct our envelope as a string and only convert to binary chunks if we encounter binary data
+  const parts = [];
+  parts.push(JSON.stringify(envHeaders));
+
+  for (const item of items) {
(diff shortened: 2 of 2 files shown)


dcramer:
Is there any reason we cant automatically do this somehow?

dcramer:
Also lets open a ticket on the core JS repo to make sure whatever we do here becomes a native API

HazAT:
Only to avoid the SDK even knowing about Spotlight/Sidecar 
```
fetch("http://localhost:8969/stream", {
      method: "POST",
      body: serializeEnvelope(envelope),
      headers: {
        "Content-Type": "application/x-sentry-envelope",
      },
      mode: "cors",
    }).catch((err) => {
      console.error('[Spotlight]', err);
    });
```

HazAT:
The advantage of an explicit `connect` function from the sidecar is that it's stays in sync with the actual sidecar
If we add this code into the SDKs you need a specific version + if we change the URL / Port, it gets outdated.

I think initially this is fine until this all is a bit more robust.

packages/sidecar/package.json
```diff
@@ -25,6 +25,9 @@
     "./vite-plugin": {
       "import": "./src/vite-plugin.js"
     },
+    "./connect": {
```
dcramer, line 28: is there no other sane way to define these main exports e.g. in like an index.ts or something?

dcramer (approved):
did not test

dcramer:
We should reconsider this and put a spotlight: true/false into Sentry.init() in SDKs imo.

Issue #90, opened 2023-11-21
Trace rendering failing


dcramer:
Trace rendering fails with "Cannot create a faux parent" when incomplete traces.

https://github.com/dcramer/peated/pull/107

We should also consider adding an empty trace dialog, capture if there were errors, and direct people to open a GitHub issue.

dcramer:
Likely duplicate of #61 lets keep open til we confirm.

dcramer:
Data for reproduction:

```
{
    "sortedSpans": [
        {
            "op": "pageload",
            "parent_span_id": "a57d7764a3fba599",
            "span_id": "a5a88679eff265f1",
            "tags": {
                "routing.instrumentation": "remix-router",
                "effectiveConnectionType": "4g",
                "deviceMemory": "8 GB",
                "hardwareConcurrency": "16"
            },
            "trace_id": "49a7c4e270c1439586117393348d57a5",
            "origin": "auto.pageload.remix",
            "start_timestamp": 1700593973693.6,
            "timestamp": 1700593974058.3,
            "description": "routes/_index",
            "transaction": {
                "contexts": {
                    "trace": {
                        "op": "pageload",
                        "parent_span_id": "a57d7764a3fba599",
                        "span_id": "a5a88679eff265f1",
                        "tags": {
                            "routing.instrumentation": "remix-router",
                            "effectiveConnectionType": "4g",
                            "deviceMemory": "8 GB",
                            "hardwareConcurrency": "16"
                        },
                        "trace_id": "49a7c4e270c1439586117393348d57a5",
                        "origin": "auto.pageload.remix"
                    }
                },
                "spans": [
                    {
                        "description": "unloadEvent",
                        "op": "browser",
                        "parent_span_id": "a5a88679eff265f1",
                        "span_id": "8cea8e11530044f9",
                        "start_timestamp": 1700593973.8448,
                        "timestamp": 1700593973.8448,
                        "trace_id": "49a7c4e270c1439586117393348d57a5",
                        "origin": "auto.browser.browser.metrics"
                    },
                    {
                        "description": "domContentLoadedEvent",
                        "op": "browser",
                        "parent_span_id": "a5a88679eff265f1",
                        "span_id": "80b6c67a71b799c3",
                        "start_timestamp": 1700593973.9361,
                        "timestamp": 1700593973.9361,
                        "trace_id": "49a7c4e270c1439586117393348d57a5",
                        "origin": "auto.browser.browser.metrics"
                    },
                    {
[...]

Issue #61, opened 2023-11-20
Distributed Traces don't show all spans


Lms24:
This trace is a connected BE->FE trace but I only see spans from the BE. It says though that there are 48 spans. Probably something going wrong with us creating the span tree:

<img width="1290" alt="image" src="https://github.com/getsentry/spotlight/assets/8420481/1969565b-e394-415f-8836-28f9c04864e0">

Lms24:
fyi, I'm currently debugging pageload<>SSR spans in Astro where this still happens. Gonna assign myself

PR #93, opened 2023-11-21
fix: Handle missing root transactions

dcramer:
There are (currently unknown) situations where a span batch may not be able to identify a root transaction. That is, all spans available in the transaction group are siblings.

This may not be the ideal fix, and we'll revisit shortly. This is simply to unblock UI errors.

Refs GH-90, GH-61


dcramer:
Test currently shows behavior we do not desire - its creating _two_ orphan parents.

The output should basically be:

- inject one faux parent and call it an orphan?
- two correct spans become children of it

dcramer:
If there are fixes we'll follow up w/ new PR. Merging so I can continue to debug w/ all branches combined.

PR #146, opened 2023-11-28
Ruby SDK version requirement: update to 5.15

natikgadzhi:
## Summary

Got to love PRs with one character change! This updates required sentry-ruby version from 5.14 to 5.15.

5.14 shipped today without Spotlight support. So 5.15 is optimistic, assuming we merge https://github.com/getsentry/sentry-ruby/pull/2175 soon enough and ship it.

## Review

Welp, saying Spotlight requires a version of SDK that haven't shipped yet is slightly suspicious, but I think it's better than saying things work on 5.14. Plus, there's not a whole lot of people on that website just yet, right? 

/cc @HazAT


Lms24 (approved):
Thanks!

PR #149, opened 2023-11-28
Fixed failing of test cases.

Shubhdeep12:
fixes: #150 

Added aliases for paths in the 'vitest.config.ts' as well, as this was causing test cases to fail.


Lms24 (approved):
Thanks a lot! I see this is currently blocking all PRs. Will merge this.

PR #153, opened 2023-11-28
feat: Use Esc for back navigation

### LINKED ISSUE #51: Figure out good back navigation UX
Currently, going back (e.g. from an error detail to the errors list) is broken. It used to be possible by pressing [ESC], however it had weird interchanges with pressing ESC on top level pages (e.g. errors/traces list). I think we should re-think how navigating back is done and most obvious for users. 

Ideas:
* Add a back button
* Intercept browser back button presses/history changes to navigate back. This might be tricky though with the React Router `MemoryRouter` 
* Bring back the [ESC] back functionality (is this obvious for users though?)
* Avoid explicit back logic but reset tabs when clicking onto the tabs?

HazAT:
Fixes #51

### DIFF
--- a/packages/overlay/src/components/Tabs.tsx
+++ b/packages/overlay/src/components/Tabs.tsx
@@ -1,6 +1,8 @@
-import { NavLink, useNavigate } from 'react-router-dom';
+import { NavLink, useLocation, useNavigate } from 'react-router-dom';
 import { type IntegrationTab } from '~/integrations/integration';
 import classNames from '../lib/classNames';
+import { getSpotlightEventTarget } from '../lib/eventTarget';
+import useKeyPress from '../lib/useKeyPress';
 
 export type Props = {
   /**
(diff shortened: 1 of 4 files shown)


Lms24 (approved):
Nice! I think this is a simple and good solution for now and definitely much better than what we had (didn't have) before. 

Once concern but maybe it's not even an issue: If you navigate through a lot of spans for example and then start going back to close spotlight, you'll have to press ESC quite a few times. Maybe we need to add some kind of level-back navigation for later but otoh that might also be counter intuitive. Let's ship this!

packages/overlay/src/components/Tabs.tsx
```diff
@@ -50,6 +63,7 @@ export default function Tabs({ tabs, nested }: Props) {
             <NavLink
               to={`${nested ? '' : '/'}${tab.id}`}
               key={tab.id}
+              replace={true}
```
Lms24, line 66: Oh I see this actually solves the problem I mentioned in my review, right?
So we just need to apply it to the respective places

HazAT: Yeah, might be a bit more delicate there - tried it quickly and didn't work ootb

We could also consider hitting `ESC` in quick succession closes Spotlight.

dcramer:
Thought: could it navigate Up vs Back?

e.g. /traces/123/spans/1234/ => /traces/123/ => / => close

HazAT:
@dcramer That's exactly what it does - or maybe I misunderstanding

packages/overlay/src/components/Tabs.tsx
Lms24: Instead of this, maybe we add another key combination for dedicated open/close functionality? #158

PR #155, opened 2023-11-28
ref(astro): Use Spotlight integration on server-side instead of custom code

Lms24:
This PR refactors hooking into server-side Sentry in astro apps. In contrast to regular Node apps, users don't need to set `spotlight: true` in their Astro SDK because we can simply add the integration in the Astro Spotlight integration. This removes the custom hack we had before that accesses the Sentry client via the global object.

I'll open a follow-up PR to pass a custom sidecar URL to the server snippet

### DIFF
--- a/demos/astro-playground/astro.config.mjs
+++ b/demos/astro-playground/astro.config.mjs
@@ -27,6 +27,9 @@ export default defineConfig({
     react({ include: ['**/react/*'] }),
     sentry({
       debug: true,
+      sourceMapsUploadOptions: {
+        enabled: false, // tmp deactivate until version after 7.83.0
+      },
     }),
     spotlight(),
   ],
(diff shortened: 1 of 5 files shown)


demos/astro-playground/astro.config.mjs
```diff
@@ -27,6 +27,9 @@ export default defineConfig({
     react({ include: ['**/react/*'] }),
     sentry({
       debug: true,
+      sourceMapsUploadOptions: {
+        enabled: false, // tmp deactivate until version after 7.83.0
+      },
```
Lms24, line 32: fixed by https://github.com/getsentry/sentry-javascript/pull/9688 but we need to bump the SDK here, too. 

Also reminder to adjust the minimum required SDK version before the launch

PR #162, opened 2023-11-28
feat(astro): Toggle notification dot when receving a severe event

Lms24:
- feat(overlay): Add `onOpen` and `onSevereEvent` callbacks
- changeset
- feat(astro): Toggle notification dot when receving a severe event


HazAT (approved):
💯

Issue #144, opened 2023-11-28
Tree shaking Spotlight out of prod


dcramer:
Its critical that the way people install Spotlight makes it only available in development bundles.

dcramer:
I believe we're good at this point at least in my really non scientific test

PR #164, opened 2023-11-28
Improve setup docs to cover Vite/always show dev check

dcramer:
Let's try to avoid footguns to the best of our ability.

Refs GH-144


dcramer:
image

PR #179, opened 2023-11-30
POC for synchronous error rendering

dcramer:
High level:

- Expose `Spotlight.trigger(eventName, eventPayload)`
- Create "sentry:showError" event (expection is integrations namespace events).
- Trigger "sentry:showError" with eventId and event to synchronously render current errors.

### DIFF
--- a/packages/overlay/src/components/Overview.tsx
+++ b/packages/overlay/src/components/Overview.tsx
@@ -46,8 +46,6 @@ export default function Overview({
     setTriggerButtonCount(notificationCountSum);
   }, [notificationCountSum, setTriggerButtonCount]);
 
-  const initialTab = tabs.length ? `/${tabs[0].id}` : '/no-tabs';
-
   return (
     <>
       <Tabs tabs={tabs} setOpen={setOpen} />


demos/astro-playground/sentry.client.config.ts
```diff
@@ -1,7 +1,21 @@
...
+import { getCurrentHub, type Event } from '@sentry/astro';
 
 Sentry.init({
   debug: true,
   integrations: [new Sentry.BrowserTracing(), new Sentry.Replay()],
   tracesSampleRate: 1.0,
 });
+
+if (process.env.NODE_ENV === 'development') {
+  import('@spotlightjs/astro').then(Spotlight => {
```
dcramer, line 11: so we could actually add localStorage settings into spotlight and make this a default behavior for _any_ unhandled error coming in too.

can also keep this behavior (i think we NEED it) and continue w/ a setting to toggle it

dcramer:
To ship this theres a few things, some imo required, some could be second pass:

1. Clean up the router implementation. We probably need to be a bit smarter about where all the overlay code lives vs where the rendering layer starts. We should do _all_ setup before rendering. This is still a bit all over the place, and may need abstracted via  React Context helper.

2. It'd be nice to have type safety on Spotlight.trigger.

3. We need to figure out a path to make this little hack to inject the error smarter. On server-side this is actually pretty dang easy, but we'll still need to be framework aware (ughhhhh). I do THINK the default of "pop the modal on any unhandled error" is ok but I'm 100% sure thisll be annoying with SOA...

dcramer:
> We need to figure out a path to make this little hack to inject the error smarter. On server-side this is actually pretty dang easy, but we'll still need to be framework aware (ughhhhh). I do THINK the default of "pop the modal on any unhandled error" is ok but I'm 100% sure thisll be annoying with SOA...

I guess the tricky bit here, and what we'd probably want to do...

```
if spotlight: <inject spotlight javascript code on the Django 500.html page>
```

packages/overlay/src/index.tsx
```diff
@@ -121,16 +142,32 @@ export async function init({
...
-      defaultEventId={defaultEventId}
-      showTriggerButton={showTriggerButton}
-      sidecar={sidecar}
-      anchor={anchor}
-    />,
+    <MemoryRouter initialEntries={[initialTab]}>
+      <App
+        integrations={initializedIntegrations}
+        fullScreen={fullScreen}
+        defaultEventId={defaultEventId}
```
Lms24, line 165: I think we can remove this option then, right? (already marked as a removal candidate in the JSDoc)

Lms24 (commented):
I think the big problem here is that we can't reliably detect what error kind of error should open Spotlight vs. which one shouldn't.

Checking `handled` won't do the trick because JS SDKs always set `handled: false` unless users manually called `captureException`.

packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -21,6 +23,21 @@ export default function sentryIntegration(options?: SentryIntegrationOptions) {
...
       addSpotlightIntegrationToSentry(options);
+
+      const spotlightEventTarget = getSpotlightEventTarget();
+
+      const onRenderError = (e: CustomEvent) => {
+        log('Sentry Event', e.detail.event_id);
+        if (e.detail.event) sentryDataCache.pushEvent(e.detail.event);
+        // TODO: handle async
+        openSpotlight(`/errors/${e.detail.eventId}`);
+      };
```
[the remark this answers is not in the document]
dcramer: That could work though we likely need a larger context manager. I dont think we should abstract __too__ much away from React

PR #183, opened 2023-11-30
fix: captureMessage Events

### LINKED ISSUE #142: Overlay crashes when I click on all spans in _some_ traces, but not others
### Environment

- `@spotlightjs/overlay@0.0.16`
- Rails 7.0.3 🤦🏼
- Sidecar 0.0.11

### Steps to Reproduce

1. Opening a demo app (should I push this up?)
2. Opening Sidecar Overlay
3. Clicking on a trace
4. Clicking into a specific span

[Here's the video of what I'm doing](https://www.loom.com/share/1b3545cc68e2420eb91be9aca2d1d1f6?sid=b0934a04-6808-47e5-ac6c-27c839c3b5c3)

### Expected Result

I expect a popover panel with the span data.

### Actual Result

The Overlay app crashes and is removed from DOM.

HazAT:
Fixes #142

image


Shubhdeep12:
Hi @HazAT a query - was mentioned in the issue as well. Is it fine to include messages in Errors? 

I mean either there should only be error events under "Errors" or the label of the tag should be "Events"

HazAT:
We keep it `Errors` for now since this is it's main use case - I don't want to make another tab for Events/Errors etc.
We'll iterate on that.

PR #185, opened 2023-11-30
fix(overlay): Account for inverted BE->FE traces

Lms24:
Our Fullstack SDKs (Astro, Sveltekit, Next, etc) have the (weird but somewhat logical) behaviour of creating traces for SSR page loads that create intuitively inverted traces: The backend transaction is the root transaction and the frontend txn it's child:
image

This appears weird because the browser txn starts earlier (as the browser makes the request to the server), so it should be the root transaction. Ideally yes, but our browser SDK only initializes after the server responds with the pageload response. At this time, we init the browser SDK, pick up the server's <meta> tag and therefore assign the parent span id to the server transaction.

In Spotlight, span tree creation and grouping algorithm didn't account for this but we assumed that sorting the list of spans by start timestamps would bring transaction and trace grouping into the correct order. However, with the scenario described above, we accidentally overwrote an already created span tree for the browser transaction instead of appending it to the server transaction's span.

This PR fixes this bug by
1. Accounting for the fact that not all parent spans might have been visited before we get to a child span
2. Only resetting the child array of a span if it is in fact empty

Runtime might be slightly impacted by this but I think this is fine. 

closes https://github.com/getsentry/spotlight/issues/61


packages/overlay/src/integrations/sentry/utils/traces.ts
```diff
@@ -12,17 +12,21 @@ export function groupSpans(spans: Span[]) {
...
+    log(`parent of span (${span.span_id}, ${span.op}, p: ${span.parent_span_id}) is ${parent?.span_id}`);
+    span.children = span.children || [];
     if (parent) {
-      if (!parent.children) parent.children = [];
+      if (!parent.children) {
+        parent.children = [];
+      }
       parent.children.push(span);
     } else if (span.parent_span_id) {
       const parentParent = sortedSpans.find(s => !s.parent_span_id);
```
Lms24, line 23: I think this might be off, too but not sure. Didn't cause problems for this case

Issue #184, opened 2023-11-30
Run Sidecar on custom port


Shubhdeep12:
Hi, 

Feature Request - I believe it would be beneficial to provide users with the option to run the sidecar on different port(if needed).
Usecase - One potential use case for this feature is when a user wants to run Spotlight for multiple applications at the same time.
Solution - one possible solution is that `npx @spotlightjs/spotlight optionalCustomPORT`

If this is a valid request, please confirm, I'll push a commit and raise a PR for this.
Thanks

HazAT:
Yes, we can just read `process.args` and add an optional argument to `setupSidecar`

Shubhdeep12:
Yes, i was thinking about the same. I'll create a PR in sometime. 
Thanks for confirming.

PR #189, opened 2023-11-30
feat(sidecar): Sidecar support custom port

Shubhdeep12:
fixes: #184


packages/sidecar/src/main.ts
```diff
@@ -205,11 +205,25 @@ function log(...args: any[]) {
 
 let serverInstance: Server;
 
-export function setupSidecar(): void {
+const isValidPort = (value: string) => {
+  const portNumber = parseInt(value, 10);
+  return /^\d+$/.test(value) && portNumber > 0 && portNumber <= 65535;
+};
+
+export function setupSidecar(port: string): void {
```
Lms24, line 213: Can we make this parameter optional and accept string | number? I get the string use case for CLI but we also call this in some frameworks (sveltekit, docs coming soon) and it feels unintuitive to pass the port as a string.

Shubhdeep12: Yup, makes sense. fixing it

Lms24 (approved):
Thank you!

Issue #178, opened 2023-11-30
Breadcrumbs need improved


dcramer:
![image](https://github.com/getsentry/spotlight/assets/23610/1f029897-61b9-48fa-adc5-f46e3405423e)

dcramer:
Mostly just needs to look good and functional. Eg it’s using flex box and ends up misaligned when it’s more or less tabular data. CSS grid prob would fix

PR #194, opened 2023-12-01
fix(breadcrumbs): breadcrumbs ui improvement and minor bug fix

Shubhdeep12:
image

- some UI improvement
- bug fix - if statement for checking empty screen in breadcrumbs was missing a condition

Fixes: #178

### DIFF
--- a/packages/overlay/src/integrations/sentry/components/EventBreadcrumbs.tsx
+++ b/packages/overlay/src/integrations/sentry/components/EventBreadcrumbs.tsx
@@ -1,3 +1,5 @@
+import { Fragment } from 'react';
+import classNames from '~/lib/classNames';
 import { Breadcrumb, SentryEvent } from '../types';
 import Time from './Time';
 
@@ -21,9 +23,9 @@ function extractBreadcrumbs(event: SentryEvent): Breadcrumb[] | undefined {
 
 export default function EventBreadcrumbs({ event }: { event: SentryEvent }) {
   const breadcrumbs = extractBreadcrumbs(event);
-  if (!breadcrumbs) {
+  if (!breadcrumbs || (Array.isArray(breadcrumbs) && breadcrumbs.length === 0)) {
     return (
-      <div className="space-y-4 px-6">
+      <div className="space-y-4">
         <div className="text-primary-300">
           No breadcrumbs available for this event. Try adding some to make debugging easier.
         </div>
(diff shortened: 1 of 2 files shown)


packages/overlay/src/integrations/sentry/components/EventBreadcrumbs.tsx
```diff
@@ -21,9 +23,9 @@ function extractBreadcrumbs(event: SentryEvent): Breadcrumb[] | undefined {
 
 export default function EventBreadcrumbs({ event }: { event: SentryEvent }) {
   const breadcrumbs = extractBreadcrumbs(event);
-  if (!breadcrumbs) {
+  if (!breadcrumbs || (Array.isArray(breadcrumbs) && breadcrumbs.length === 0)) {
```
Lms24, line 26: I think we don't need the isArray check as `breadcrumbs` can either be undefined or an array:

```suggestion
  if (!breadcrumbs || !breadcrumbs.length)) {
```

Lms24: Scratch that, I think the type signature of `extractBreadcrumbs` is outdated. It already only returns `Breadcrumb[]` so let's change the signature and simplify this check.

Lms24 (commented):
Thanks! I think this already looks much better! I have one more suggestion for a quick improvement:

Can we display the breadcrumb category above the timestamp and make it visually more distinctive? E.g. make the category bolder and a bit larger (or the time smaller). I'll leave the details up to you ;)

packages/overlay/src/integrations/sentry/components/EventBreadcrumbs.tsx
Shubhdeep12: @Lms24 I think the suggestion you made is needed.
it was a miss from my end, extractbreadcrumbs always return array, so I think there's no need to check whether breadcrumbs is Arrray.

thanks

Shubhdeep12:
> Thanks! I think this already looks much better! I have one more suggestion for a quick improvement:
> 
> Can we display the breadcrumb category above the timestamp and make it visually more distinctive? E.g. make the category bolder and a bit larger (or the time smaller). I'll leave the details up to you ;)

Sure.

Lms24 (approved):
Awesome, thanks!

Issue #158, opened 2023-11-28
(low prio): Open with keypress


Lms24:
Something like Cmd+F12 (similar to dev tools) could open/close spotlight.

With ESC now being primarily used to navigate back, a second key (combination) might be handy.

Shubhdeep12:
How about Ctrl + Shift + S

S stands for Sentry and Spotlight 😅

Lms24:
Sure that also works 😅

Shubhdeep12:
So should we add this? Let me know, i'll push a commit for this.

Lms24:
Yes, feel free to open a PR! If you do it, please verify that the integration with the Astro dev overlay still works. Testing with the Astro playground demo is probably helpful.

Shubhdeep12:
Hi @Lms24 looks like we completed this one and we are using ctrl/cmd + F12 for this.

Should we close this issue now?

Lms24:
Huh right, not sure why this is still open, closing!

PR #196, opened 2023-12-01
feat(overlay): open/close overlay on keypress

Shubhdeep12:
#158 

Added an event listener for opening/closing overlay using `Ctrl + Shift + s`
Also made this to astro dev as well.

### DIFF
--- a/packages/overlay/src/App.tsx
+++ b/packages/overlay/src/App.tsx
@@ -59,10 +59,18 @@ export default function App({
       setOpen(false);
     };
 
+    const handleKeyPress = (event: KeyboardEvent) => {
+      if (event && event.ctrlKey && event.shiftKey && event.key.toLowerCase() === 's') {
+        setOpen(prevValue => !prevValue);
+      }
+    };
+
+    document.addEventListener('keydown', handleKeyPress);
     spotlightEventTarget.addEventListener('open', onOpen);
     spotlightEventTarget.addEventListener('close', onClose);
 
     return () => {
+      document.removeEventListener('keydown', handleKeyPress);
       spotlightEventTarget.removeEventListener('open', onOpen);
       spotlightEventTarget.removeEventListener('close', onClose);
     };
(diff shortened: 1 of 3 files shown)


packages/overlay/src/App.tsx
```diff
@@ -59,10 +59,18 @@ export default function App({
       setOpen(false);
     };
 
+    const handleKeyPress = (event: KeyboardEvent) => {
+      if (event && event.ctrlKey && event.shiftKey && event.key.toLowerCase() === 's') {
+        setOpen(prevValue => !prevValue);
+      }
+    };
+
+    document.addEventListener('keydown', handleKeyPress);
```
Lms24, line 68: Is there a specific reason you put this into the `useEffect` block?

Asking because if we can move it to the top level of the `App` component, I'd prefer using the `useKeyPress` hook we already use for the ESC key navigation.

packages/overlay/src/App.tsx
```diff
@@ -59,10 +59,18 @@ export default function App({
       setOpen(false);
     };
 
+    const handleKeyPress = (event: KeyboardEvent) => {
+      if (event && event.ctrlKey && event.shiftKey && event.key.toLowerCase() === 's') {
```
Lms24, line 63: Let's accept the Cmd key on MacOS as well. CTRL on macbooks is a bit unintuitive. Seems like this is safe to use because it [doesn't map to the windows key
](https://stackoverflow.com/a/3922353). I'm fine with both combinations working on MacOS.

```suggestion
      if (event && (event.ctrlKey || event.metaKey) && event.shiftKey && event.key.toLowerCase() === 's') {
```

Lms24 (commented):
Awesome! Gave it a quick test and it seems to work well. I Have two requests though. 

Also, please rebase your branch. I just merged in an ESC key fix and I wanna make sure this doesn't collide after the fix. Thank you!

Lms24:
Sounds good to me!

HazAT (approved):
💯

PR #198, opened 2023-12-01
fix(tabs): Default active detail tab in Events

Shubhdeep12:
Currently in EventDetails 
Default active tab is Details
But we are only utilizing react-router active nodes, due to which Details tab is not shown in active state.

Added condition to check both active node OR active boolean

### DIFF
--- a/.changeset/strange-lions-swim.md
+++ b/.changeset/strange-lions-swim.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/overlay': patch
+---
+
+feat(overlay): Add tags to error event contexts view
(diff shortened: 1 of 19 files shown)


Lms24 (changes requested):
I think manually keeping track of an active tab was a leftover from refactoring spotlight to an integration based architecture. Let's not use locally kept active state but always use the router's active node.

dcramer:
Coincidentally just was mentioning this in Slack ;)

We def need to refactor these to use router and I broke this when I was making other changes

Shubhdeep12:
Hi @dcramer @Lms24 
just pushed a fix that will use react router active node only.

HazAT:
Thx

PR #204, opened 2023-12-03
docs(sidecar): Sidecar optional port config docs added and extension recommendation for viewing MDX files added

Shubhdeep12:
Updated the docs for Sidecar about the optional config ability.
And, added an extension recommendation in the code for visualizing the mdx file for contributors.


HazAT (changes requested):
Honestly, I am not sure if we need this - anyone needing this will find it and it just makes the default docs "more complicated"

Shubhdeep12:
Ok, thanks for confirming @HazAT. I think it's better to wait then, and based on the user's feedback in the future, we can make a decision to include this.

Closing the PR.

PR #210, opened 2023-12-04
Fix: test files restructure

Shubhdeep12:
It would be better to have all test files under __tests__ folder.


Lms24 (commented):
@Shubhdeep12 can we adjust this PR to move tests under a `overlay/tests` directory? And Within this directory, create structures that resemble the respective source file path.

This is how we work in the main sentry-javascript SDK repo and if we make changes, I wanna align them to this model.

Lms24 (commented):
Sorry for the extra work, I think I wasn't clear in my first review:

Can we move the test files within `overlay/test` to a sub directory that matches the sub directory of the source file's location? 

So for example: the test file for `src/integrations/sentry/data/sentryDataCache.ts` will be in `test/integrations/sentry/data/sentryDataCache.ts`.

Shubhdeep12:
> Sorry for the extra work, I think I wasn't clear in my first review:
> 
> Can we move the test files within `overlay/test` to a sub directory that matches the sub directory of the source file's location?
> 
> So for example: the test file for `src/integrations/sentry/data/sentryDataCache.ts` will be in `test/integrations/sentry/data/sentryDataCache.test.ts`.
> 
> EDIT: naming error in the test file path

np, I got it now. let me make the changes

PR #212, opened 2023-12-04
ref(astro): Check for Astro 4 `config.devToolbar` in Spotlight integration

### LINKED ISSUE #211: Support Astro's new `devToolbar` option
In the newest Astro 4.0 beta (4.0.0-beta.4), the dev toolbar is now on by default and it was promoted from the experimental properties. We should update our package accordingly to take both ways of checking for the toolbar into account.

image

Lms24:
Incorporate changes from https://github.com/withastro/astro/pull/9271

We now prefer the new `devToolbar` and `addDevToolbarApp` APIs over the overlay terms and check for the top-level config options correctly.

Upgraded to Astr 4 beta for types and added backward-compatible type cast (should be safe) to remove a few `ts-ignore`s we previously threw in.

closes #211

### DIFF
--- a/packages/astro/src/index.ts
+++ b/packages/astro/src/index.ts
@@ -66,7 +66,7 @@ const createPlugin = (options?: SpotlightAstroIntegrationOptions): AstroIntegrat
           const importPath = path.dirname(url.fileURLToPath(import.meta.url));
           const pluginPath = path.join(importPath, 'overlay/index.ts');
 
-          // `addDevToolbarApp` was added in Astro 4.0.0-beta.4
+          // `addDevToolbarApp` was added in Astro 4.0.0-beta.4.
           if (typeof addDevToolbarApp === 'function') {
             addDevToolbarApp(pluginPath);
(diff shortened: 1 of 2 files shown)


packages/astro/src/index.ts
```diff
@@ -30,26 +46,33 @@ const createPlugin = (options?: SpotlightAstroIntegrationOptions): AstroIntegrat
...
-            showTriggerButton = config.devOverlay ? false : true;
-          }
-          // @ts-ignore
-          if (config.experimental?.devOverlay) {
-            // @ts-ignore
-            showTriggerButton = config.experimental?.devOverlay ? false : true;
-          }
+          // Since Astro 4.0.0-beta.4, `devToolbar` is set and enabled by default.
+          // briefly, `devOverlay` was also added to the config but is now deprecated.
+          // Setting either of these to `true` or not setting any of them in the config file
```
matthewp, line 51: I'm so sorry 😔

PR #213, opened 2023-12-04
Add roadmap

HazAT:
Add a page about Spotlights Roadmap


dcramer:
Quick thoughts here:

V1:

- Overlay similar to Chrome Devtools leveraging Sentry SDKs to capture events
- Launch with native Astro support

V2:

- Expand platform native support (TBD)
- Enable Spotlight to replace native framework error pages

TBQH, we could really jsut focus on "whats left to do" as a roadmap pov kind of concept. I would be coy about _what_ we are going to support as that'll vary.

PR #216, opened 2023-12-04
add link to Syntax Hackweek episode

souredoutlook:
Adding context re: hackweek at Sentry

### DIFF
--- a/packages/website/src/content/docs/about.mdx
+++ b/packages/website/src/content/docs/about.mdx
@@ -7,7 +7,7 @@ sidebar:
   order: 0
 ---
 
-Spotlight started out as a Hackweek project at Sentry:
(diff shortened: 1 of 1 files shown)


Lms24 (approved):
Nice!

HazAT (approved):

packages/website/src/content/docs/about.mdx
```diff
@@ -7,7 +7,7 @@ sidebar:
   order: 0
 ---
 
-Spotlight started out as a Hackweek project at Sentry:
+Spotlight started out as a [Hackweek project](https://syntax.fm/show/666/hackweek-projects-realtime-markdown-editor-and-a-hardware-recording-button) at Sentry:
```
bruno-garcia, line 10: Find it odd that links to a project that's unrelated to spotlight though

souredoutlook: Fair - I think this is the only recent public facing record we have for what Hackweek is. 

They should interview David about Spotlight 😏

PR #217, opened 2023-12-04
Limit focused events to session-captured trace IDs

### LINKED ISSUE #188: Isolate Sentry Events in Spotlight to current tab
When having multiple tabs of the same application in dev mode open at the same time, spotlight will not distinguish events based on open tabs but instead show all events of all tabs in each tab. This is confusing and missleading and sometimes causes weird traces and errors (tracing w/o performance annoys us a little here...).  

(Although tbf, this actually helped me identify a [bug in the Astro SDK](https://github.com/getsentry/sentry-javascript/pull/9709) lol)

We should find a way to isolate traces and errors to the current tab. (This would all be much easier, if our SDK had a proper concept of sessions but what can you do...)

dcramer:
Adjust the behavior of Traces and Errors to show events which are associated with a known-local trace ID. If no trace IDs are known, assume all events should be present instead.

Fixes #188 

image


dcramer:
https://github.com/getsentry/sentry-javascript/issues/9737

Lms24 (approved):
I generally really like this change! We should be aware that the limitation you mentioned for Remix also exists for Astro SSR error pages (e.g. astro playground `/ssr-error` page) but maybe that's also okay. In fact, this might just trigger the fallback logic to show all events.

Lms24:
I'm gonna wait with merging #228 until this is merged (if we still wanna merge before tomorrow).

PR #226, opened 2023-12-05
feat: Dependency-inject custom logger into sidecar

### LINKED ISSUE #201: Improve sidecar logging
Right now it spits out quite a bit of debug information.

1. Consider making some of the log statements debug statements (e.g. received an event)
2. Consider DI to make it so e.g. the Astro logger could be used in Astro (I dont think this is feasible fwiw).

See also https://www.npmjs.com/package/debug

Lms24:
closes #201

This PR adjusts `setupSidecar` to

- accept an `options` object instead of just the port number
- inject a custom logger via the options object
- use a common logger interface and object that invokes the default logger or the custom logger if passed in

Furthermore, this PR:

- updates the Astro Spotlight package to use the Astro inegration logger for the side car
- updates the CLI and Vite plugin APIs for manually using the sidecar

Before opening this PR:

### DIFF
--- a/.changeset/poor-coins-sip.md
+++ b/.changeset/poor-coins-sip.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/spotlight': patch
+---
+
+ref(spotlight): Adjust `setupSidecar` call to pass port in options object
--- a/.changeset/brown-wasps-boil.md
+++ b/.changeset/brown-wasps-boil.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/astro': patch
+---
+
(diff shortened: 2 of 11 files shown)


.changeset/poor-coins-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spotlightjs/spotlight': patch
+---
+
+ref(spotlight): Adjust `setupSidecar` call to pass port in options object
```
Lms24, line 5:
```suggestion
ref(spotlight): Adjust `setupSidecar` calls to pass port in options object
```

Issue #230, opened 2023-12-05
Session-local events are not always reliable.


dcramer:
There are some cases where the local-to-session behavior does not work as desired. This generally is when the Sentry instrumentation is misconfigured or the SDK itself has incorrect behavior.

An example of this being an upstream problem can be seen in Remix:

https://github.com/getsentry/sentry-javascript/issues/9737

We'll use this isue to track this concern and identify improvmeents to the heuristics.

One example improvement we could make is, when the behavior is unreliable (a flag maybe?), bias towards events that are seen only within a period of time since the session became active. We still have an upper bound issue there, but possibly a similar check could exist for that.

Another issue we might consider making is a flag to disable this feature. We might consider this feature experimental (albeit desired), so keeping it on by default would still be the goal.

Refs #188

MisterJimson:
Just some feedback from my experience using Spotlight for the first time.

We have a mono repo with a backend and 2 frontend projects. When I open the overlay in the frontend, all of the backend errors and traces are shown as from a different session, even when initiated from the frontend session I am viewing. 

The backend correctly shows spans as part of the trace, so it does know the two are related.

dcramer:
@MisterJimson they have a diff trace ID tho, yeah?

The issue is we're looking for frontend-initiated traces. Its possible the instrumentation is bad here, or we have a bug.

Whats the stack? Specifically whats the frontend SDK and whats the backend SDK that should be interacting with it?

BYK:
Overlay is gone so the concept of "session-local events" is also gone.

PR #231, opened 2023-12-05
Add experiments configuration

dcramer:
Create a new init param (`experiments`) allowing us to configure additional behavior which may not adhere to semver in change cycles.

Refactors various types and constants.

Refs #230


Lms24 (approved):
Tested with Astro and it seems to work. Going to merge this in to cut 1.0

Issue #258, opened 2023-12-10
Provide a docker container with Spotlight 


mridang:
Id like to run Spotlight as a sidecar container and while this is possible to do so by building my own container, it would be handy to have an official one published

PR #260, opened 2023-12-10
Add docker image for sidecar

dcramer:
I wasn't able to get the port fully exposed on WSL2, which I think might be an issue w/ the image. Putting this up as a draft, but also need to figure out where we're publishing it (e.g. to Sentry's org?).

Fixes GH-258

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM node:lts-bullseye-slim AS base
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
+
+FROM base AS build
+WORKDIR /app
+COPY . /app
+# https://github.com/pnpm/pnpm/issues/6295
+RUN echo "dedupe-peer-dependents=false" > .npmrc
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
+RUN pnpm run --filter="./packages/sidecar" build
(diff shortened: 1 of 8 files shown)


sandstrom:
@dcramer Any updates on this?

A Docker image would be great! 🎉

dcramer:
Was blocked on figuring out how to get this on Sentry's docker presence, but let me circle back today!

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM node:lts-bullseye-slim AS base
```
Shubhdeep12, line 1: Hi @dcramer Just out of curiosity, why did you choose this Docker image of Node with this tag?

dcramer: Just chose an LTS version and went w/ slim to minimize the image. Lots of folks suggested using Google's lightweight images and/or chainguard, but the Docker image complexity gets intense fast. I do think it would be better to go w/ the Google approach as a v2 though.

dcramer:
Not sure if my brain is rusty or I fucked something up, but:

1. Docker publishing works, and seems to work great now.
2. I can't connect to HTTP on the image

```
docker run --name spotlight --net host ghcr.io/getsentry/spotlight:55bc409a6caa2f95861b4be2932ebd7ea9c81202
```

The image name get soutput as part of the publish step just FYI.

BeryJu:
Assuming you're running on macOS, you have explicitly expose port `8969` due to way host networking works on macOS, when doing that it seems to work as intended

PR #263, opened 2023-12-11
feat: Electron App

HazAT:
```[tasklist]
### Tasks
- [x] App Icons
- [x] Inject CSS styling into Spotlight -> Fullscreen
- [x] Remove Close button (and close functionality with keyboard)
- [x] Start Sidecar
- [x] Profit
```
image


timfish:
It's worth noting that you no longer need to use `custom-electron-titlebar` with modern versions of Electron since [Chromium added support for titlebar overlay](https://web.dev/articles/window-controls-overlay) when opening PWAs in their own windows.

You can access this via the `titleBarOverlay` `BrowserWindow` constructor option.

I've got an app which uses these options:
```ts
        // for Windows
        titleBarOverlay:
          process.platform === 'win32'
            ? { color: '#445D82', symbolColor: '#fff', height: 30 }
            : false,
        // for macOS
        titleBarStyle: 'hidden',
```

packages/overlay/src/components/Debugger.tsx
```diff
@@ -28,7 +30,7 @@ export default function Debugger({
         }
       }}
     >
-      <div className="sentry-debugger">
+      <div className={classNames('sentry-debugger', fullPage ? 'fullscreen' : '')}>
```
Shubhdeep12, line 33: Hi @HazAT , i think it would be better to have a unique classname here. Something like sentry-fullscreen or spotlight-fullscreen

HazAT: Just fyi, I did the change but the Shadow DOM we use make it not easily possible to overwrite these classes

HazAT:
I am going to merge this, the Electron App works but this PR also contains a new option `fullPage` which is helpful for rendering the HTML page.

PR #266, opened 2023-12-13
return 200 instead of 204 on stream endpoint

bruno-garcia:
With this change, I can merge https://github.com/getsentry/sentry-dotnet/pull/2961 as-is since it relies on all of Sentry's HTTP logic (including retry-after, etc). 

Otherwise I'll need to customize it to deal with 200 or 204, which is fine but since I imagine this might hit other SDKs adding support, changing the status code here might be the simplest.

### DIFF
--- a/.changeset/light-apes-applaud.md
+++ b/.changeset/light-apes-applaud.md
@@ -2,4 +2,4 @@
 '@spotlightjs/sidecar': minor
 ---
 
-stream endpoint returns 200 instead of 204
+ref(sidecar): Stream endpoint returns 200 instead of 204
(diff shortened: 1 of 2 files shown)


Lms24 (approved):
I think this is a fair change - ultimately we make things easier for SDKs if we pretend to be Relay 😅

.changeset/light-apes-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spotlightjs/sidecar': minor
+---
+
+stream endpoint returns 200 instead of 204
```
Lms24, line 5:
```suggestion
ref(sidecar): Stream endpoint returns 200 instead of 204
```

PR #267, opened 2023-12-13
install instructions dotnet

bruno-garcia:
Do I need a changeset for this? No package for the website showed for selection.

1.  Moved all snippets back a few characters, has enough padding IMO:

image

image

2. Added .NET (not ASP.NET Core or other framework but this should be clear enough):
image

ASP.NET Core has built-in environment (and env var `ASPNETCORE_ENVIRONMENT` but not referring to it here for simplicity). Also with ASP.NET Core you usr `UseSentry(o => ` instead of `SentrySdk.Init`.

Note the option now is called `EnableSpotlight` not aligning. The option to set a custom URL is `SpotlightUrl`. We could have `Spotlight` as a prop and use `implicit` operators to allow assigning either `string` or `bool` to it but it feels like a hack in C#. I expect us to have more props in the future Spotlight related and wanted to leave `Spotlight` as a name free to avoid breaking changes later so we can add a `SpotlightOption` class and prop `Spotlight`.

Finally, note that we're on `4.0.0-beta.5` now. Didn't list the `-beta.X` there to simplify things.

### DIFF
--- a/packages/website/src/components/sentryCodeBlock.mdx
+++ b/packages/website/src/components/sentryCodeBlock.mdx
@@ -66,7 +66,7 @@ import { Tabs, TabItem } from '@astrojs/starlight/components';
   :::caution
   Requires `sentry-dotnet` version `4.0.0` or higher.
   :::
-  ```csharp
+  ```csharp {3-6}
   SentrySdk.Init(o =>
   {
       o.Dsn = "___DSN___";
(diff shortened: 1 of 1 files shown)


packages/website/src/components/sentryCodeBlock.mdx
```diff
@@ -8,58 +8,73 @@ import { Tabs, TabItem } from '@astrojs/starlight/components';
...
+      # You should only load this in your development environment
+      config.spotlight = Rails.env.development?
+  end
+  ```
+  </TabItem>
+  <TabItem label=".NET">
+  :::caution
+  Requires `sentry-dotnet` version `4.0.0` or higher.
+  :::
+  ```csharp
```
Lms24, line 69: I suggest we highlight the spotlight-relevant lines:

```suggestion
  ```csharp {3-6}
```

Lms24 (approved):
Awesome, thanks! Changeset is not necessary for the website/docs. (I believe it shouldn't be selectable if you run `pnpm changeset:add` but need to check to confirm)

packages/website/src/components/sentryCodeBlock.mdx
bruno-garcia: Oh that's what this is. Thanks! ;)

Shubhdeep12 (approved):

PR #306, opened 2024-01-10
Add options for each registry for manual publishing

dcramer:
Adds a boolean for each of Docker, NPM, and Electron, to configure manual publishing workflow.

This is useful both in testing as well as resolving errors that may happen during a publish step.


HazAT (approved):
👍

PR #310, opened 2024-01-10
build docker image for arm64

BeryJu:
Before opening this PR:

Followup to #260, add a docker image that is built for arm64 (both to be used on linux and M1 macOS)


dcramer:
https://github.com/getsentry/spotlight/actions/runs/7482563907

PR #311, opened 2024-01-12
used spotlight package in dockerfile instead of just sidecar to serve index file

Shubhdeep12:
… index page.

Before opening this PR:

### DIFF
--- a/packages/spotlight/bin/run.js
+++ b/packages/spotlight/bin/run.js
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 const { setupSidecar } = await import('../dist/sidecar.js');
-const port = process.argv[2];
+const port = process.argv.length >= 3 ? Number(process.argv[2]) : undefined;
 import { join } from 'path';
 import { fileURLToPath } from 'url';
 setupSidecar({ port, basePath: join(fileURLToPath(import.meta.url), '../../dist/overlay/') });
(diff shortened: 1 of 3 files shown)


packages/spotlight/bin/run.js
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 const { setupSidecar } = await import('../dist/sidecar.js');
-const port = process.argv[2];
+const port = process.argv.length >= 3 ? Number(process.argv[2]) : undefined;
```
Lms24, line 3: good catch!

Lms24 (approved):

HazAT (approved):
💯

Shubhdeep12:
This will fail the docker test
seems like we have to build everything - overlay, sidecar and then spotlight to make this work

HazAT:
> This will fail the docker test seems like we have to build everything - overlay, sidecar and then spotlight to make this work

Right makes sense, we should build everything

packages/spotlight/bin/run.js
Shubhdeep12: yes, but it was done by @dcramer initially in another PR. I thought its better to add it here as well.

Shubhdeep12:
I think now it will work :)

packages/spotlight/bin/run.js
dcramer: would be great if we could figure out how to dedupe this script, but this is beyond my domain knowledge w/ npm packages

dcramer (approved):

Issue #294, opened 2023-12-20
PHP Stack trace upside down


adnweedon:
### Environment

I'm using the electron app v0.0.12, connected to PHP SDK 4.2.0.

### Steps to Reproduce

If I throw an exception in a controller in my code, and then view it in Spotlight, the stack trace preview at the top is the first file it hits in the system, and the code preview of the line where I throw the exception is at the very bottom of the stack if I scroll down.

### Expected Result

The sentry.io UI shows the line where an exception is triggered at the top, and then works its way out as the you go down the stack.

### Actual Result

I don't really want to put a screenshot as it reveals a load of our code structure, but the expanded code snippet with the highlighted line is the public/index.php, and if I scroll to the bottom and expand the last item in the trace, it is the line where I throw the Exception - the stack just needs reversing!

Lms24:
Hmm we have some logic in spotlight's sentry data cache to reverse the stack frames list because SDKs seem to send it in different order. Maybe we're either not including the php SDK when reversing or we wrongfully include it when it's events don't need reversing 😓

Shubhdeep12:
Hi @adnweedon 
if this is similar to this issue - https://github.com/getsentry/spotlight/issues/262
Please update the app and check again.
Thanks

adnweedon:
I'ev just updated to 0.0.13 and it's still an issue!

adnweedon:
It looks like https://github.com/getsentry/spotlight/commit/00c86f1b574450ab1a0f62861a688965325699e7 stopped reversing it, but I'm not sure what has changed between then and now that means it's upside down again, so I don't want to just undo the PHP bit?

Shubhdeep12:
Hi @adnweedon
This is fixed in the recent version of Spotlight.
Please verify.
Thanks.

adnweedon:
Yep, all sorted - thanks for that!

PR #312, opened 2024-01-13
fixes: error stacktrace reversing

Shubhdeep12:
Before opening this PR:
#294 

Removed the logic to not reverse the stacktrace of error events when platform is PHP or python.

### DIFF
--- a/packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
+++ b/packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
@@ -241,11 +241,13 @@ function reverseStackTraces(errorEvent: SentryErrorEvent): void {
   if (!errorEvent.exception || !errorEvent.exception.values) {
     return;
   }
+
+  // TODO: To check if stacktrace is already reverse or not for php and python.
   // Check if errorEvent.platform contains php, or python and if so return early
   // as the stacktrace is already in the correct order.
-  if (errorEvent.platform && (errorEvent.platform.includes('php') || errorEvent.platform.includes('python'))) {
-    return;
-  }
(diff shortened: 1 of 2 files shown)


packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
```diff
@@ -241,11 +241,13 @@ function reverseStackTraces(errorEvent: SentryErrorEvent): void {
   if (!errorEvent.exception || !errorEvent.exception.values) {
     return;
   }
+
+  // TODO: To check if stacktrace is already reverse or not for php and python.
```
Lms24, line 245: We have a sample php error (`envelope_php_error.txt`) in the overlay package. It should be sent to the overlay if you run `pnpm sample` in the overlay package. Let's check if we can remove the commented code before merging this.

Shubhdeep12: yes @Lms24 checked that. And based on @HazAT response here - https://discord.com/channels/621778831602221064/1195676075271913502/1195677537683120139

I made this code comment. in case if we wanna check and come back to this later.

let me know if we should remove the commented code.

Lms24: Ah sorry, I missed in your reply that you checked the sample data. 

Yes, I'd prefer if we removed the commented code. We can always revert this PR or go back in the git history if we encounter problems.

Shubhdeep12: Yes. 
Done. Removed the comments.

Issue #313, opened 2024-01-14
To capture DB queries similar to Sentry dashboard.


Shubhdeep12:
Based on @HazAT 's input creating this issue to have performance section on spotlight to capture DB queries

Shubhdeep12:
Hi @HazAT lets close this also if we are good.

Thanks

PR #314, opened 2024-01-14
feat(performance tab): To capture DB queries on spotlightjs

Shubhdeep12:
Before opening this PR:
#313 

image
image
image
image
image
image

And a prop name fixed- changed from backTo -> backto


HazAT (approved):
I think it's a good start - we should polish this more

PR #319, opened 2024-01-17
Docs: fixed nextjs docs

Shubhdeep12:
Before opening this PR:
fixed code highlight

### DIFF
--- a/.changeset/beige-beans-eat.md
+++ b/.changeset/beige-beans-eat.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/overlay': minor
+---
+
+Added performance and nested queries tab in spotlight to capture DB queries
--- a/.changeset/wicked-olives-roll.md
+++ b/.changeset/wicked-olives-roll.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/overlay': patch
+---
+
+Reverse the error stack trace for php and python platform events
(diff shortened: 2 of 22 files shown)


Lms24 (approved):

Lms24 (changes requested):
Sorry, missed something. I believe the `spotlight` option doesn't need to be set at all in the client `Sentry.init` call. I don't see it being used anywhere in the Browser SDK code and the spotlight overlay hooks itself into Sentry to send events to the sidecar: https://github.com/getsentry/spotlight/blob/fed3835e989b10b82d52a079be15738bf5527c61/packages/overlay/src/integrations/sentry/sentry-integration.ts#L63-L85

Shubhdeep12:
> Sorry, missed something. I believe the `spotlight` option doesn't need to be set at all in the client `Sentry.init` call. I don't see it being used anywhere in the Browser SDK code and the spotlight overlay hooks itself into Sentry to send events to the sidecar: https://github.com/getsentry/spotlight/blob/fed3835e989b10b82d52a079be15738bf5527c61/packages/overlay/src/integrations/sentry/sentry-integration.ts#L63-L85

Yes @Lms24. I missed it. Fixed the docs.

Please review.


Btw,  does it makes sense to add this logic? Ideally we should give user this flexibility to stop tracing for just client side.

Lms24 (approved):

Lms24:
@Shubhdeep12 for now I'd not add the logic to the browser SDKs because we're reworking quite a lot of stuff in the SDK for v8. Let's address this (in some capacity) if people request to disable showing events from the browser.

PR #322, opened 2024-01-18
Vite Integration docs for spotlight/sidecar fix.

theetherGit:
Before opening this PR:

Vite Integration docs for spotlight/sidecar fix.

### DIFF
--- a/packages/website/src/content/docs/sidecar/vite.mdx
+++ b/packages/website/src/content/docs/sidecar/vite.mdx
@@ -16,12 +16,12 @@ We provide a Vite plugin that automatically starts & stops Sidecar for you when 
   </TabItem>
   <TabItem label="pnpm">
     ```sh
-    pnpm add @spotlightjs/spotlight --save-dev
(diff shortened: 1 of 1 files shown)


Shubhdeep12 (approved):
thanks

Lms24 (commented):
Hmm it's been a while since I worked on this but IIRC, the `@spotlightjs/spotlight` package should also be able to run the sidecar. Is there a particular reason, adding the `@spotlightjs/sidecar` package makes more sense?

Lms24 (changes requested):
Just gonna request changes so we don't accidentally merge this before we clarify the intent

Shubhdeep12:
Yes @Lms24 
running npx @spotlightjs/spotlight will run the sidecar. 

 But looks like its one of the way to run sidecar for vite powered apps. And for them it is needed to import this package. 
And this was a typo error which was fixed in this PR

Lms24 (approved):
Oh yeah that makes sense, thx for explaining!

PR #330, opened 2024-01-25
feat(performance): Web vitals added

Shubhdeep12:
Before opening this PR:

- [X] I added a [Changeset Entry](https://spotlightjs.com/contribute/changesets/) with `pnpm changeset:add`
- [] I referenced issues that this PR addresses

Skipped conditions:
1. Since we are not receiving every required metric to calculate the score, we have skipped the condition that prevents the calculation of the performance score in the absence of the necessary metrics. Instead, we are calculating the score with the received metrics.
2. We are consistently using weights and conditions for the Chrome browser because we are not receiving browser information in the event data.

Screenshot from 2024-01-25 23-30-05

Screenshot from 2024-01-25 23-30-18

Screenshot from 2024-01-25 23-30-27

Screenshot from 2024-01-25 23-30-52

### DIFF
--- a/.changeset/good-balloons-compete.md
+++ b/.changeset/good-balloons-compete.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/overlay': minor
+---
+
+Added WebVitals in Performance Section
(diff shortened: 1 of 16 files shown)


HazAT (approved):
💯

Issue #344, opened 2024-01-31
Add a button to the spotlight overlay, that cleans all events


HazAT:
Also, empty the buffer in the sidecar.

PR #345, opened 2024-01-31
feat(reset): Added functionality in spotlight to reset data.

Shubhdeep12:
Before opening this PR:

#344 

1. Added a DELETE /clear API route for the sidecar to wipe out all data from the buffer.
2. Added a cta in overlay to clear data(only present when sentry integration is added).

[Screencast from 01-02-24 12:31:12 AM IST.webm](https://github.com/getsentry/spotlight/assets/43654389/f2544587-8236-4a72-8d72-f500daed0a86)

### DIFF
--- a/packages/overlay/src/components/Debugger.tsx
+++ b/packages/overlay/src/components/Debugger.tsx
@@ -1,8 +1,16 @@
+import { useContext } from 'react';
+import { useNavigate } from 'react-router-dom';
 import { ReactComponent as Logo } from '~/assets/glyph.svg';
 import { Integration, IntegrationData } from '~/integrations/integration';
+import sentryDataCache from '~/integrations/sentry/data/sentryDataCache';
+import { SentryEventsContext } from '~/integrations/sentry/data/sentryEventsContext';
+import { getNativeFetchImplementation } from '~/integrations/sentry/sentry-integration';
 import { NotificationCount } from '~/types';
 import classNames from '../lib/classNames';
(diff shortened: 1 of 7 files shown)


packages/overlay/src/components/Debugger.tsx
```diff
@@ -48,7 +57,31 @@ export default function Debugger({
...
   setTriggerButtonCount: (count: NotificationCount) => void;
   fullPage: boolean;
+  setReloadSpotlight: React.Dispatch<React.SetStateAction<number>>;
 }) {
+  const isSentryIntegrationAdded = integrations.some(integration => integration.name === 'sentry');
+  const { setEvents } = useContext(SentryEventsContext);
+  const navigate = useNavigate();
+
+  const clearEvents = () => {
+    const sidecarUrl = `http://localhost:${PORT}/clear`;
```
HazAT, line 67: I think someone in theory could use a different URL than `localhost` right?
can we pass down the URL?

Shubhdeep12: yes, we can, but it would require creating a `DELETE /stream` API endpoint instead of using the `/clear` endpoint.

 At present, the user can establish a connection between the overlay and the sidecar by providing a custom sidecar stream endpoint.

Shubhdeep12: if that is ok, i'll push the changes after confirmation.

Thanks

Issue #350, opened 2024-02-06
Clear Event button not clickable in Electron App


HazAT:
Because the Drag header overlays it

The fix for this is, make the button hideable as an option, and don't show it in electron.

Shubhdeep12:
Hi @HazAT can you please share a screenshot for this? 
it works fine when I run it locally, the button is clickable and visible.

![image](https://github.com/getsentry/spotlight/assets/43654389/91603293-a89c-46ff-9fec-ad8a76027945)

HazAT:
This is how it looks for me
<img width="1012" alt="image" src="https://github.com/getsentry/spotlight/assets/363802/f1b3e6be-44b6-4428-9598-7f70e83cfdab">

I would still and an option and hide it in the electron app
there someone can just do CMD+R to achieve the same thing

Shubhdeep12:
ok @HazAT let's remove this button for Electron App.
And we'll make reload and force-reload to clear events with their default functionality.

PR #352, opened 2024-02-07
fix(electron): Clear events button fixes for electron app.

Shubhdeep12:
Before opening this PR:

fixes: #350  

1. Removed clear events button for electron app.
2. Added clear events function in Reload and force reload toolbar options in electron app.
3. Created a event listener to clear events data.
4. Added a reset function in integration.
5. Updated the docs and added `reset` function in integration type and `showClearEventsButton` for configuration page

### DIFF
--- a/packages/overlay/src/App.tsx
+++ b/packages/overlay/src/App.tsx
@@ -1,8 +1,11 @@
-import { useEffect, useState } from 'react';
+import { useContext, useEffect, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import Debugger from './components/Debugger';
 import Trigger from './components/Trigger';
 import type { Integration, IntegrationData } from './integrations/integration';
+import sentryDataCache from './integrations/sentry/data/sentryDataCache';
+import { SentryEventsContext } from './integrations/sentry/data/sentryEventsContext';
+import { getNativeFetchImplementation } from './integrations/sentry/sentry-integration';
 import { getSpotlightEventTarget } from './lib/eventTarget';
 import { log } from './lib/logger';
 import useKeyPress from './lib/useKeyPress';
@@ -21,7 +24,7 @@ export default function App({
   fullPage = false,
 }: AppProps) {
   log('App rerender');
-
+  const { setEvents } = useContext(SentryEventsContext);
   const [integrationData, setIntegrationData] = useState<IntegrationData<unknown>>({});
   const [isOnline, setOnline] = useState(false);
   const [triggerButtonCount, setTriggerButtonCount] = useState<NotificationCount>({ count: 0, severe: false });
@@ -57,6 +60,25 @@ export default function App({
 
   const navigate = useNavigate();
 
+  const clearEvents = () => {
+    const makeFetch = getNativeFetchImplementation();
+    makeFetch(sidecarUrl, {
+      method: 'DELETE',
+      mode: 'cors',
+    }).catch(err => {
+      console.error(
+        `Sentry SDK can't connect to Sidecar is it running? See: https://spotlightjs.com/sidecar/npx/`,
+        err,
+      );
+      return;
+    });
+
+    setEvents({ action: 'RESET', e: [] });
+    sentryDataCache.resetData();
+    navigate('/errors');
+    setReloadSpotlight(prev => prev + 1);
+  };
+
   useEffect(() => {
     const onOpen = (
       e: CustomEvent<{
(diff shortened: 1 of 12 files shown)


packages/overlay/src/components/Debugger.tsx
```diff
@@ -57,30 +49,9 @@ export default function Debugger({
   isOnline: boolean;
   setTriggerButtonCount: (count: NotificationCount) => void;
   fullPage: boolean;
-  setReloadSpotlight: React.Dispatch<React.SetStateAction<number>>;
 }) {
   const isSentryIntegrationAdded = integrations.some(integration => integration.name === 'sentry');
-  const { setEvents } = useContext(SentryEventsContext);
```
HazAT, line 63: I would rather not have any traces of Electron app in Spotlight itself when not absolutely necessary. Rather let's add an option - `showClearEventsButton` to Spotlight itself, which can be default `true` and in the electron app when we init Spotlight we set it to false.

packages/overlay/src/App.tsx
```diff
@@ -81,11 +108,13 @@ export default function App({
     spotlightEventTarget.addEventListener('open', onOpen as EventListener);
     spotlightEventTarget.addEventListener('close', onClose);
     spotlightEventTarget.addEventListener('navigate', onNavigate as EventListener);
+    spotlightEventTarget.addEventListener('sentry:clearEvents', clearEvents as EventListener);
```
HazAT, line 111: Any specific reason why `sentry:clearEvents`? 
I would just call it `clearEvents`

packages/overlay/src/App.tsx
```diff
@@ -57,6 +60,30 @@ export default function App({
...
+    const sidecarUrlObject: URL = new URL(sidecarUrl);
+    const host: string = sidecarUrlObject?.hostname;
+    const port: string = sidecarUrlObject?.port;
+    const clearEventsUrl: string = `http://${host}:${port}/clear`;
+    makeFetch(clearEventsUrl, {
+      method: 'DELETE',
+      mode: 'cors',
+    }).catch(err => {
+      console.error(
+        `Sentry SDK can't connect to Sidecar is it running? See: https://spotlightjs.com/sidecar/npx/`,
```
HazAT, line 74:
```suggestion
        `Spotlight can't connect to Sidecar is it running? See: https://spotlightjs.com/sidecar/npx/`,
```

packages/overlay/src/App.tsx
```diff
@@ -1,8 +1,11 @@
-import { useEffect, useState } from 'react';
+import { useContext, useEffect, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import Debugger from './components/Debugger';
 import Trigger from './components/Trigger';
 import type { Integration, IntegrationData } from './integrations/integration';
+import sentryDataCache from './integrations/sentry/data/sentryDataCache';
```
HazAT, line 6: That's not ideal, we should find another way to do this - we are adding something from the Sentry integration to the main skeleton app making Spotlight practically required to always use Sentry.

We should refactor this that Spotlight forwards all incoming events to all integrations. Integrations that also listen to the event do something then.

HazAT (changes requested):

packages/overlay/src/components/Debugger.tsx
Shubhdeep12: 👍🏻

packages/overlay/src/App.tsx
Shubhdeep12: Right now, it clears out the data from the sidecar and events stored for sentry integration only. so I thought to make it just for sentry integration
that was the reason to add sentry: here 
and to check if sentry integration was added, only show the clear events button.

but I made that generic now, so I am changing this to "clearEvents"

packages/overlay/src/App.tsx
Shubhdeep12: 👍🏻

packages/overlay/src/App.tsx
Shubhdeep12: Yes @HazAT , that makes sense.

In earlier commits, when we added clear events, I only implemented it for Sentry. However, this approach could potentially break the spotlight if the user hasn't added Sentry integration.

To address this, I've updated the logic by checking for the existence of the `processEvent` function for every integration. If the function is present, we clear the `integrationData` for that specific integration(which includes the events for every integration)
Additionally, for integrations like Sentry, which process events but have internal storage (like we have `sentryDataCache`), I've added a reset function in the integration. We will call this function while clearing events, allowing users to clear events in the integration using the button.

Approach Summary:
- Check for the existence of the processEvent function for each integration.
- If the function is present, clear the integration data (including events).
- For integrations with internal storage, implement a reset function in the integration.
- Call the reset function while clearing events to enable users to clear events using the button.

HazAT (approved):
🥇

Issue #363, opened 2024-02-23
To have spotlight error page in case of SSR-error


Shubhdeep12:
Based on the discussion here with @dcramer  - https://discord.com/channels/621778831602221064/1201983104047120454/1204124322172502056

For Astro,
- how about closing Astro's Error overlay and display the Spotlight error page instead?
- And, till the time error is not received the errors listed but most of the time there will be error and it will be opened using a option passed in sentry integration 
- as `showFirstIncomingError: true`

Added a PR as a POC for this.
cc @Lms24 @HazAT

PR #364, opened 2024-02-23
feat: POC for adding spotlight overlay for ssr-errors 

Shubhdeep12:
Before opening this PR:

#363 

https://vimeo.com/915998606/e0200955c6

- Added a new option in sentry integration to display the first incoming error - `showFirstIncomingError?: boolean`
- if passed as true, the first encountered error will be opened up automatically.
- Also, I closed Astro's custom error overlay and made changes to the plugin created by @Lms24 in @spotlight/astro to open spotlight in ssr-error page


Lms24 (commented):
Your video shows the problem here pretty well: We're showing the island error although we actually should be showing the "random error" (SSR error) when it occurs. So maybe showing the _first_ error is not the way to go. Ideally, we rather want to show the most _recent_ error. This should be more correct in general but I guess there's the timing problem again.

Can you check if the SSR error is already sent to the overlay by the time the error page is rendered? If yes, we _should_ be good.

Another idea I just had (gonna note it down here so we don't completely forget):

We could try this (NO idea if it works). 

- We register an event processor on the server SDK that writes the most recent eventId to the global object.
  - To try this, we can simply call something like this in an Astro test app:
     ```js
    Sentry.addEventProcessor(event => {
      if (!event.type) {
        globalThis.__Sentry_last_event_id = event.event_id;
      }
    })
     ```
  - If this works, we add it to the `Spotlight` integration in the Sentry JS SDK.
- In the vite plugin, we access the global object and get the id
- We add an option to the overlay init (something like `waitForEventId: string`) and if it is defined, we immediately open spotlight and wait for the event with the id to come in and show it. 

This requires the Vite plugin to run in the same process as the Astro dev server so that they actually share the global object. I believe this is the case but I never tried it out. Might be worth to explore though!

Shubhdeep12:
Yes its worth trying. 

I'll try both the ideas out. 💡

Shubhdeep12:
@Lms24 looks like the SSR error is sent to overlay by the time the error page is rendered and we can open the last error received to overlay - which is approach 1
I've pushed the code for this approach

Shubhdeep12:
And @Lms24  when it comes to Approach 2, which is more logical in terms of error tracking, I tried this approach by adding Sentry.addEventProcessor in the Sentry.server.config.ts file in the demo app and tracked the event ID, as you suggested above.

I used that as the option in the Sentry integration while initializing the spotlight for the SSR-error page. This works for the first time, but I am not sure whether the Vite plugin caches the transform function created for the first time or something else, as it doesn't update the value retrieved from globalThis. Even if the value is updated, it always takes the value set for the first time.

packages/astro/src/snippets.ts
```diff
@@ -23,6 +28,7 @@ Spotlight.init({
   ],
   showTriggerButton: ${options.showTriggerButton === false ? 'false' : 'true'},
   injectImmediately: ${options.injectImmediately === true ? 'true' : 'false'},
+  fullPage: ${options.fullPage === true ? 'true' : 'false'},
```
BYK, line 31: Curious, why don't we do `Boolean(options.fullPage).toString()`? (and for all the others?)

Or even better, something like:

```js

const init_options = {
  integrations: integrationCalls,
  ...options,
};

return `Spotlight.init(${JSON.stringify(init_options))})`
```

packages/astro/src/vite/error-page.ts
```diff
@@ -17,15 +17,24 @@ export const errorPageInjectionPlugin: (options: ErrorPagePluginOptions) => Plug
...
+        showTriggerButton: false,
+        fullPage: true,
         // don't wait for the window.load event to be fired because in the error page,
         // this already happened before spotlight is initialized
         injectImmediately: true,
+        showRecentError: true,
       });
 
-      const modifiedCode = `${code}\n${initSnippet}\n`;
+      // Checking if there is a ErrorOverlay class added by Astro
```
BYK, line 28: This definitely feels quite fragile. Do we do this kind of find/replace on actual JS/TS code somewhere?

packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -43,13 +44,25 @@ export default function sentryIntegration(options?: SentryIntegrationOptions) {
...
         .getEvents()
         .filter(
           e =>
             e.type != 'transaction' &&
             (e.contexts?.trace?.trace_id ? sentryDataCache.isTraceLocal(e.contexts?.trace?.trace_id) : null) !== false,
-        ).length;
+        );
+
+      const errorsCount = errors.length;
+      if (options?.showRecentError && errorsCount) {
```
BYK, line 56:
```suggestion
      if (errorsCount && options?.showRecentError) {
```

Cheaper (and easier on the eyes & mind) to check the count first.

BYK (changes requested):

BYK (approved):

Issue #370, opened 2024-02-28
Docs for frameworks are inconsistent


Lms24:
Just looked through various JS framework docs for spotlight and they're a bit incosistent in terms of structure.

Let's align them with the astro doc:

- add numbered steps
- (1. installation, 2. configuration, 3. other SDKs)
- show sidecar instructions ("Run the Sidecar") when applicable

PR #371, opened 2024-02-29
fix(docs): made the docs consistent for all frameworks similar to astro

Shubhdeep12:
Before opening this PR:

- [ ] I added a [Changeset Entry](https://spotlightjs.com/contribute/changesets/) with `pnpm changeset:add`

#370

### DIFF
--- a/packages/website/src/content/docs/setup/other.mdx
+++ b/packages/website/src/content/docs/setup/other.mdx
@@ -88,28 +88,12 @@ Upon loading your UI you should now see the Spotlight toolbar in bottom-right co
 
 For more options on how to configure Spotlight's UI, see [Configuration](/reference/configuration/).
 
-:::tip
-Spotlight will automatically configure [Sentry's Browser SDKs](https://docs.sentry.io/platforms/javascript/) to send data to the default Sidecar.
+<LinkCard title="Setup Spotlight with Sentry" description="Instructions for Web and Mobile SDKs" href="#bonus-leverage-sentry-sdks" />
 
-If you are using additional runtimes, such as [Node.js](https://docs.sentry.io/platforms/node/), [PHP](https://docs.sentry.io/platforms/php/), or [Python](https://docs.sentry.io/platforms/python/), you should take a look at Sentry's SDKs on how to enable Spotlight. Generally this as easy as `spotlight: true` in `Sentry.init()`:
-
-
-```js {2}
-Sentry.init({
-  spotlight: true
-})
-```
-
-:::
-
-<LinkCard title="Setup Spotlight with Sentry" description="Instructions for Web and Mobile SDKs" href="#leverage-sentry-sdks" />
-
----
 
 ## 3. Run the Sidecar
 
 Spotlight's [architecture](/architecture/) requires a sidecar to be running alongside your service. This is to enable streaming data collection from your backend services into the Spotlight overlay.
-
 
 <CardGrid>
   <LinkCard title="Vite" href="/sidecar/vite/" />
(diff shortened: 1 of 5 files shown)


packages/website/src/content/docs/setup/other.mdx
```diff
@@ -116,7 +118,7 @@ Spotlight's [architecture](/architecture/) requires a sidecar to be running alon
 </CardGrid>
 
 
-## Leverage Sentry SDKs
+## Bonus: Leverage Sentry SDKs
 
 To get the most value out of Spotlight, configure Sentry's SDKs. You don't need to be using the Sentry service to take advantage of Sentry's SDKs to power Spotlight.
```
Lms24, line 123:
```suggestion
To get the most value out of Spotlight, configure Sentry's SDKs. You don't have to use sentry.io but you can use the them to just power Spotlight.
```

packages/website/src/content/docs/setup/sveltekit.mdx
```diff
@@ -9,16 +9,21 @@ import { Tabs, TabItem } from '@astrojs/starlight/components';
 This guide gets Spotlight up and running in SvelteKit with Sentry. 
 You don't strictly need Sentry for Spotlight to work, we highly recommend it.
```
Lms24, line 10:
```suggestion
You don't strictly need Sentry for Spotlight to work but we highly recommend it.
```

packages/website/src/content/docs/setup/sveltekit.mdx
```diff
@@ -9,16 +9,21 @@ import { Tabs, TabItem } from '@astrojs/starlight/components';
...
 
-## Install Sentry 
+## 1. Installation
+
+#### Install Sentry
```
Lms24, line 14: Is there a reason we're skipping the h3 level (`###`) here?

packages/website/src/content/docs/setup/sveltekit.mdx
```diff
@@ -9,16 +9,21 @@ import { Tabs, TabItem } from '@astrojs/starlight/components';
...
 
 Follow the [Sveltekit Getting Started](https://docs.sentry.io/platforms/javascript/guides/sveltekit/) guide on Sentry to install and configure Sentry.
```
Lms24, line 16:
```suggestion
Follow the [Sveltekit Getting Started](https://docs.sentry.io/platforms/javascript/guides/sveltekit/) guide to install and configure Sentry.
```

packages/website/src/content/docs/setup/sveltekit.mdx
Shubhdeep12: No, there isn't a specific reason for choosing h4 @Lms24 
The visual difference in sizes between h3 and h2 was minimal, making it difficult to distinguish on the UI, i.e. why I opted for h4 to ensure clearer visual hierarchy.

Lms24: Ah I see, yeah no worries then let's leave it as is. Maybe we should make the difference between h2 and h3 stronger in general but let's not do this in this PR.

packages/website/src/content/docs/setup/other.mdx
```diff
@@ -104,7 +106,7 @@ Sentry.init({
 
 ---
 
-## Run the Sidecar
+## 3. Run the Sidecar
```
Lms24, line 109: Just noticed (can't mark the line though in the GH UI): Can we remove the `:::tip` block in lines 91-103? I think this is essentially a duplicate of the newer "Leverage Sentry SDKs" section

Shubhdeep12: Done, removed. the Leverage Sentry SDK section is fine. no need of this tip.

Lms24 (approved):
thanks!

PR #386, opened 2024-03-28
feat: Make Spotlight compatible with Sentry SDK version 8

Lms24:
This PR makes spotlight compatible with version 8 of the Sentry JS SDK:

- Convert the old class-based v7 integration into a function-based integration (`Integration` interface is no longer compatible w/ v7 < 7.99.0)
- Add extra logic to enable SDK integrations if no DSN is provided
- Update e2e and demo apps to v8 alpha versions

This PR is directed against `v2-dev`. Once SDK v8 goes stable, we'll make a stable spotlight release.

This is a **breaking change** because the browser SDK integration is no longer compatible with the v7 SDK versions below 7.99.0. So users will be able to use the SDK with >=7.99.0 or 8.x.

ref #387


packages/astro/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@spotlightjs/astro",
   "description": "Add spotlight to your Astro toolbar, for better debugging.",
-  "version": "1.0.19",
+  "version": "2.0.0-alpha.0",
```
Lms24, line 4: Usually, this step would happen in the "Prepare Publish" workflow but changesets handles pre-releases a bit differently. In the hope of not having to do too many, for now I opted to just carry out the chagneset pre-release preparations manually and I'll publish directly from `v2-dev` once this is merged. Should work 🤞

packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -143,16 +143,66 @@ export function processEnvelope(rawEvent: RawEventContext) {
...
+    return;
+  }
+
+  const sentryClient = sentryGlobal.getClient();
+  if (!sentryClient) {
+    warn("Couldn't find a Sentry SDK client. Make sure you're using a Sentry SDK with version 8.0.0 or higher");
+    return;
+  }
+
+  if (!sentryClient.getOptions().dsn) {
```
Shubhdeep12, line 184: Hi @Lms24 we should not force the user to remove DSN altogether to use spotlight.

Shubhdeep12 (changes requested):

packages/overlay/src/integrations/sentry/index.ts
Lms24: That's not what we're doing here. We only check if users configured Sentry without a DSN. If that's the case, the SDK will not initialize integrations by default (e.g. the ones that instrument the global error handlers). So in order for Spotlight to pick up events from the host app, we need to initialize the integrations "manually". 
If users already set a DSN, there's nothing for us to do other than adding the spotlight integration (see code below).

Shubhdeep12: Oh, i got it. When dsn is set, sdk will initialize the integrations. 

But what about invalid DSN?

Actually, i tested this locally with dsn value pointed to env variable, and with env variable value empty
And for me, integrations were not intialized.

Shubhdeep12: Let me check if integrations were not intialized or spotlight integration was not added. 
Because spotlight was not retrieving any envelopes.

MIGRATION.md
```diff
@@ -0,0 +1,15 @@
+# Updating from Spotlight 1.x to 2.x
+
+Spotlight was updated to version 2 to ensure compatibility with version 8 of Sentry's JavaScript SDKs.
+
+Here's what you need to do to update to version 2:
+
+- Ensure that you use a version 8 of your `@sentry/<sdk>` SDKs
```
mydea, line 7: I would propose to support v7.99.0+ (https://github.com/getsentry/sentry-javascript/blob/v7/CHANGELOG.md#7990) which should work, and makes it easier for people to update. That version should have stuff that we need to make this work, I believe?

demos/astro-playground/astro.config.mjs
```diff
@@ -31,6 +31,6 @@ export default defineConfig({
         enabled: false, // tmp deactivate until version after 7.83.0
       },
     }),
-    spotlight(),
+    spotlight({ debug: true }),
```
mydea, line 34: is this on purpose?

Lms24: nope. It probably doesn't really matter, it's a demo/playground app and debug logs were helpful here 😅 

but I'll remove it again

MIGRATION.md
Lms24: we could probably even go down to 7.96.0 (https://github.com/getsentry/sentry-javascript/blob/v7/CHANGELOG.md#7960) where we started exposing the functional integrations in browser. 

Or we set it to 7.100.0 where we exposed the integrations in node. Probably the safer choice just to be sure the API is available everywhere.

demos/astro-playground/astro.config.mjs
Lms24: oh nvm, 7.97.0 it is (missed that the 7.96.0 release was incomplete)

MIGRATION.md
Lms24: > we could probably even go down to 7.96.0
oh nvm, 7.97.0 it is (missed that the 7.96.0 release was incomplete)

Lms24: update: 7.97 doesn't yet support `afterAllSetup`. 7.99.0 does, so trying this next

packages/overlay/src/integrations/sentry/index.ts
Shubhdeep12: Yes @Lms24 i was adding an invalid DSN
So let's use `getDsn` instead of `getOptions.dsn` because it looks like we validate `dsn` in SDK before setting `_dsn` in senryClient.

```suggestion
  if (!sentryClient.getDsn()) {
```

MIGRATION.md
Lms24: ok, we now require 7.99.0 or greater 👍

packages/overlay/src/integrations/sentry/index.ts
Lms24: good call! I was experimenting with emulating the `isEnabled` check but it's probably better to limit the check to the DSN since users might want to explicilty disable the SDK via `enabled: false`.

Lms24: Updated in [c43f007](https://github.com/getsentry/spotlight/pull/386/commits/c43f0074482b2b70e8493124133596d79d0f1f57)

mydea (approved):

Shubhdeep12 (approved):

PR #393, opened 2024-04-04
fix(overlay): To display metrics event as well in event envelopes

Shubhdeep12:
Before opening this PR:

Earlier, we were skipping the events with the type as "statsd" (metric events). 
Because the type of items received in an envelope was not JSON. 

Now, with this fix, we will assign the string directly instead of parsing if the type is not JSON in the event envelope


HazAT (approved):

HazAT:
Nice

PR #401, opened 2024-05-28
Added a check for spotlight v1.x.x to not load sentry sdk >= v8

Shubhdeep12:
Before opening this PR:

Added a check for sentry integration in Spotlight, for v1.x.x to load only if - 
- no browser sdk found.
- browser sdk with version <= 7.x.x

image

### DIFF
--- a/packages/overlay/src/integrations/sentry/index.ts
+++ b/packages/overlay/src/integrations/sentry/index.ts
@@ -10,7 +10,7 @@ import InfoTab from './tabs/InfoTab';
 import PerformanceTab from './tabs/PerformanceTab';
 import SdksTab from './tabs/SdksTab';
 import TracesTab from './tabs/TracesTab';
-import { WindowWithSentry } from './types';
+import { LegacyCarrier, WindowWithSentry } from './types';
 import { checkBrowserSDKVersion } from './utils/sdkValidation';
 
 const HEADER = 'application/x-sentry-envelope';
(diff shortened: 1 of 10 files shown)


packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -18,98 +21,108 @@ type SentryIntegrationOptions = {
...
-      const localTraces = sentryDataCache.getTraces().filter(t => sentryDataCache.isTraceLocal(t.trace_id) !== false);
-
-      return [
-        {
-          id: 'errors',
-          title: 'Errors',
-          notificationCount: {
-            count: errorsCount,
-            severe: errorsCount > 0,
+  if (checkBrowserSDKVersion()) {
```
Lms24, line 24: Can we extract the two integrations here so that not the entire code is indented by `checkBrowserSDKVersion`?

packages/overlay/src/integrations/sentry/utils/sdkValidation.ts
```diff
@@ -0,0 +1,11 @@
+import { SDK_VERSION_SUPPORT_REGEX } from '../constants';
+import { WindowWithSentry } from '../types';
+
+export function checkBrowserSDKVersion() {
+  const spotlightV1SupportedSentryRegex = new RegExp(SDK_VERSION_SUPPORT_REGEX);
```
Lms24, line 5: l: I'd rather just inline the regex here instead of initializing a new `RegExp` with the constant. Makes things a bit simpler and as far as I can tell, we're not reusing the regex.

packages/website/src/components/InstallCommand.mdx
```diff
@@ -7,17 +7,17 @@ import { Tabs, TabItem } from '@astrojs/starlight/components';
 <Tabs>
   <TabItem label="npm">
   ```sh 
-  npm add @spotlightjs/spotlight
+  npm i @spotlightjs/spotlight --save-dev
```
Lms24, line 10: good catch!

Lms24 (commented):
Good idea! heads-up, I'm gonna work soon on the v2 stable version. Just need to adjust the `window.__SENTRY__` code a bit to get it working with https://github.com/getsentry/sentry-javascript/pull/12206 
Anyway, this isn't relevant for this PR as we'll merge it into v1

packages/overlay/src/integrations/sentry/index.ts
Lms24: So basically, do something like:

```
if (checkBrowserSDKVersion()) {
  return _sentryIntegration();
} else {
  return _sdkNotCompatibleSentryIntegration()
}
```

No particular strong feelings about naming or details just wanna keep this somewhat readable :)

Shubhdeep12:
hey @Lms24 
- Updated the code based on review comments.
- Also update windowWithSentry type as well similar to - https://github.com/getsentry/spotlight/pull/402/files because we are not able to get the version of SDK using just the legacyCarrier for v > 8.6.0

packages/overlay/src/integrations/sentry/utils/sdkValidation.ts
Shubhdeep12: yup, sure

Lms24:
@Shubhdeep12 thanks a lot for making the changes! I think this looks good but I just had a revelation: 

The plan for the v2 release is to merge `v2_dev` into `main` once we go stable. Before this merge, we'll branch off a `v1` branch from `main` to which we can theoretically apply patches if we need to create v1 patches in the future. v2 development will continue on `main`.

With this in mind, we don't necessarily want this commit in the history for v2, but only for v1.So I suggest we wait with merging until v2 is stable and only then merge it into the `v1` branch. Does that make sense to you?

I'm targeting Monday for v2 stable by the way so the wait shouldn't be too long 🤞

Lms24:
Yup, that makes sense, as I said, I think the change is a good idea! Also, waiting for stable gives us the additional bonus that people will feel less sceptical about updating to a stable version than to an alpha version.

BYK:
Is this still relevant @Lms24 and @Shubhdeep12?

Lms24:
Yup, still relevant. Should go into `1.x` (feel free to adjust as you see fit!)

BYK:
Is there a strong reason for us to support 1.x branch and do hotfixes for that?

Shubhdeep12:
> Is there a strong reason for us to support 1.x branch and do hotfixes for that?

All new fixes and features will be added to 2.x(main branch) only.

This specific fix is only added for the users using spotlight v 1.x with sentry sdk v >=8.x because there was a breaking change in sentry js sdk v8.x, which spotlight 1.x doesn't support.

BYK:
>This specific fix is only added for the users using spotlight v 1.x with sentry sdk v >=8.x because there was a breaking change in sentry js sdk v8.x, which spotlight 1.x doesn't support.

Understand this part. Asking if there's anyone out there relying on v1.x or a good reason for us to maintain v1.x (otherwise I'll just close this PR and do not support v1)

packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -96,19 +99,36 @@ export default function sentryIntegration(options?: SentryIntegrationOptions) {
...
+    },
+    reset: () => {},
+  } satisfies Integration<Envelope>;
+}
+
+export default function sentryIntegration(options?: SentryIntegrationOptions) {
+  if (checkBrowserSDKVersion()) {
+    return _sentryIntegration(options);
+  } else {
+    warn(`This version of Spotlight supports Sentry browser SDK version less than 8.x.x. Check out Spotlight v2.x.x`);
```
Lms24, line 128:
```suggestion
    warn(`This version of Spotlight only supports Sentry browser SDK versions below 8.x.x. Check out Spotlight v2.x.x`);
```

Lms24:
image

@BYK Seems like v1 is still used by a lot of people. If it's not too much work, I'd vote we merge this and cut a final v1 release to let people know about the need to update to v2 if they use a v8 JS SDK. but I'll let you make the call.

BYK:
Closing as we are not planning to backdate support. No reason to stick with Spotlight 1.x at this point.

PR #402, opened 2024-05-31
feat: Support versioned Sentry carrier object introduced in 8.6.0+

Lms24:
Adjusts the global carrier accessing logic to handle the [versioned carrier change](https://github.com/getsentry/sentry-javascript/pull/12206) in version 8.6.0 of the JS SDKs.

I'll look into setting up some more e2e test apps with different SDK versions to cover the logic a bit better.

### DIFF
--- a/packages/overlay/src/integrations/sentry/index.ts
+++ b/packages/overlay/src/integrations/sentry/index.ts
@@ -96,17 +96,17 @@ export default function sentryIntegration(options?: SentryIntegrationOptions) {
   } satisfies Integration<Envelope>;
 }
 
-type VersionedCarrier = { version: string } & Record<Exclude<string, 'version'>, { acs: V8AsyncContextStrategy }>;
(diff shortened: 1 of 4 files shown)


packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -211,3 +207,39 @@ function addSpotlightIntegrationToSentry(options?: SentryIntegrationOptions) {
...
+    const versionedCarrier = sentryCarrier[sentryCarrier.version];
+    const scope =
+      typeof versionedCarrier?.stack?.getScope === 'function' ? versionedCarrier?.stack?.getScope?.() : undefined;
+    if (typeof scope?.getClient === 'function') {
+      return scope.getClient();
+    }
+  }
+
+  // 8.0.0-8.5.0 way to get the client
+  if (sentryCarrier.stack) {
```
Lms24, line 228: actually, this is incorrect, let me fix 😅

Lms24: Fixed ✅

mydea (approved):

PR #423, opened 2024-06-23
Added Trace Info

Shubhdeep12:
Before opening this PR:

- [] I referenced issues that this PR addresses

Added a Trace Info page for traces.

image
image

added General info and the tags which are present in transactions in Trace.


packages/overlay/src/integrations/sentry/components/developerInfo/EnvelopeDetails.tsx
```diff
@@ -17,7 +17,7 @@ export default function EnvelopeDetails({ data }: { data: { envelope: Envelope;
           <>
             {header.event_id && (
               <>
-                Event Id: <span className="text-primary-500">&mdash;</span> {header.event_id}
+                Event Id <span className="text-primary-500">&mdash;</span> {header.event_id}
```
BYK, line 20: intentional change?

packages/overlay/src/integrations/sentry/tabs/TracesTab.tsx
```diff
@@ -8,6 +8,7 @@ export default function TracesTab() {
     <SentryEventsContextProvider>
       <Routes>
         <Route path="/:traceId/:spanId" element={<TraceDetails />} />
+        <Route path="/:traceId/info" element={<TraceDetails />} />
```
BYK, line 11: Why don't we split `TraceDetails` component into smaller pieces and create `TraceSpanView`, `TraceInfoView` and `TraceDetails` or `TraceView`? It would save us a bunch of conditionals and would make the code more re-usable

BYK (changes requested):

packages/overlay/src/integrations/sentry/components/developerInfo/EnvelopeDetails.tsx
Shubhdeep12: Yes, we have  "-" after label "Event Id". So its redundant.

packages/overlay/src/integrations/sentry/tabs/TracesTab.tsx
Shubhdeep12: Ideally TraceInfo and SpanDetail should be a component instead of a route. Because they are just a drawer above the TraceDetails page.

And yes, we should split the code.

I'll do this splitting rn.

Shubhdeep12: Hi @BYK I'll make this change in next PR.

To answer your query:
- Yes, its doesnt make sense to have 3 diff routes for Trace detail page. I have added the use of the route below.
 ` <Route path="/:traceId/:spanId" element={<TraceDetails />} />`  -  To open Span detail drawer
  `<Route path="/:traceId/info" element={<TraceDetails />} />`   -  To open Trace Info drawer
 `<Route path="/:traceId" element={<TraceDetails />} />`  -  Main Trace detail page
- As mentioned above ideally opening of drawer(span info view and trace info view) should be just a component handled by states and params(if needed).
- Otherwise the Trace detail already have children components added to it.

BYK (approved):

Issue #421, opened 2024-06-21
To add a search capability in trace view


Shubhdeep12:
Reference: https://sentry.io/changelog/search-for-your-traces/

Let's add a basic search functionality in the trace details page to search b/w spans.

PR #424, opened 2024-06-23
Added Search in Trace detail

Shubhdeep12:
Before opening this PR:

#421 
image

[Screencast from 06-23-2024 07:28:07 PM.webm](https://github.com/getsentry/spotlight/assets/43654389/eb5bb8d1-8cd9-4051-ac84-c603b5cb6989)


packages/overlay/src/integrations/sentry/components/traces/spans/SpanItem.tsx
```diff
@@ -44,12 +46,16 @@ const SpanItem = ({
       setSpanNodeWidth(newLeftWidth);
     }
   };
+  const isQueried = query
+    ? span.op?.includes(query) || span.description?.includes(query) || span.span_id.includes(query)
+    : false;
```
BYK, line 51: I'd put the check for description last as that'd be the most expensive check I presume (due to length?)

```suggestion
  const isQueried = query
    ? span.span_id.includes(query) || span.op?.includes(query) || span.description?.includes(query)
    : false;
```

Also not a fan of those `op?` and `description?` optionals but I'm guessing we have to live with it?

packages/overlay/src/integrations/sentry/components/traces/spans/SpanItem.tsx
```diff
@@ -94,7 +101,7 @@ const SpanItem = ({
           </span>
         </div>
         <div
-          className="waterfall"
+          className={classNames('waterfall', isQueried ? '!bg-transparent' : '')}
```
BYK, line 104: Not happy that we had do spread `query` and `isQueried` around so much. I understand that this is kind of the result of us using `classNames` instead of creating specific CSS classes/styled components. Is there a way out of this? :D

packages/overlay/src/integrations/sentry/components/traces/spans/SpanTree.tsx
```diff
@@ -36,6 +38,7 @@ export default function SpanTree({
             totalDuration={totalDuration}
             spanNodeWidth={spanNodeWidth}
             setSpanNodeWidth={setSpanNodeWidth}
+            query={query}
```
BYK, line 41: I'm wondering if this should be passed in the context at this point. Have you considered that?

packages/overlay/src/integrations/sentry/hooks/useDebounce.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react';
+
+export default function useDebounce<T>(value: T, delay: number): T {
```
BYK, line 3: This is definitely _NOT_ proper debouncing. It just delays setting the value. A real debounce would not change anything until the last update + debounce period. It also looks like this code was lifted from https://github.com/uidotdev/usehooks/blob/90fbbb4cc085e74e50c36a62a5759a40c62bb98e/index.js#L239-L253. If that's the case, we should give a reference back there.

Finally, if you want to have proper debouncing this answer seems like the way: https://stackoverflow.com/a/75556586/90297. The one below that also shows "the hook way" with useMemo and proper cancellation as it apparently uses the `debounce` decorater from lodash.

BYK (changes requested):
Mainly rejecting for `debounce` issue. But please take a look at my other comments too.

packages/overlay/src/integrations/sentry/hooks/useDebounce.ts
Shubhdeep12: Hi @BYK 


> https://github.com/uidotdev/usehooks/blob/90fbbb4cc085e74e50c36a62a5759a40c62bb98e/index.js#L239-L253. If that's the case, we should give a reference back there.

No, not lifted from here.



>Finally, if you want to have proper debouncing this answer seems like the way: https://stackoverflow.com/a/75556586/90297. The one below that also shows "the hook way" with useMemo and proper cancellation as it apparently uses the debounce decorater from lodash.

Sure, let me enhance the hook based on the example above. 
But since we are debouncing search. here we will have to maintain 2 states - one for instantly updating the search query in input(this cannot be debounced) and another for passing the search query in children components(this one will be debounced).

packages/overlay/src/integrations/sentry/components/traces/spans/SpanTree.tsx
Shubhdeep12: hmm, yes we can use context here as well.
should i create one or we can fix it as a whole in sentry integration for other states as well in next PR?

packages/overlay/src/integrations/sentry/components/traces/spans/SpanItem.tsx
Shubhdeep12: Yeah, that is because of tailwind CSS.
Although if needed we can pass the classnames and create the styles in .css files.

packages/overlay/src/integrations/sentry/components/traces/spans/SpanItem.tsx
Shubhdeep12: Yes, it can be possible of having spans here without op or description. so we need a check here.

Shubhdeep12: Definitely i'll make the change suggested above

packages/overlay/src/integrations/sentry/components/traces/TraceDetails.tsx
```diff
@@ -62,6 +76,23 @@ export default function TraceDetails() {
...
       </div>
+      <div className="bg-primary-950 text-primary-50 border-primary-600 hover:border-primary-500 relative mx-6 mb-4 mt-2 flex h-auto w-auto gap-2 rounded-md border py-1 pl-4 pr-6 outline-none transition-all">
+        <input
+          type="text"
+          className="text-primary-50 h-auto w-full flex-1 bg-transparent outline-none transition-all"
+          onChange={handleSearch}
+          value={query}
+          placeholder="Search in Trace"
+        />
+        {query && query.length > 0 && (
```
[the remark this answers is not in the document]
Shubhdeep12: 😅

packages/overlay/src/integrations/sentry/components/traces/TraceDetails.tsx
```diff
@@ -32,6 +36,16 @@ export default function TraceDetails() {
   const startTimestamp = trace.start_timestamp;
   const totalDuration = trace.timestamp - startTimestamp;
 
+  function handleSearch(e: FormEvent<HTMLInputElement>) {
+    e.preventDefault();
+    setQuery(e.currentTarget.value);
+  }
+
+  function handleResetSearch(e: MouseEvent<SVGSVGElement>) {
+    e.preventDefault();
```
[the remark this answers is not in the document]
Shubhdeep12: 😅

packages/overlay/src/integrations/sentry/components/traces/TraceDetails.tsx
```diff
@@ -32,6 +36,16 @@ export default function TraceDetails() {
   const startTimestamp = trace.start_timestamp;
   const totalDuration = trace.timestamp - startTimestamp;
 
+  function handleSearch(e: FormEvent<HTMLInputElement>) {
+    e.preventDefault();
```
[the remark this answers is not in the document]
Shubhdeep12: 😅

packages/overlay/src/integrations/sentry/hooks/useDebounce.ts
BYK: >But since we are debouncing search. here we will have to maintain 2 states - one for instantly updating the search query in input(this cannot be debounced) and another for passing the search query in children components(this one will be debounced).

Not against debouncing, on the contrary I like it. It's just the debouncing logic is not right here.

packages/overlay/src/integrations/sentry/components/traces/TraceDetails.tsx
```diff
@@ -13,7 +13,11 @@ export default function TraceDetails() {
   const { traceId, spanId } = useParams();
   const [spanNodeWidth, setSpanNodeWidth] = useState<number>(50);
   const [query, setQuery] = useState<string>('');
-  const debouncedQuery = useDebounce(query, 200);
+  const [debouncedQuery, setDebouncedQuery] = useState<string>('');
```
BYK, line 16: I liked the `useDebounce` hook usage. I guess we can bring it back in the form of `useDebouncedState` or something which simply returns a `debounce`d setter.

packages/overlay/src/integrations/sentry/hooks/useDebounce.ts
```diff
@@ -1,17 +1,27 @@
...
-  }, [value, delay]);
+  let timer: number;
 
-  return debouncedValue;
+  const debounceFunction = (func: T, delayMs: number, ...args: Parameters<T>) => {
+    clearTimeout(timer);
+    timer = window.setTimeout(() => {
+      func(...args);
+    }, delayMs);
+  };
```
BYK, line 19: May also use this instead: https://www.npmjs.com/package/lodash.debounce

BYK (approved):
It's okay but I think your initial pass where `useDebounce` was containing the logic was better. This version feels a bit less polished and usable.

To emphasize: my main issue was the debounce logic not being correct in the initial version. It seems to be fixed in this version.

packages/overlay/src/integrations/sentry/hooks/useDebounce.ts
Shubhdeep12: Yes, even i found that an issue, because we were setting the state again and again in hook, which was not an efficient solution and that would cost us multiple re-renders.

packages/overlay/src/integrations/sentry/components/traces/TraceDetails.tsx
Shubhdeep12: Hi @BYK,

I just want to add a point here. With the current useDebounce hook, we are passing a callback function instead of setting state directly. I believe this approach is more versatile since it allows the hook to be used anywhere in the app for various actions, such as setting state, making API calls, or performing other actions.

What do you think?

packages/overlay/src/integrations/sentry/hooks/useDebounce.ts
Shubhdeep12: Yes this would be the simpler and the best solution here 😅  
what do you think instead of hook, should we use this package?

packages/overlay/src/integrations/sentry/components/traces/spans/SpanTree.tsx
BYK: >should i create one or we can fix it as a whole in sentry integration for other states as well in a separate PR?

Separate PR as I understand there are more things we can move to context from what you say.

packages/overlay/src/integrations/sentry/components/traces/TraceDetails.tsx
BYK: Just going to try using `lodash.debounce` and see how that goes :)

packages/overlay/src/integrations/sentry/hooks/useDebounce.ts
BYK: See my other comment, will try to use this and see how it goes

BYK (approved):

PR #426, opened 2024-06-29
ref: More robust client init code generation

BYK:
The current client init code generation is a bit fragile. It also does not pass the sidecarUrl option to any other potential integrations. This PR fixes that issue and makes the code generation more robust and readable.

### DIFF
--- a/packages/astro/src/snippets.ts
+++ b/packages/astro/src/snippets.ts
@@ -12,21 +12,18 @@ export type ClientInitOptions = {
 const buildClientImport = (importPath: string) => `import * as Spotlight from ${JSON.stringify(importPath)};`;
 
 const buildClientInit = (options: ClientInitOptions) => {
-  const integrationCalls = options.integrationNames
-    ? options.integrationNames.map(i => `Spotlight.${i}()`).join(', ')
-    : `Spotlight.sentry({sidecarUrl: ${options.sidecarUrl ? `'${options.sidecarUrl}'` : undefined}})`;
+  let initOptions = JSON.stringify({
+    showTriggerButton: options.showTriggerButton !== false,
+    injectImmediately: options.injectImmediately,
+    debug: options.debug,
+    sidecarUrl: options.sidecarUrl,
+  });
 
-  return `
-Spotlight.init({
-  integrations: [
-    ${integrationCalls}
-  ],
-  showTriggerButton: ${options.showTriggerButton === false ? 'false' : 'true'},
-  injectImmediately: ${options.injectImmediately === true ? 'true' : 'false'},
-  debug: ${options.debug === true ? 'true' : 'false'},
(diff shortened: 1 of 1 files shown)


packages/astro/src/snippets.ts
```diff
@@ -12,21 +12,18 @@ export type ClientInitOptions = {
...
-    ${integrationCalls}
-  ],
-  showTriggerButton: ${options.showTriggerButton === false ? 'false' : 'true'},
-  injectImmediately: ${options.injectImmediately === true ? 'true' : 'false'},
-  debug: ${options.debug === true ? 'true' : 'false'},
-  ${options.sidecarUrl ? `sidecarUrl: '${options.sidecarUrl}'` : ''}
-});
-`;
+  const integrationOptions = JSON.stringify({ sidecarUrl: options.sidecarUrl });
+  const integrations = (options.integrationNames || ['sentry']).map(i => `Spotlight.${i}(${integrationOptions})`);
```
Shubhdeep12, line 23: sidecarUrl is only needed in sentry integration. 
if sidecarUrl is passed while initiating Spotlight. is should only be passed into Sentry integration

Shubhdeep12 (changes requested):

packages/astro/src/snippets.ts
Shubhdeep12: Sure.

We know that the spotlight is comprised of 2 packages 
- sidecar (local server which will receive and store the events from sentry SDK)
- overlay (ui for spotlight)


Overlay again is comprised of integrations.
There are 2-3 integrations that overlay provide. The user can build and add custom integration as well.

Out of integrations which overlay provides -  **_sentry_** is one of the integration which overlay add by default.

This sentry integration receives the custom events from sidecar. And gives us the tabs like Errors, Traces, Developer info, etc.

Right now, in Spotlight.init function, we add a sidecar url, which automatically gets added to default sentry integration.

Let me know in case of any follow up query. 
Thanks

dcramer: I'm ok with this change (and we should document it) as the sidecar is intended to be a generic proxy relay. That means any backend service could effectively utilize the sidecar's channels and build a widget from it inside of Spotlight.

BYK: @Shubhdeep12 just discussed with @dcramer. The idea is sidecar being open to _any_ integration so although no other integration _needs_ it, we'll just pass it in by default to allow other integrations to make use of the sidecar.

Shubhdeep12: Ok cool.

Shubhdeep12 (approved):
Creating a separate issue for
- updating docs for custom intrgrations.
- to make sidecarURL (optional) in other intrgrations provided by Spotlight.
- to pass tye sidecarURL from Spotlight.init to the integration's options if the integrations array is paased by props while initiating Spotlight.

PR #434, opened 2024-07-05
feat: Experimental Vite plugin for auto injecting Spotlight

BYK:
This patch creates a new Vite plugin under the main `@spotlightjs/spotlight` package that automatically injects spotlight for dev mode. It also replaces Vite's error page shown on compilation errors with Spotlight.


demos/sveltekit/vite.config.ts
```diff
@@ -1,9 +1,10 @@
 import { sentrySvelteKit } from '@sentry/sveltekit';
 import { sveltekit } from '@sveltejs/kit/vite';
 import { defineConfig } from 'vitest/config';
+import spotlight from '@spotlightjs/spotlight/vite-plugin';
 
 export default defineConfig({
-	plugins: [sveltekit(), sentrySvelteKit({ autoUploadSourceMaps: false })],
+	plugins: [sveltekit(), sentrySvelteKit({ autoUploadSourceMaps: false }), spotlight()],
```
BYK, line 7: @dcramer this is how you use the plugin. Just import and enable for any project. For Astro, we still need `@spotlightjs/astro` for the tighter integration.

packages/spotlight/src/vite-plugin.ts
```diff
@@ -0,0 +1,260 @@
...
+    res.writeHead(200, { 'Content-Type': 'application/json' });
+    res.end(responseJson);
+  });
+};
+
+export default function spotlight({ port }: { port?: number } = {}): PluginOption {
+  Sentry.init({
+    tracesSampleRate: 0,
+    spotlight: true,
+  });
```
Lms24, line 195: Why do we init the SDK here? Shouldn't spotlight be decoupled from the SDK (might have missed some decisions in the last weeks)? Most spotlight users will use the `@sentry/astro` SDK anyway

packages/spotlight/src/vite-plugin.ts
```diff
@@ -0,0 +1,260 @@
...
+    name: 'spotlight',
+    apply: 'serve',
+    transform(code: string, id: string /*, _opts = {}*/) {
+      // Only modify Vite's special client script
+      // The magic value below comes from vite/constants:CLIENT_ENTRY
+      if (!id.endsWith('vite/dist/client/client.mjs')) return;
+
+      const initSnippet = [
+        'createErrorOverlay=()=>{}',
+        `import * as Spotlight from '/@fs${spotlightPath}'`,
```
Lms24, line 209: are we missing a '/' here? Just asking because in `snippets.ts` we have the additional slash

```suggestion
        `import * as Spotlight from '/@fs/${spotlightPath}'`,
```

Lms24 (commented):
I like the idea of using spotlight via a vite plugin! Had two questions but LGTM generally!

(Hope you don't mind me reviewing this; I just saw your message in #discuss-spotlight and wanted to check out what we're working on :D)

BYK:
@Lms24 
>(Hope you don't mind me reviewing this; I just saw your message in #discuss-spotlight and wanted to check out what we're working on :D)

On the contrary! I'm actively looking for feedback so your time and feedback here is much appreciated.

packages/spotlight/src/vite-plugin.ts
BYK: Ah, good catch! I used to use `path.resolve()` when getting `spotlightPath` and then refactored the code a bit but forgot this part (and obviously didn't test this properly).

packages/spotlight/src/vite-plugin.ts
BYK: This is on the compiler side and since this is a generic `vite` plugin, we are not guaranteed to have a Sentry SDK, especially on the compiler suite. The main reason is to be able to relay the compiler errors to Spotlight for now. I think I can hack something together to remove the need for a full SDK tbh.

packages/spotlight/src/vite-plugin.ts
BYK: Scratch that, we get a full URL and that's why it works.

packages/spotlight/src/vite-plugin.ts
BYK: Done, thanks a lot for the push again :)

packages/spotlight/src/vite-plugin.ts
```diff
@@ -0,0 +1,293 @@
...
+        'Spotlight.init({injectImmediately: true, integrations: [Spotlight.sentry({openLastError: enableOverlay})]})',
```
Shubhdeep12, line 264: Hi @BYK correct me if i m wrong here.
this code snippet will be responsible of injecting spotlight overlay?
if yes, then ideally we should allow user to pass any overlay init options, right?

And, for SSR errors, we have the logic in spotlight/astro to display fullpage spotlight.

Shubhdeep12:
Should we upgrade vite to v5?
right now its v4.5.3

packages/spotlight/src/vite-plugin.ts
BYK: >if yes, then ideally we should allow user to pass any overlay init options, right?

Yes it is. I was torn between providing options or just making it a straightforward, seamless experience. Is there an option you think we definitely should allow changing here?

>And, for SSR errors, we have the logic in spotlight/astro to display fullpage spotlight.

Yup, I'll check that out good pointer.

BYK:
>Should we upgrade vite to v5? right now its v4.5.3

Yup, I'm aware. I'm testing this with a different project using v5 and wanted to make sure it works well on both. I may add another demo project for that which uses v5 (or upgrade this one to v5 and use the new one for the v4).

packages/spotlight/src/vite-plugin.ts
Shubhdeep12: > Yes it is. I was torn between providing options or just making it a straightforward, seamless experience. Is there an option you think we definitely should allow changing here?

I think we can give every option, because projects using this vite plugin, will rely on the config here only for spotlight.

packages/astro/src/index.ts
```diff
@@ -74,27 +62,11 @@ const createPlugin = (options?: SpotlightAstroIntegrationOptions): AstroIntegrat
...
-        const { setupSidecar } = await import('@spotlightjs/sidecar');
-        setupSidecar({ logger });
       },
     },
   };
 };
 
 export default createPlugin;
-
-export * from '@spotlightjs/overlay';
```
Lms24, line 100: Did we check if removing this export actually actually removes exports form the pacakge? If so that would be a breaking change and we shouldn't do it in a minor.

packages/astro/src/index.ts
```diff
@@ -36,15 +32,7 @@ const createPlugin = (options?: SpotlightAstroIntegrationOptions): AstroIntegrat
...
-          // Importing this plugin dynamically because for some reason, the top level import
-          // caused a client error because the source-map library code was bundled into the client
-          const { sourceContextPlugin } = await import('./vite/source-context');
-
-          config.vite.plugins = [
-            errorPageInjectionPlugin({ importPath: thisFilePath }),
-            sourceContextPlugin(),
-            ...(config.vite.plugins || []),
-          ];
+          config.vite.plugins = [spotlight({ showTriggerButton: false }), ...(config.vite.plugins || [])];
```
Lms24, line 35: No objections here just wanted to confirm that the previous behaviour of `source-map` code ending up in client bundles is no longer an issue?

packages/sidecar/src/main.ts
```diff
@@ -1,7 +1,7 @@
-import { createWriteStream, readFile } from 'fs';
-import { IncomingMessage, Server, ServerResponse, createServer, get } from 'http';
-import { extname, join } from 'path';
-import { createGunzip, createInflate } from 'zlib';
+import { createWriteStream, readFile } from 'node:fs';
+import { IncomingMessage, Server, ServerResponse, createServer, get } from 'node:http';
+import { extname, join } from 'node:path';
+import { createGunzip, createInflate } from 'node:zlib';
```
Lms24, line 4: No objections to this but I think this is a good reminder that we haven't set the `engines` property in the `package.json`. So technically, we don't set a minimum node version. IIRC the `node:` prefixes were introduced with `14.18.0` so I'm fine with changing this now, given the age of Node 14. Setting `engines` doesn't have to happen in this PR but I think we should set it.

packages/sidecar/package.json
```diff
@@ -40,6 +40,9 @@
     "typescript": "^5.0.2",
     "vite": "^4.5.3"
   },
+  "optionalDependencies": {
+    "@sentry/node": "^8.14.0"
+  },
```
Lms24, line 45: why do we need this? I don't see `@sentry/node` being used anywhere

packages/sidecar/vite.config.ts
```diff
@@ -14,6 +14,9 @@ export default defineConfig({
     ssr: './src/main.ts',
     outDir: './dist',
     sourcemap: true,
+    rollupOptions: {
+      external: [...dependencies, 'node:path', 'node:http', 'node:fs', 'node:zlib'],
```
Lms24, line 18: l: Instead of passing the explicit list here (and risking other builtins being bundled), we could use the [`builtinModules` API](https://nodejs.org/api/module.html#modulebuiltinmodules). We do this in the JS SDK repo. WDYT?

packages/spotlight/vite.config.ts
```diff
@@ -14,12 +15,13 @@ export default defineConfig({
...
         overlay: resolve(__dirname, 'src/overlay.ts'),
         sidecar: resolve(__dirname, 'src/sidecar.ts'),
+        'vite-plugin': resolve(__dirname, 'src/vite-plugin.ts'),
       },
       // the proper extensions will be added
       //   fileName: 'sentry-spotlight',
     },
     rollupOptions: {
-      external: dependencies,
+      external: [...dependencies, 'node:path', 'node:crypto'],
```
Lms24, line 24: l: same thought around `builtinModules` here

Lms24 (commented):
Thanks for addressing my previous review and sorry for the delay!

I still think this is a good idea. Only had some concerns around potential API breakage but I might have missed something. Other than that just some suggestions and minor things.

(sorry for the mutiple individual reviews; I rarely use the VSCode review extension but if I do I end up constantly switching between the "add to review" and "leave comment" button 😅 )

packages/sidecar/src/main.ts
Lms24: source: https://nodejs.org/api/esm.html#node-imports

packages/spotlight/src/vite-plugin.ts
Lms24: thanks for resolving this!

packages/astro/src/index.ts
BYK: I think it just allows people to use `@spotlightjs/astro` as if it was `@spotlightjs/spotlight` or `@spotlightjs/overlay` but with added functionality. So I'd not remove it.

packages/astro/src/index.ts
```diff
@@ -18,11 +16,9 @@ type AstroConfigWithExperimentalDevOverlay = AstroConfig & {
   };
 };
 
-const createPlugin = (options?: SpotlightAstroIntegrationOptions): AstroIntegration => {
```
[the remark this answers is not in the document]
BYK: Yeah they are identical or compatible. Mostly a name change.

packages/astro/src/index.ts
BYK: Good question, and I don't know. I'll check.

packages/sidecar/package.json
BYK: Oh we don't. Leftover from earlier iterations. Great catch, thanks!

packages/sidecar/src/main.ts
BYK: Sounds good, thanks a lot! I think we are already Node v18+ or something?

packages/sidecar/vite.config.ts
BYK: Happy to give it a look, but in a follow up as this pattern is prevalent in the spotlight repo.

packages/spotlight/vite.config.ts
BYK: Same as above, will definitely give it a look but not in this PR.

Lms24 (approved):
Thanks! All good from my end!

packages/astro/src/index.ts
BYK: Ran `pnpm build` in `demos/astro-playground` and inspected all `.js` files under `dist/client`. I see references to source map URLs but no bundled in source map so I think we are clear.

packages/sidecar/src/main.ts
BYK: #440

packages/sidecar/vite.config.ts
BYK: #441

packages/spotlight/vite.config.ts
BYK: #441

packages/sidecar/src/main.ts
Lms24: The Sentry Node SDK supports Node >= 14.18.0 but we can be more restrictive with Spotlight. Given Node 16 is EOL, 18 sounds fine to me.

Lms24: https://github.com/getsentry/sentry-javascript/blob/eb23dc4a4514b7e495af92fee9ccdfbce1dbdecc/packages/node/package.json#L9-L11

packages/astro/src/index.ts
BYK: This is now verified.

BYK: Nope, it is here and I finally understand what you mean. Will follow up.

PR #439, opened 2024-07-12
fix: Overhaul envelope parsing to be spec compliant

BYK:
We got some parsing errors recently when using gib-potatoe as
our testbed. Turns out our initial parser is a bit naive and
cannot work with certain payloads.

This PR rewrites the parser based on the spec at
https://develop.sentry.dev/sdk/envelopes/#serialization-format

It also adds new test cases where the old parser fails some of
them.

Hoping this parser to be a bit more efficient too.


packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -103,34 +103,51 @@ export default function sentryIntegration(options?: SentryIntegrationOptions) {
...
-  const [rawHeader, ...rawEntries] = data.trim().split(/\n/gm);
-
-  const header = JSON.parse(rawHeader) as Envelope[0];
-  const items: Envelope[1][] = [];
-  for (let i = 0; i < rawEntries.length; i += 2) {
-    // guard both rawEntries[i] and rawEntries[i + 1] are defined and not empty
-    if (!rawEntries[i] || !rawEntries[i + 1]) {
-      continue;
+  let prevCursor = 0;
+  let cursor = data.indexOf('\n');
```
jan-auer, line 115: The last newline in an envelope is always optional, which includes an empty envelope with just its header. This means you should default to to the end of the string here. 

Right now, `indexOf` will return `-1` and the slice below will skip the last character (presumably a `}`), which will fail parsing. 

Test with `const data = '{}'`, which should succeed and give you a completely empty envelope.

In items below you handle this correctly.

packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -103,34 +103,51 @@ export default function sentryIntegration(options?: SentryIntegrationOptions) {
...
+  const envelopeHeader = JSON.parse(data.slice(prevCursor, cursor).toString()) as Envelope[0];
```
jan-auer, line 116: If you get a buffer in, shouldn't you specify encoding when converting to string?

packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -103,34 +103,51 @@ export default function sentryIntegration(options?: SentryIntegrationOptions) {
...
+  const items: EnvelopeItem[] = [];
+  while (cursor < data.length - 1) {
+    prevCursor = cursor + 1;
+    cursor = data.indexOf('\n', prevCursor);
+    const itemHeaderLine = data.slice(prevCursor, cursor);
+    const itemHeader = JSON.parse(itemHeaderLine.toString()) as EnvelopeItem[0];
+    prevCursor = cursor + 1;
+    const payloadLength = itemHeader.length;
+    if (payloadLength !== undefined) {
+      cursor += payloadLength + 1;
```
jan-auer, line 127: `cursor` points to the character before the payload start. `+ payloadLength` points it at the last character of the payload (except for empty, but irrelevant) and `+ 1` points it at the newline after payload contents. This is expected going into the next iteration of the loop.

IIUC, in the other branch `cursor` points at the character **after** the the newline, in which case the next item headers would fail to parse. Could you add a test case that contains an item without length, followed by another item?

PR #454, opened 2024-07-24
feat: Move contextlines code to sidecar

### LINKED ISSUE #449: Move contextlines endpoint to sidecar
The contextlines functionality is very useful but right now it is stuck in the Vite plugin. Moving this to the sidecar would make it available for all platforms, including backend SDKs which may not provide contextlines.

BYK:
This change moves the contextlines provider to the sidecar from the Vite plugin opening it up to all SDKs and integrations, not just Vite-based ones. This also means we do not clobber Vite dev server's URL-space (not that it was a problem, but alas).

This required adding a setter to `SentryDataCache` class as it was used as a singleton from its module and handles the additional context line fetching per earlier PRs. It would be good to refactor this usage to make it more contained/encapsulated.

Closes #449.

### DIFF
--- a/.changeset/great-chairs-grab.md
+++ b/.changeset/great-chairs-grab.md
@@ -1,0 +1,7 @@
+---
+'@spotlightjs/spotlight': minor
+'@spotlightjs/overlay': minor
+'@spotlightjs/sidecar': minor
+---
+
+Move contextlines provider to sidecar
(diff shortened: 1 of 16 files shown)


dcramer (approved):
Didn't review in detail but general change makes sense - not sure if we'll be able to keep this kind of behavior if Sidecar runs remotely though (which is fine)

BYK:
Aware of the dependency on being local. Will look into that later.

PR #462, opened 2024-07-26
feat: Add 'open in editor' icon

### LINKED ISSUE #459: Add click to open in editor functionality
We should make each error line / context line clickable somewhere which then opens it in the default editor.

BYK:
Closes #459.

Adds a little pencil icon that opens the file in the editor when clicked. This is done on the sidecar side.

PS: Need to document `LAUNCH_EDITOR` env variable. We use `EDITOR` by default and if it is `vim` or something terminal based, it is opened on the terminal that runs the sidecar, which becomes confusing at times.

image


dcramer:
needs a flex items-center on the parent container so everythings vertically centered :)

dcramer (approved):

packages/sidecar/src/main.ts
```diff
@@ -209,6 +210,34 @@ function handleClearRequest(req: IncomingMessage, res: ServerResponse): void {
   }
 }
 
+function openRequestHandler(req: IncomingMessage, res: ServerResponse) {
+  // We're only interested in handling a POST request
+  if (req.method !== 'POST') {
```
Shubhdeep12, line 215: lets add support for OPTIONS as well.

packages/sidecar/src/main.ts
```diff
@@ -209,6 +210,34 @@ function handleClearRequest(req: IncomingMessage, res: ServerResponse): void {
...
+  let requestBody = '';
+  req.on('data', chunk => {
+    requestBody += chunk;
+  });
+
+  req.on('end', () => {
+    launchEditor(
+      // filename:line:column
+      // both line and column are optional
+      requestBody,
```
Shubhdeep12, line 230: I think this is not working.

As in stacktrace we didn't get the absolute path of the file.
For example: 
In my usecase i am getting filename from sentry sdk as "components/Navbar.tsx"

And we are sending the same filename with line and colno to launchEditor and it is not opening the code editor.

Ideally launchEditor needs absolute path - which is my case is "/home/newuser/Documents/shubhdeepchhabra.in/components/Navbar.tsx"

let me know if am wrong somewhere.

packages/overlay/src/integrations/sentry/components/events/error/Frame.tsx
```diff
@@ -44,26 +49,39 @@ export default function Frame({
...
-        ) : frame.module ? (
-          <span className="text-primary-100">{`${formatFilename(frame.module)} in `}</span>
+        {fileName ? (
+          <span className="text-primary-100">
+            {formatFilename(fileName)}
+            {isOpen && !frame.filename?.includes(':') ? (
+              <PenIcon
+                width={14}
+                height={14}
+                title="Open in editor"
```
Shubhdeep12, line 69: in order to make this title work.

lets update `svgr()` to 
```
svgr({
  svgrOptions: {
     titleProp: true,
  },
 }),
```

in `packages/overlay/vite.config.ts` under plugins

Shubhdeep12 (changes requested):
Hi @BYK added some suggestions and a query.

packages/overlay/src/integrations/sentry/components/events/error/Frame.tsx
```diff
@@ -44,26 +49,39 @@ export default function Frame({
...
-          <span className="text-primary-100">{`${formatFilename(frame.filename)} in `}</span>
-        ) : frame.module ? (
-          <span className="text-primary-100">{`${formatFilename(frame.module)} in `}</span>
+        {fileName ? (
+          <span className="text-primary-100">
+            {formatFilename(fileName)}
+            {isOpen && !frame.filename?.includes(':') ? (
+              <PenIcon
+                width={14}
+                height={14}
```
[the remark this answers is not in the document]
Shubhdeep12: image

packages/sidecar/src/main.ts
```diff
@@ -209,6 +210,34 @@ function handleClearRequest(req: IncomingMessage, res: ServerResponse): void {
...
+      // callback if failed to launch (optional)
+      (fileName: string, errorMsg: string) => {
+        logger.error(`Failed to launch editor for ${fileName}: ${errorMsg}`);
+      },
+    );
+    res.writeHead(204);
```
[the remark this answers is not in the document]
BYK: Ah, right! I meant to do it and then forgot. Thanks a lot for catching this!

packages/sidecar/src/main.ts
BYK: So I was testing this with the built-in Astro project with `pnpm dev:playground`. There, the stacktrace filename gets converted into an absolute path through the `contextlines` endpoint and that's why this worked for me locally.

I now added logic to `resolve` relative paths based on a provided `basePath` like the static file serve handler or relative to where the process was started which should be the project base path ideally.

If you can provide me a repro case, I'd love to make this more robust.

Shubhdeep12: Yeah, we can do this. 

- But basePath is optional and undefined by default unless user pass it explicitly.
  - there are 3-4 ways to run sidecar server, we need to handle all of them.
- And, i think this will also cause issue -> rn sidecar only run once on a machine and handle multiple instances. For example -  i can run multiple spotlight overlay instances with single sidecar instance. how we would get diff basepath.

BYK: >But basePath is optional and undefined by default unless user pass it explicitly. there are 3-4 ways to run sidecar server, we need to handle all of them.

I think the default of `process.cwd()` covers most cases at the moment.

>And, i think this will also cause issue -> rn sidecar only run once on a machine and handle multiple instances. For example - i can run multiple spotlight overlay instances with single sidecar instance. how we would get diff basepath.

Agreed. This should be handled from the front-end/overlay side tho, ensuring absolute URLs if/when possible. This may not always be available to us but I'd rather deal with this as it comes through user-feedback.

Shubhdeep12: Yes agreed!
we can go ahead now with default value

Shubhdeep12 (approved):

Shubhdeep12:
Hey @BYK sorry, missed the auto-merge update and i approved the PR..
we still have to add that default value, right?

BYK:
@Shubhdeep12 
>we still have to add that default value, right?

It's already in. Right here: https://github.com/getsentry/spotlight/pull/462/files#diff-3d272de828f072bb4c96f6d45c9506aca3614dd09b21b4188e5223e916e5cfd3R215

PR #469, opened 2024-08-01
feat: Add `#debug` flag to standalone spotlight web UI

BYK:
This enables the `debug` flag on Spotlight init for debugging.


Lms24 (approved):
Just to confirm: This HTML file is the one that's served from the sidecar for a standalone spotlight window/tab, right?

The change looks reasonable to me!

BYK:
Yup, this is the standalone page :)

PR #470, opened 2024-08-01
fix: Fix logical mistakes in integration event processing

BYK:
We had multiple logical mistakes, such as not matching on lower-cased content type headers or skipping the entire event handler when a single integration did not define an event processor. This PR fixes these logical issues.


Lms24 (approved):
Good catch!

Issue #403, opened 2024-05-31
Export `SpotlightBrowser` Sentry integration


Lms24:
If users run spotlight standalone/with the electron app, the current automatic `SpotlightBrowser` injection logic won't work. I propose we export the `spotlightBrowserIntegration` from the `@spotlightjs/overlay` and `@spotlightjs/spotlight` packages so that users can register the integration themselves in their browser apps and don't have to rely on the auto injection mechanism.

Important: We'll keep auto injection as-is (meaning it's on by default but can be deactivated). This is just an additional way to get events from the browser into spotlight.

```[tasklist]
### Tasks
- [ ] Export `spotlightBrowserIntegration`
- [ ] Add documentation
```

Shubhdeep12:
Hi @Lms24 how about just adding a attribute(client) in sentryIntegration along with options.

i think then user can pass the BrowserClient while Spotlight.init like this

Spotlight.init({
    debug: true,
    integrations: [sentry({ client: Sentry.getClient() })],
})

Shubhdeep12:
let me know if this sounds good. i'll push a code then

Shubhdeep12:
hey @Lms24 
i think here we are concerned of the scenario when user is using spotlight in app with browser sdk but dont want overalay on app and is using electron app.

how about instead of exporting sentryBrowserIntegration, we add a `client` attribute to current sentry integration(which is already exported from spotlight)?

User can send Sentry.client() in it and spotlight  gets injected in that client, otherwise default way of injecting.

it will be something like - 

```
Spotlight.init({
	debug: true,
	integrations: [sentry({ client: Sentry.getClient() })],
})
```

Lms24:
This would be a possibility but unfortunately, it only fixes half the problem. For example, take the Spotlight Electron app. If people build a frontend application but want to use Spotlight with the app instead of the built-in overlay, there is no overlay that's injected into the frontend application. Meaning we couldn't pass a client to the electron app. We can however export the spotlightBrowserIntegration from `@spotlightjs/spotlight` which users can then register in their `Sentry.init` function in the frontend application. 

Does this sound reasonable to you?

Shubhdeep12:
[...]

PR #471, opened 2024-08-01
feat: Auto-correct Sentry Browser SDK content type

BYK:
This PR automatically corrects Sentry Browser SDK events' content type as it avoids using the standard `application/x-sentry-envelope` for CORS reasons. This also comes with the side effect of fixing URL parsing when it contains a query string or when a content-type header is not set.

### DIFF
--- a/packages/sidecar/src/main.ts
+++ b/packages/sidecar/src/main.ts
@@ -44,7 +44,12 @@ type SideCarOptions = {
   incomingPayload?: IncomingPayloadCallback;
 };
 
-type RequestHandler = (req: IncomingMessage, res: ServerResponse) => void;
+type RequestHandler = (
+  req: IncomingMessage,
+  res: ServerResponse,
+  pathname?: string,
+  searchParams?: URLSearchParams,
+) => void;
 
 function getCorsHeader(): { [name: string]: string } {
   return {
@@ -56,7 +61,12 @@ function getCorsHeader(): { [name: string]: string } {
 }
 
 function enableCORS(handler: RequestHandler): RequestHandler {
-  return function corsMiddleware(req: IncomingMessage, res: ServerResponse) {
+  return function corsMiddleware(
+    req: IncomingMessage,
+    res: ServerResponse,
+    pathname?: string,
+    searchParams?: URLSearchParams,
+  ) {
     const headers = {
       ...getCorsHeader(),
       ...getSpotlightHeader(),
@@ -71,7 +81,7 @@ function enableCORS(handler: RequestHandler): RequestHandler {
       res.end();
       return;
     }
-    return handler(req, res);
+    return handler(req, res, pathname, searchParams);
   };
 }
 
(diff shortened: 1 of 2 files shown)


packages/sidecar/src/main.ts
```diff
@@ -132,12 +152,20 @@ function streamRequestHandler(buffer: MessageBuffer<Payload>, incomingPayload?:
       });
 
       stream.on('end', () => {
-        buffer.put([`${req.headers['content-type']}`, body]);
+        let contentType = req.headers['content-type']?.split(';')[0].toLocaleLowerCase();
+        if (searchParams?.get('sentry_client')?.startsWith('sentry.javascript.browser') && req.headers.origin) {
+          // This is a correction we make as Sentry Browser SDK may send messages with text/plain to avoid CORS issues
+          contentType = 'application/x-sentry-envelope';
```
Lms24, line 158: out of curiousity: What kind of problem did this previously cause?

BYK: I've been trying to use the standalone mode for end to end tracing in a Rails app. Since Spotlight was not on the same page, the Sentry integration did not work properly. I then decided to use the hacked DSN approach to force enable the browser SDK which sent the events with `text/plain;charset=UTF-8` content type. Sidecar simply passed this down to the overlay. The Sentry integration there only listens to the envelope content type so doesn't even get triggered. I tried adding text/plain but that didn't work either way we didn't strip the charset part or do case normalization. It also felt wrong to hook onto text/plain. Hence this solution 😅

Lms24: Ahh this makes a lot of sense. Reminds me that we should export the sentry integration from the `@spotlight/spotlight`(?) package so that users can register it in their browser SDK setup. https://github.com/getsentry/spotlight/issues/403

BYK: Was gonna look into that but I also noticed there's already a Spotlight integration in the node SDK: https://github.com/getsentry/sentry-javascript/blob/develop/packages/node/src/integrations/spotlight.ts#L6

Maybe just move that into common, enable when `spotlight: true` is passed to init and remove the one from Spotlight? I feel like that's the better approach. WDYT?

Lms24: We can't do this in the browser because this means the spotlight integration would be part of the bundle. Bundlers can't tree shake the integration if we control adding it with a runtime flag like `spotlight: true`.

So for browser SDKs, users will always need to register the integration in `Sentry.init` (like `browserTracingIntegration` today). 

Also I don't think we can easily hoist the same integration into `core` because the sending mechanism (node `http` vs `fetch`) will differ.

Lms24: RE exporting it from `@sentry/browser` vs `@spotlightjs/spotlight` I don't particularly care. We can also export it from the SDKs, as long as we keep things opt-in and tree-shakeable.

PR #478, opened 2024-08-06
feat: Add IIFE bundle for single script include

BYK:
This change adds an Immediately Invoked Function Expression (IIFE) bundle which automatically initializes Spotlight upon loading. Greatly simplifies the HTML include method for Spotlight.

It also fixes a bug where Spotlight won't be initialized if it loads after page `load` event and `injectImmediately` was not set to `true`.


Lms24 (approved):
Looks fine to me from a technical perspective!

Issue #133, opened 2023-11-24
Feature: Spotlight just for Frontend


HazAT:
Make Spotlight a mode that, if it fails to connect to the sidecar from the frontend and can try to directly send stuff from the Sentry Browser SDK to the Overlay.
For pure frontends, this would make it possible to run it without the sidecar.

lforst:
A really cool architecture here would be to have a webworker that intercepts all the requests from the browser and acts as a sidecar, until it detects an actual sidecar and then forwards/proxies everything.

Lms24:
Spotlight overlay already has an event target we can use to communicate. Maybe we send events via the target to the sentry datacache.

BYK:
Closing this out as this is now a reality. The [drop-in IIFE CDN bundle](https://spotlightjs.com/setup/html/) works especially well in this scenario.

PR #498, opened 2024-08-20
feat(overlay): Allow to directly add sentry envelopes via trigger

mydea:
I am currently working on a hackweek project where I want to integrate the spotlight overlay into a chrome browser extension.

There, I cannot use sidecar (because I cannot launch a node service), but I can intercept envelopes sent to sentry myself. I'd like to be able to just run the spotlight overlay in the extension and directly push envelopes to it.

For this, I added a new hook, so you can add envelopes like this:

```
trigger("sentry:addEnvelope", sentryEnvelope);
```

### DIFF
--- a/packages/overlay/src/integrations/sentry/index.ts
+++ b/packages/overlay/src/integrations/sentry/index.ts
@@ -50,7 +50,7 @@ export default function sentryIntegration(options: SentryIntegrationOptions = {}
 
       on('sentry:showError', onRenderError as EventListener);
 
-      const onAddEnvelope = (e: CustomEvent<RawEventContext>) => {
+      const onAddEnvelope = (e: CustomEvent<string | Buffer>) => {
         if (!e.detail) return;
         processEnvelope({
           contentType: HEADER,
--- a/packages/overlay/src/integrations/sentry/index.ts
+++ b/packages/overlay/src/integrations/sentry/index.ts
@@ -52,7 +52,7 @@ export default function sentryIntegration(options: SentryIntegrationOptions = {}
       const onAddEnvelope = (e: CustomEvent) => {
         if (!e.detail.envelope) return;
         processEnvelope({
-          contentType: 'application/x-sentry-envelope',
+          contentType: HEADER,
           data: e.detail.envelope,
         });
       };
--- a/packages/overlay/src/integrations/sentry/index.ts
+++ b/packages/overlay/src/integrations/sentry/index.ts
@@ -41,19 +41,20 @@ export default function sentryIntegration(options: SentryIntegrationOptions = {}
         });
       }
 
-      const onRenderError = (e: CustomEvent) => {
-        log('Sentry Event', e.detail.event_id);
-        if (!e.detail.event) return;
-        sentryDataCache.pushEvent(e.detail.event).then(() => open(`/errors/${e.detail.event.event_id}`));
+      const onRenderError = (e: CustomEvent<{ event?: SentryEvent; event_id: string }>) => {
+        const { event_id, event } = e.detail;
+        log('Sentry Event', event_id);
+        if (!event) return;
+        sentryDataCache.pushEvent(event).then(() => open(`/errors/${event.event_id}`));
       };
 
       on('sentry:showError', onRenderError as EventListener);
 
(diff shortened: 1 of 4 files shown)


Lms24 (approved):

BYK:
Nice, this looks something like #133 I guess.

packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -49,8 +49,19 @@ export default function sentryIntegration(options: SentryIntegrationOptions = {}
 
       on('sentry:showError', onRenderError as EventListener);
 
+      const onAddEnvelope = (e: CustomEvent) => {
+        if (!e.detail.envelope) return;
+        processEnvelope({
+          contentType: 'application/x-sentry-envelope',
```
BYK, line 55:
```suggestion
          contentType: HEADER,
```

packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -49,8 +49,19 @@ export default function sentryIntegration(options: SentryIntegrationOptions = {}
 
       on('sentry:showError', onRenderError as EventListener);
 
+      const onAddEnvelope = (e: CustomEvent) => {
+        if (!e.detail.envelope) return;
+        processEnvelope({
+          contentType: HEADER,
+          data: e.detail.envelope,
```
BYK, line 56: Curious, why not just `e.detail` and pass the envelope as the event payload?

mydea: I basically just copied this from the `showError` scenario - but actually you are right, I think this is incorrect, will adjust it!

packages/overlay/src/integrations/sentry/index.ts
```diff
@@ -41,16 +41,28 @@ export default function sentryIntegration(options: SentryIntegrationOptions = {}
         });
       }
 
-      const onRenderError = (e: CustomEvent) => {
-        log('Sentry Event', e.detail.event_id);
-        if (!e.detail.event) return;
-        sentryDataCache.pushEvent(e.detail.event).then(() => open(`/errors/${e.detail.event.event_id}`));
+      const onRenderError = (e: CustomEvent<{ event?: SentryEvent; event_id: string }>) => {
```
mydea, line 44: While at it I also typed this, for clarity - I think this is correct 😅  cc @BYK

packages/website/src/content/docs/reference/configuration.md
```diff
@@ -155,7 +155,7 @@ import { trigger } from '@spotlightjs/spotlight';
 
 trigger('sentry:showError', {
   event: string,
-  eventId: string,
+  event_id: string,
```
mydea, line 158: I think these docs were incorrect? We use `event_id` in the code, as far as I can see 😅

BYK: Ugh, sorry :(

BYK (changes requested):
Okay, so for this to work properly, we need to do this at a higher level, not inside Sentry integration: https://github.com/getsentry/spotlight/blob/86055ccd24f9980cace6e2286bea131bbccc50de/packages/overlay/src/sidecar.ts#L17

The name `sidecar.ts` is not great but essentially, that `listener` should be triggered which runs all integrations listening on the data and then updates their data with "setIntegrationData()`. This would allow bypassing sidecare much easier for other integrations too as it would provide a generic event ingestion system without the sidecar.

BYK:
Pushed out #506 which should make doing the right thing in this patch much easier. LMK if you want to do it yourself or okay me doing it :)

mydea:
If you can do it, I'd greatly appreciate it - I am rather busy right now so not sure when I can get around to it 😅

BYK:
Should be done in #508. It still expects a full envelope in as a `string` or `Buffer`. Hoping that's a good-enough API? (directly analogous to the sidecar SSE format)

mydea:
> Should be done in #508. It still expects a full envelope in as a `string` or `Buffer`. Hoping that's a good-enough API? (directly analogous to the sidecar SSE format)

That should work just fine I believe, thank you! :)

PR #501, opened 2024-08-21
feat(overlay): Allow to skip sidecar

mydea:
For the frontend-only use-case, we do not have a sidecar, so we want to skip connecting to it & do not show a warning for that.

ref https://github.com/getsentry/spotlight/issues/133


BYK:
This change looks a bit invasive. Do we really need a `skipSidecar` option? Like how about we pass `sidecarUrl: false` or even better, we just say we cannot connect to sidecar but we can run in FE-only mode?

BYK (changes requested):
Blocking till we settle on the way forward. Otherwise changes look OK.

dcramer:
I think we can close this out?

At this point we've decided that we will 1) never skip the sidecar (a variety of reasons), and 2) additionally shortcut the sidecar straight to the embed when possible

mydea:
I can continue to use the forked version for the chrome extension (where a sidecar is not possible), if we do not want to "fully" support this!

BYK:
@mydea I'll explore ways to do this automatically rather than using an explicit flag. That's why I'm keeping the PR open but I don't think we'll be moving forward with this particular PR/code change.

PR #508, opened 2024-08-27
feat(overlay): Add direct event sending w/o sidecar

BYK:
Working towards #133, supersedes #498

### DIFF
--- a/packages/overlay/src/index.tsx
+++ b/packages/overlay/src/index.tsx
@@ -39,35 +39,35 @@ function createStyleSheet(styles: string) {
 /**
  * Open the Spotlight debugger Window
  */
-export function openSpotlight(path?: string | undefined) {
+export async function openSpotlight(path?: string | undefined) {
   trigger('open', { path });
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/overlay/src/index.tsx
```diff
@@ -39,28 +39,35 @@ function createStyleSheet(styles: string) {
 /**
  * Open the Spotlight debugger Window
  */
-export async function openSpotlight(path?: string | undefined) {
```
Lms24, line 42: h: this is publicly exported, right? This is semver-breaking if we change from Promise<void> to void

BYK: Ehh, can revert. Just, there was no reason for these to be async.

Lms24: yeah good point! I mean we can also just cut a major of the overlay package 😅 but it probably makes sense to wait for a change that justifies a major, I'll let you be the judge of that 😅

Lms24 (approved):

PR #509, opened 2024-08-27
feat: Add no-sidecar Sentry SDK integration for overlay

BYK:
Ref #133.

### DIFF
--- a/demos/astro-playground/package.json
+++ b/demos/astro-playground/package.json
@@ -25,7 +25,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "svelte": "^5.0.0-next.1",
-    "typescript": "^5.2.2"
+    "typescript": "^5.4.0"
   },
   "volta": {
     "extends": "../../package.json"
--- a/demos/astro-playground/package.json
+++ b/demos/astro-playground/package.json
@@ -25,7 +25,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "svelte": "^5.0.0-next.1",
-    "typescript": "^5.4.0"
+    "typescript": "^5.2.2"
   },
   "volta": {
     "extends": "../../package.json"
--- a/packages/overlay/package.json
+++ b/packages/overlay/package.json
@@ -42,7 +42,7 @@
     "eslint": "^8.45.0",
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-react-refresh": "^0.4.3",
-    "typescript": "^5.4.0",
+    "typescript": "^5.0.2",
     "vite": "^4.5.3",
     "vite-plugin-dts": "^3.5.2",
     "vite-plugin-svgr": "^3.2.0",
--- a/packages/overlay/package.json
+++ b/packages/overlay/package.json
@@ -42,7 +42,7 @@
     "eslint": "^8.45.0",
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-react-refresh": "^0.4.3",
-    "typescript": "^5.0.2",
+    "typescript": "^5.4.0",
     "vite": "^4.5.3",
     "vite-plugin-dts": "^3.5.2",
     "vite-plugin-svgr": "^3.2.0",
--- a/packages/overlay/src/integrations/sentry/sentry-integration.ts
+++ b/packages/overlay/src/integrations/sentry/sentry-integration.ts
@@ -1,19 +1,8 @@
 import type { Client, Envelope, Event, Integration } from '@sentry/types';
 import { serializeEnvelope } from '@sentry/utils';
-import { DEFAULT_SIDECAR_URL } from '~/constants';
+import { trigger } from '../../lib/eventTarget';
 import { log } from '../../lib/logger';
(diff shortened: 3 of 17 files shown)


BYK:
Okay this works but we need some persistent storage as the events don't survive across page loads or even a simple reload.

BYK:
Seemingly best way forward:

- Use `IndexedDB` with expiry as shown in https://stackoverflow.com/a/35518068/90297
- On initialization (`afterAllSetup`), expire all stored and expired events, send rest to Spotlight
- Store each envelope in `beforeSend` hook before forwarding to Spotlight

Lms24:
Why do we need to care about previously expired events? I'm not yet sure I understand why we'd need to store anything in IndexdDB. Here's what I'm currently thinking:

1. Spotlight Overlay and Sentry SDK (with spotlight integration) intialize on page load
2. Sentry SDK sends event
3. Spotlight Sentry integration's `beforeEnvelope` listener is invoked and sends the event via `trigger('event',...)`
4. Event appears in spotlight
5. When page is reloaded/hard navigation to next step -> back to step 1

Lms24:
whoops, sorry for the accidental close!

packages/overlay/src/integrations/sentry/sentry-integration.ts
```diff
@@ -1,42 +1,27 @@
...
- *
- * @param options - Configuration options for the integration ({@link SpotlightBrowserIntegrationOptions})
+ *    the same page via the "direct message" method (w/o a need for the sidecar)
  *
  * @returns Sentry integration for Spotlight.
  */
-export const spotlightIntegration = (options?: SpotlightBrowserIntegrationOptions) => {
-  const _sidecarUrl = options?.sidecarUrl ?? DEFAULT_SIDECAR_URL;
-
+export const spotlightIntegration = () => {
```
Lms24, line 17: Do we export this from the overlay package? I don't think so but I'm just checking, since this is a breaking change if it was public API.

BYK:
>I guess the only advantage we'd get with storing events is that we can show events from past page loads. Which is probably not always what we/users want but I can see it being helpful. If that's the goal then let's do it but IMO, this is something we should handle within spotlight and independent of if the event came in from the sidecar or from the event target.

This is the current state anyway. Sidecar stores all past events (and without expiry which we may wanna address in the future due to memory concerns), and replays them as the first thing when a connection is made. When a user goes through pages or reloads, the events persist. In my own testing, I was also surprised when they were missing.

Regarding at which level this should be done, I agree it makes sense to move it to Spotlight core so I'll tie this in to the new manual event trigger system.

packages/overlay/src/integrations/sentry/sentry-integration.ts
BYK: Yeah I'm aware. I'm fine as it won't really break if people pass in an options object, it will just be ignored which should be fine in this specific scenario.

demos/astro-playground/package.json
```diff
@@ -25,7 +25,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "svelte": "^5.0.0-next.1",
-    "typescript": "^5.2.2"
+    "typescript": "^5.4.0"
```
BYK, line 28: Had to upgrade TS to get support for `Promise.withResolvers()` which is now maintstream.

Lms24 (commented):
Sounds good to me! Though I'm a bit concerned with the TS bump (see comments). Given 5.4 isn't that old yet, I suggest we think twice if we need to bump it. I think in this case, rewriting `Promise.withResolvers` should be doable. If you still want to bump, we can of course also cut a major.

packages/overlay/src/integrations/sentry/sentry-integration.ts
BYK: Possibly. But again, this is not meant to be used externally so 🤷🏻‍♂️

packages/overlay/src/lib/db.ts
```diff
@@ -0,0 +1,95 @@
...
+export const DB_VERSION = 2;
+
+let _DB: IDBDatabase | null = null;
+
+export function clearDBCache() {
+  _DB = null;
+}
+
+function createDB(): Promise<IDBDatabase> {
+  const { promise, resolve, reject } = Promise.withResolvers();
```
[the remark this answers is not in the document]
BYK: Done

packages/overlay/package.json
```diff
@@ -42,7 +42,7 @@
     "eslint": "^8.45.0",
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-react-refresh": "^0.4.3",
-    "typescript": "^5.0.2",
+    "typescript": "^5.4.0",
```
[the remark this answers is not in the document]
BYK: Reverted. Adding the summary from our private chat:

>so in this context, it was harmless as I assumed, it just opens the door for adding potentially bw-incompatible types

demos/astro-playground/package.json
BYK: Reverted.

Lms24 (approved):
LGTM! Thanks for making the changes!

PR #513, opened 2024-08-30
fix: Stacktraces not reversing when error has full context

### LINKED ISSUE #489: Stack Trace is sometimes upside down in Astro Server-side Events
### Environment

- Astro 4.13.3
- Spotlight 2.1.6
- Node 20.11.0

### Steps to Reproduce
- 
- Throw a server-side error in a astro app during SSR (sorry, don't have a reproducer yet but soon)

or

- take this JSON event: [evt.json](https://github.com/user-attachments/files/16601699/evt.json)

### Expected Result

The stack trace in the Spotlight overlay should show the stack frame where the error was thrown on top as the first frame

### Actual Result

The stack frame where the error was thrown is on the bottom of the stack trace ==> the stack trace is upside down.

Image

(also it's off by two LoC but that might be an Astro problem, not sure)

BYK:
Fixes #489.

### DIFF
--- a/.changeset/stale-jobs-give.md
+++ b/.changeset/stale-jobs-give.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/overlay': patch
+---
+
+Fix stacktraces are not reversed sometimes (appearing in wrong order)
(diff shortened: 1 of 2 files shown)


Lms24 (approved):
thx!

PR #517, opened 2024-09-09
fix: Fix error on autocomplete key presses

### LINKED ISSUE #516: Unexpected error on form autocompletion select
### Environment

I'm on  "@spotlightjs/spotlight": "^2.3.2", and I see this error every time I select an option in an input with a native browser autocompletion.

### Steps to Reproduce

1. Click on a field with an autocompletion
2. Select any option that the browser provides using a mouse
3. See the error in the console and in spotlight

### Expected Result

No errors

### Actual Result

Image

Image

The error happens in this part of the code, at `key.toLowerCase()`
```javascript
function uR(e, t, n = !1) {
  P.useEffect(() => {
    function r(i) {
      n || i.stopPropagation(), e.every(
        (s) => s in i ? i[s] : i.key.toLowerCase() === s.toLowerCase()
      ) && t();
    }
    return window.addEventListener("keyup", r), () => window.removeEventListener("keyup", r);
  }, [e, t, n]);
}
```

Actually, I found the bug [here](https://github.com/getsentry/spotlight/blob/16d4282937ba51453986f9fd2dd05cdde7b9cee8/packages/overlay/src/lib/useKeyPress.ts#L18). I'm wondering why TS is silent, there's no `key` in `e` in the second part of the ternary.
```javascript
keys.every((key: string) =>
  key in e ? e[key as keyof KeyboardEvent] : e.key.toLowerCase() === key.toLowerCase(),
)
```
It should be like this, because there's no key in the event:
```javascript
keys.every((key: string) =>
  key in e && e[key as keyof KeyboardEvent],
)
```

BYK:
Fixes #516.

### DIFF
--- a/packages/overlay/src/lib/useKeyPress.ts
+++ b/packages/overlay/src/lib/useKeyPress.ts
@@ -1,23 +1,28 @@
 import { useEffect } from 'react';
 
+type ModifierKeys = 'altKey' | 'ctrlKey' | 'shiftKey' | 'metaKey';
+
 /**
  * useKeyPress
- * @param {string[]} keys - an array of keys to respond to, compared against event.key
+ * @param {string} key - The letter or name of the key to respond to -- this is normalized to lower case
+ * @param {ModifierKeys[]} modifiers - The modifiers that needs to be activated such as ctrlKey
  * @param {function} action - the action to perform on key press
  * @param {boolean} propagate - whether to stop event propagation (default is false)
  */
-export default function useKeyPress(keys: string[], action: () => void, propagate = false) {
+export default function useKeyPress(key: string, modifiers: ModifierKeys[], action: () => void, propagate = false) {
+  const normalizedKey = key.toLowerCase();
   useEffect(() => {
-    function onKeyup(e: KeyboardEvent) {
+    function onKeyup(evt: KeyboardEvent) {
       if (!propagate) {
-        e.stopPropagation();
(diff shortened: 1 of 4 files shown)


packages/overlay/src/lib/useKeyPress.ts
```diff
@@ -1,29 +1,34 @@
...
   useEffect(() => {
-    function onKeyup(e: KeyboardEvent) {
+    function onKeyup(evt: KeyboardEvent) {
       if (!propagate) {
-        e.stopPropagation();
+        evt.stopPropagation();
+      }
+
+      // We don't respond to modifier-only key presses
+      if (!evt.key) {
```
Shubhdeep12, line 21: Hi @BYK just want to confirm,
Will this be fine or we should only check for modifiers here.
Something like
[Shift, meta, alt, ctrl].includes(evt.key) then return;

BYK: I figured (and decided) it doesn't make much sense to just act on modifiers (hence the comment above) so not sure if that would be useful. If there's no `key` then it is very likely that this is a modifier-only key event or a synthetic event. We check for modifier states below.

Shubhdeep12: Ok cool. I got confused by the comment above.

Shubhdeep12 (approved):
LGMT 🚀🚀

PR #549, opened 2024-11-05
fix: Overlay issues fixed

Shubhdeep12:
Before opening this PR:

- Fixed some conditional rendering in TraceIcon and SpanDetails which was showing 0 on UI.
- Fixed routing for query summary page by encoding the query description which can be a long text.
  - Encoded query desciption and then decoded on query summary page.

### DIFF
--- a/packages/overlay/src/integrations/sentry/components/traces/TraceIcon.tsx
+++ b/packages/overlay/src/integrations/sentry/components/traces/TraceIcon.tsx
@@ -26,7 +26,7 @@ export default function TraceIcon({ trace }: TraceIconProps) {
       {dominantPlatforms.map(platform => (
         <PlatformIcon key={platform} title={platform} size={21} platform={platform} />
       ))}
-      {remainingPlatforms.length && (
+      {remainingPlatforms.length > 0 && (
         <div
           title={remainingPlatforms.join(', ')}
           className="h-[21px] w-[21px] bg-black p-0.5 text-xs font-bold text-white"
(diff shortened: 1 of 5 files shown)


BYK (approved):

packages/overlay/src/integrations/sentry/components/traces/TraceIcon.tsx
```diff
@@ -26,7 +26,7 @@ export default function TraceIcon({ trace }: TraceIconProps) {
       {dominantPlatforms.map(platform => (
         <PlatformIcon key={platform} title={platform} size={21} platform={platform} />
       ))}
-      {remainingPlatforms.length && (
+      {remainingPlatforms.length > 0 && (
```
BYK, line 29: Ah, this is such a nice catch! That said wouldn't it render `false` in this case? Should we not turn this into a ternary with `: null` at the end @Shubhdeep12

Shubhdeep12: its fine with `false`
react treat `false` as `null` or `undefined`

reference: https://react.dev/learn/conditional-rendering#logical-and-operator-:~:text=But%20if%20the%20condition%20is%20false%2C%20the%20whole%20expression%20becomes%20false.%20React%20considers%20false%20as%20a%20%E2%80%9Chole%E2%80%9D%20in%20the%20JSX%20tree%2C%20just%20like%20null%20or%20undefined%2C%20and%20doesn%E2%80%99t%20render%20anything%20in%20its%20place.

Issue #548, opened 2024-11-04
Add click to copy for file names


BYK:
As a user, I should be able to click a button next to the filenames to copy them to my clipboard

BYK:
@Shubhdeep12 yup, somewhere there or close to that "open in editor" button

PR #550, opened 2024-11-05
feat(overlay): added copy filename and changed styling of error frame

Shubhdeep12:
Before opening this PR:

#548 
- Added copy filename button on error frame.
- changed styling of error frames.
image

image

image
image


.changeset/shaggy-dodos-sparkle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@spotlightjs/overlay': patch
```
BYK, line 2: This is a new feature so update is a `minor` not `patch` :)

```suggestion
'@spotlightjs/overlay': minor
```

packages/overlay/src/integrations/sentry/components/events/error/Frame.tsx
```diff
@@ -48,51 +70,44 @@ export default function Frame({
...
+          {fileName ? (
             <span className="text-primary-100">
-              {frame.lineno}
-              {frame.colno !== undefined && `:${frame.colno}`}
+              {formatFilename(fileName)}
+              {' in '}
             </span>
-          </>
-        )}
-        {isOpen && !frame.filename?.includes(':') ? (
```
Shubhdeep12, line 80: looks like we missed this condition `!frame.filename?.includes(':')`

BYK: This was intentional but we can revisit the decision in the future if necessary.

PR #553, opened 2024-11-06
feat(spotlight): Stabilize embedded JS filename for external use

### LINKED ISSUE #552: Make the entrypoint JS usable by other pages
Currently we have the CDN-hosted IIFE entrypoint for FE-only workflows. But if one wanted to inject the Spotlight UI to a server-side framework such as Django or Rails, there's no easy way to just include the JS file as the filename is hash-versioned. We should make the name stable as there's no use in hashing the filename when shipping an already-versioned package.

Moreover, this entrypoint assumes fullscreen Spotlight, which we should find a good solution for the scenario described above.

BYK:
Fixes #552.

With this change one can now include Spotlight on any page locally from `http://localhost:8969/assets/main.js` (or wherever sidecar is running from).

### DIFF
--- a/packages/overlay/src/index.tsx
+++ b/packages/overlay/src/index.tsx
@@ -23,10 +23,10 @@ export {
   DEFAULT_ANCHOR,
   DEFAULT_EXPERIMENTS,
   DEFAULT_SIDECAR_URL,
-  React,
-  ReactDOM,
   off,
   on,
+  React,
+  ReactDOM,
   trigger,
 };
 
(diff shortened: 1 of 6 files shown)


packages/overlay/src/index.tsx
```diff
@@ -107,19 +107,22 @@ export async function init(options: SpotlightOverlayOptions = {}) {
...
-    debug = false,
     integrations,
     experiments = DEFAULT_EXPERIMENTS,
-    fullPage = false,
     showClearEventsButton = true,
     initialEvents = undefined,
     startFrom = undefined,
   } = options;
 
+  const isLoadedFromSidecar = new URL(sidecarUrl).origin === document.location.origin;
```
Shubhdeep12, line 119: [NOT RELATED TO THIS PR]

Looking at these, I think we can give the support of query params to users for changing config of spotlight when running overlay via sidecar url. 
Something like 'localhost:8969?integrations=sentry,console&fullScreen=true'

Shubhdeep12 (approved):

packages/overlay/src/index.tsx
BYK: Aha, pretty good idea! I wonder if we can do something similar with the Electron app too. For that we may have to fall back to config files :(

PR #555, opened 2024-11-06
fix(overlay): UI fixes.

Shubhdeep12:
Before opening this PR:

- Fixed condition for showing open file in editor icon on error frame - missed here https://github.com/getsentry/spotlight/pull/550
- changed route and added navigation in performance tab to make queries tab default active.
- Fixed showing of 0 in false condition in span details

### DIFF
--- a/packages/overlay/src/integrations/sentry/components/events/error/Frame.tsx
+++ b/packages/overlay/src/integrations/sentry/components/events/error/Frame.tsx
@@ -51,10 +51,9 @@ function FileActions({ frame }: { frame: EventFrame }) {
     return null;
   }
   const resolvedFilename = resolveFilename(frame.filename);
-  const canOpenFileInEditor = !frame.filename?.includes(':');
   return (
     <div className="flex items-center gap-2">
-      {canOpenFileInEditor && <OpenInEditor file={`${resolvedFilename}:${frame.lineno}:${frame.colno}`} />}
+      <OpenInEditor file={`${resolvedFilename}:${frame.lineno}:${frame.colno}`} />
       <CopyToClipboard data={resolvedFilename} />
     </div>
   );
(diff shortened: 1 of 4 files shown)


packages/overlay/src/integrations/sentry/components/events/error/Frame.tsx
```diff
@@ -51,9 +51,10 @@ function FileActions({ frame }: { frame: EventFrame }) {
     return null;
   }
   const resolvedFilename = resolveFilename(frame.filename);
+  const canOpenFileInEditor = !frame.filename?.includes(':');
```
BYK, line 54: Ah, this removal was intentional. We should not need this anymore. Did you find an edge case?

packages/overlay/src/integrations/sentry/components/performance/PerformanceTabDetails.tsx
```diff
@@ -58,7 +58,8 @@ export default function PerformanceTabDetails() {
           <Route path="webvitals" element={<WebVitals />} />
           <Route path="webvitals/:page" element={<WebVitalsDetail />} />
           {/* Default tab */}
-          <Route path="*" element={<Queries showAll={showAll} />} />
+          <Route path="queries" element={<Queries showAll={showAll} />} />
+          <Route path="*" element={<Navigate to="/performance/queries" replace />} />
```
BYK, line 62: Kind of weird that we need to do this.

But thanks for the fix, it's been annoying me 😅

BYK (changes requested):
Asking changes for the edit button check. Rest looks great!

packages/overlay/src/integrations/sentry/components/events/error/Frame.tsx
Shubhdeep12: Not edge case, but i thought we are giving the ability to open specific files to get to the error.
Not sure if opening other build files makes much sense.
but we can do this.
reverting it

BYK (approved):

BYK:
Need to update the changelog entry now 😅

Shubhdeep12:
> Need to update the changelog entry now 😅

yeah, my bad.
its updated now.

Issue #524, opened 2024-09-12
Not all URLs are controllable by `sidecarUrl`


freeatnet:
### Environment

- `@spotlightjs/spotlight`: `2.4.1` 
- A Next.js 14.2.7 app running over an ngrok tunnel

### Steps to Reproduce

1. Set up a Next.js app and reverse-proxy it through an ngrok tunnel for mobile development.
2. Set up Spotlight [as documented](https://spotlightjs.com/setup/nextjs/).
3. Set up a Sidecar URL as follows:
    1. Update the `Spotlight.init({ sidecarUrl: "https://my-tunnel.ngrok.io/_sidecar/stream" })`, using a prefix in the path, because `/stream` is needed for an app route.
    2. In `next.config.js`, set up [a rewrite:](https://nextjs.org/docs/app/api-reference/next-config-js/rewrites#rewriting-to-an-external-url) `{ source: "/_spotlight/stream", destination: "http://localhost:8969/stream" }`
5. Launch the app and use the overlay within the page.

### Expected Result

All calls needed for the overlay are done through `sidecarUrl` (like with `tunnelRoute` [in Sentry SDK](https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#configure-tunneling-to-avoid-ad-blockers)).

### Actual Result

The overlay causes 404s calling `https://my-tunnel.ngrok.io/clear` (breaking the "clear events" button sync) and `https://my-tunnel.ngrok.io/contextlines`.

### Triage

At least two locations in the code assume that the origin is the only custom part of the `sidecarUrl`:
1. [`packages/overlay/src/integrations/sentry/data/sentryDataCache.ts:49`](https://github.com/getsentry/spotlight/blob/42022e524fcabb2ea44220daf91d8e0365645373/packages/overlay/src/integrations/sentry/data/sentryDataCache.ts#L49C10-L49C30)
2. [`packages/overlay/src/App.tsx:140`](https://github.com/getsentry/spotlight/blob/42022e524fcabb2ea44220daf91d8e0365645373/packages/overlay/src/App.tsx#L139-L140)

My current workaround is to add `/clear` and `/contextlines` to the Next.js rewrites above.  It'd be helpful if `sidecarUrl` acted as the base URL for all sidecar-related endpoints; however, this might require a breaking change in the config semantics.

BYK:
[...]

PR #558, opened 2024-11-10
Fix: Sidecar URL issues fixed across spotlight

Shubhdeep12:
Before opening this PR:

related #524

### DIFF
--- a/packages/overlay/src/App.tsx
+++ b/packages/overlay/src/App.tsx
@@ -136,10 +136,10 @@ export default function App({
   // See https://github.com/remix-run/react-router/issues/7634
   const navigate = useNavigate();
   const clearEvents = useCallback(async () => {
-    const { origin } = new URL(sidecarUrl);
-    const clearEventsUrl: string = `${origin}/clear`;
-
     try {
+      const { origin } = new URL(sidecarUrl);
+      const clearEventsUrl: string = `${origin}/clear`;
+
       await db.reset();
       await fetch(clearEventsUrl, {
         method: 'DELETE',
--- a/packages/overlay/src/components/OpenInEditor.tsx
+++ b/packages/overlay/src/components/OpenInEditor.tsx
@@ -1,12 +1,12 @@
 import type React from 'react';
 import { useCallback } from 'react';
+import sentryDataCache from '~/integrations/sentry/data/sentryDataCache';
 import { ReactComponent as PenIcon } from '../assets/pen.svg';
 
 export default function OpenInEditor({ file }: { file: string }) {
   const openInEditor = useCallback(
     (evt: React.MouseEvent) => {
-      // TODO: Make this URL dynamic based on sidecarUrl!
-      fetch('http://localhost:8969/open', {
+      fetch(`${sentryDataCache.getSidecarUrl()}/open`, {
         method: 'POST',
         body: file,
         credentials: 'omit',
--- a/packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
+++ b/packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
@@ -34,7 +34,8 @@ class SentryDataCache {
 
   protected subscribers: Map<string, Subscription> = new Map();
 
-  protected contextLinesProvider: string = new URL(DEFAULT_SIDECAR_URL).origin + CONTEXT_LINES_ENDPOINT;
+  protected sidecarUrl: string = new URL(DEFAULT_SIDECAR_URL).origin;
+  protected contextLinesProvider: string = this.sidecarUrl + CONTEXT_LINES_ENDPOINT;
 
   constructor(
     initial: (SentryEvent & {
@@ -46,7 +47,12 @@ class SentryDataCache {
 
   setSidecarUrl(url: string) {
     const { origin } = new URL(url);
+    this.sidecarUrl = origin;
     this.contextLinesProvider = origin + CONTEXT_LINES_ENDPOINT;
+  }
+
+  getSidecarUrl() {
+    return this.sidecarUrl;
   }
 
   inferSdkFromEvent(event: SentryEvent) {
--- a/packages/overlay/src/sidecar.ts
+++ b/packages/overlay/src/sidecar.ts
@@ -8,7 +8,9 @@ export function connectToSidecar(
   setOnline: React.Dispatch<React.SetStateAction<boolean>>,
 ): () => void {
   log('Connecting to sidecar at', sidecarUrl);
-  const source = new EventSource(sidecarUrl);
+  const { origin } = new URL(sidecarUrl);
+  const sidecarStreamUrl: string = `${origin}/stream`;
+  const source = new EventSource(sidecarStreamUrl);
 
   for (const [contentType, listener] of Object.entries(contentTypeListeners)) {
     source.addEventListener(contentType, listener);
--- a/packages/website/src/content/docs/reference/configuration.md
+++ b/packages/website/src/content/docs/reference/configuration.md
@@ -35,15 +35,15 @@ init({
 
 #### `sidecarUrl`
 
-**type:** `string` **default:** `"http://localhost:8969/stream"`
+**type:** `string` **default:** `"http://localhost:8969"`
 
-The Sidecar event-source stream endpoint URL.
+The Sidecar event-source endpoint URL.
 
 Set this option if you have the sidecar running on another URL than the default one.
 
 ```ts
 init({
-  sidecarUrl: 'http://localhost:8969/stream',
+  sidecarUrl: 'http://localhost:8969',
 });
 ```
 
--- a/packages/website/src/content/docs/sidecar/index.mdx
+++ b/packages/website/src/content/docs/sidecar/index.mdx
@@ -21,7 +21,7 @@ Also in the Sentry SDKs the URL can be configured via the `spotlight` option, fo
 import sentry_sdk
 
 sentry_sdk.init(
-    spotlight="http://localhost:8969/stream",
+    spotlight="http://localhost:8969",
 )
 ```
 
--- a/.changeset/twenty-fireants-fix.md
+++ b/.changeset/twenty-fireants-fix.md
@@ -1,0 +1,8 @@
+---
+'@spotlightjs/spotlight': minor
+'@spotlightjs/overlay': minor
+'@spotlightjs/sidecar': minor
+---
+
+- Sidecar url made generic to support all sidecar server routes.
+- No use of static sidecar url.
--- a/packages/overlay/src/constants.ts
+++ b/packages/overlay/src/constants.ts
@@ -1,4 +1,4 @@
-export const DEFAULT_SIDECAR_URL = 'http://localhost:8969/stream';
+export const DEFAULT_SIDECAR_URL = 'http://localhost:8969';
 
 export const DEFAULT_EXPERIMENTS = {
   'sentry:focus-local-events': true,
--- a/packages/overlay/src/types.ts
+++ b/packages/overlay/src/types.ts
@@ -24,7 +24,7 @@ export type SpotlightOverlayOptions = {
    * will use this URL instead of the default URL to connect to the sidecar
    * and to listen to incoming events.
    *
-   * @default "http://localhost:8969/stream"
+   * @default "http://localhost:8969"
    */
   sidecarUrl?: string;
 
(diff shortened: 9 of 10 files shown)


Shubhdeep12:
Didn't removed getting origin out of sidecar url because it can affect the users already passing sidecar stream url

packages/overlay/src/App.tsx
```diff
@@ -136,10 +136,10 @@ export default function App({
   // See https://github.com/remix-run/react-router/issues/7634
   const navigate = useNavigate();
   const clearEvents = useCallback(async () => {
-    const { origin } = new URL(sidecarUrl);
-    const clearEventsUrl: string = `${origin}/clear`;
-
     try {
+      const { origin } = new URL(sidecarUrl);
+      const clearEventsUrl: string = `${origin}/clear`;
```
BYK, line 141: Why move these into the `try-catch` block? Due to potentially malformed URL issues?

packages/overlay/src/components/OpenInEditor.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { useCallback } from 'react';
+import sentryDataCache from '~/integrations/sentry/data/sentryDataCache';
 import { ReactComponent as PenIcon } from '../assets/pen.svg';
 
 export default function OpenInEditor({ file }: { file: string }) {
   const openInEditor = useCallback(
     (evt: React.MouseEvent) => {
-      // TODO: Make this URL dynamic based on sidecarUrl!
-      fetch('http://localhost:8969/open', {
+      fetch(`${sentryDataCache.getSidecarUrl()}/open`, {
```
BYK, line 9: Should we maybe use the `new URL` constructor here or use something like `urljoin` just in case?

packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
```diff
@@ -34,7 +34,8 @@ class SentryDataCache {
 
   protected subscribers: Map<string, Subscription> = new Map();
 
-  protected contextLinesProvider: string = new URL(DEFAULT_SIDECAR_URL).origin + CONTEXT_LINES_ENDPOINT;
+  protected sidecarUrl: string = new URL(DEFAULT_SIDECAR_URL).origin;
+  protected contextLinesProvider: string = this.sidecarUrl + CONTEXT_LINES_ENDPOINT;
```
BYK, line 38: Speaking of hypocrisy... 😅  -- Same comment above, we should probably use some sort of a URL builder at this point.

packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
```diff
@@ -34,7 +34,8 @@ class SentryDataCache {
 
   protected subscribers: Map<string, Subscription> = new Map();
 
-  protected contextLinesProvider: string = new URL(DEFAULT_SIDECAR_URL).origin + CONTEXT_LINES_ENDPOINT;
+  protected sidecarUrl: string = new URL(DEFAULT_SIDECAR_URL).origin;
```
BYK, line 37: Do we need the `new URL().origin` trick with you changing the `DEFAULT_SIDECAR_URL` thingy in the other file?

packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
```diff
@@ -46,9 +47,14 @@ class SentryDataCache {
 
   setSidecarUrl(url: string) {
     const { origin } = new URL(url);
+    this.sidecarUrl = origin;
```
BYK, line 50: If we go back to the `url.origin` we are not solving the problem reported in #524 as they want to be able to use the sidecar under a main root URL. That means we should let go of using `.origin` and slice the path up and just strip the last `/stream` to get the "root" sidecar URL. Makes sense?

packages/overlay/src/sidecar.ts
```diff
@@ -8,7 +8,9 @@ export function connectToSidecar(
   setOnline: React.Dispatch<React.SetStateAction<boolean>>,
 ): () => void {
   log('Connecting to sidecar at', sidecarUrl);
-  const source = new EventSource(sidecarUrl);
+  const { origin } = new URL(sidecarUrl);
+  const sidecarStreamUrl: string = `${origin}/stream`;
+  const source = new EventSource(sidecarStreamUrl);
```
BYK, line 13: Feels like this should be the ground truth for the sidecarURL and we should not have anything stored in `sentryDataCache` regarding the URL?

packages/website/src/content/docs/sidecar/index.mdx
```diff
@@ -21,7 +21,7 @@ Also in the Sentry SDKs the URL can be configured via the `spotlight` option, fo
 import sentry_sdk
 
 sentry_sdk.init(
-    spotlight="http://localhost:8969/stream",
+    spotlight="http://localhost:8969",
```
BYK, line 24: No no no definitely not. The SDKs expect the `/stream` part and at this point we should keep it that way unfortunately.

BYK (changes requested):

packages/website/src/content/docs/sidecar/index.mdx
Shubhdeep12: yeah, but ideally we should just ask for sidecar url.
I know initially, we were only having stream url i.e. why i guess we asked for stream url directly.

let me think if we could do something to support /stream url as well here.

packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
Shubhdeep12: i am removing the logic to extract origin.

packages/overlay/src/integrations/sentry/data/sentryDataCache.ts
Shubhdeep12: yeah lets use `new URL()`

packages/overlay/src/components/OpenInEditor.tsx
Shubhdeep12: yeah lets use `new URL()`

packages/overlay/src/App.tsx
Shubhdeep12: yes correct

packages/website/src/content/docs/sidecar/index.mdx
BYK: @Shubhdeep12 I think a cheap hack is to see if we have the URL ending with `/stream` and try to guess from there.

That said not having the `/stream` part in the SDKs is a no go, as it requires us to change all existing SDKs to support this _and_ have a fallback or cut a major release. Don't want either so that ship seems to have sailed.

packages/overlay/src/sidecar.ts
Shubhdeep12: Hey @BYK, could you please explain what you meant here?

The sidecarURL is either provided by the user or defaults to a pre-set URL, and then it's passed here in this file. Ideally, to use the sidecarURL effectively, we should store it somewhere.

Are you suggesting that instead of storing it in SentryDataCache, we should store it somewhere else, like in a context, so that other integrations can also access the sidecarURL?

packages/website/src/content/docs/sidecar/index.mdx
Shubhdeep12: No, I wasn't intending to remove the /stream path. I meant that the user should focus on just sending the sidecarURL while initializing Spotlight. We'll handle which path to use for different cases, like:

- /stream for EventSource,
- /clear for clearing all events,
- /open for opening the editor, etc.

Shubhdeep12: Can we do this-
- Remove the /stream suffix from sidecar URL passed in Spotlight.init
- Store the sidecar url
- and then using /stream, /open, /clear ,etc based on the need in the code

Shubhdeep12: Just pushed this, let me know your thoughts on this.

packages/overlay/src/App.tsx
```diff
@@ -136,10 +136,9 @@ export default function App({
   // See https://github.com/remix-run/react-router/issues/7634
   const navigate = useNavigate();
   const clearEvents = useCallback(async () => {
-    const { origin } = new URL(sidecarUrl);
-    const clearEventsUrl: string = `${origin}/clear`;
-
     try {
+      const clearEventsUrl: string = new URL('/clear', sidecarUrl).href;
```
BYK, line 140: Very nice!

packages/overlay/src/components/OpenInEditor.tsx
```diff
@@ -1,19 +1,21 @@
 import type React from 'react';
 import { useCallback } from 'react';
+import { useSpotlightContext } from '~/lib/useSpotlightContext';
 import { ReactComponent as PenIcon } from '../assets/pen.svg';
 
 export default function OpenInEditor({ file }: { file: string }) {
+  const { sidecarUrl } = useSpotlightContext();
```
BYK, line 7: 🚀

packages/overlay/src/components/OpenInEditor.tsx
```diff
@@ -1,19 +1,21 @@
...
+  const sidecarOpenUrl: string = new URL('/open', sidecarUrl).href;
```
BYK, line 8: Since you abstracted out `sidecarUrl` and this `new URL(x, sidecarUrl).href` looks like a pattern, do you think we can change the variable from the context to a function:

```ts
const { getSidecarUrl } = useSpotlightContext();
const openUrl = getSidecarUrl('open');
```

packages/overlay/src/index.tsx
```diff
@@ -22,7 +23,7 @@ export {
   CONTEXT_LINES_ENDPOINT,
   DEFAULT_ANCHOR,
   DEFAULT_EXPERIMENTS,
-  DEFAULT_SIDECAR_URL,
+  DEFAULT_SIDECAR_STREAM_URL as DEFAULT_SIDECAR_URL,
```
BYK, line 26: mistake?

packages/overlay/src/index.tsx
```diff
@@ -130,12 +132,13 @@ export async function init(options: SpotlightOverlayOptions = {}) {
   const finalExperiments = { ...DEFAULT_EXPERIMENTS, ...experiments };
 
   // Sentry is enabled by default
-  const defaultIntegrations = () => [sentry({ sidecarUrl })];
+  const defaultIntegrations = () => [sentry({ sidecarUrl: sidecarBaseUrl })];
```
BYK, line 135: Since we have that context thing, do we still need to pass this as an init option?

packages/website/src/content/docs/reference/configuration.md
```diff
@@ -43,7 +43,7 @@ Set this option if you have the sidecar running on another URL than the default
 
 ```ts
 init({
-  sidecarUrl: 'http://localhost:8969/stream',
+  sidecarUrl: 'http://localhost:8969/sidecar',
```
BYK, line 46: Did you mean
```suggestion
  sidecarUrl: 'http://localhost:8969/',
```

BYK (changes requested):
I love where this patch is going, amazing!

Just added some notes and I think I spotted two errors. Once we resolve the errors, I'm happy to merge even without you addressing my other suggestions.

packages/overlay/src/index.tsx
Shubhdeep12: Yes, because this integration is exported, user can pass custom sidecar URL here as well.

BYK: I understand but that doesn't sound something sensible to me.

packages/overlay/src/index.tsx
Shubhdeep12: no, here we are exporting stream url for users. `DEFAULT_SIDECAR_URL` is an alias here not constant

packages/website/src/content/docs/reference/configuration.md
Shubhdeep12: it is a mistake, this should not be changed.

packages/overlay/src/index.tsx
Shubhdeep12: yeah, now we have sidecarurl in Spotlight context.

Removed sidecarurl from props -> https://github.com/getsentry/spotlight/pull/558/commits/9abb99156926d2e2e4ba920bddecd38297597f0c

packages/overlay/src/index.tsx
BYK: Ah, this is to keep the interface the same. Great!

BYK (approved):

PR #559, opened 2024-11-12
feat(spotlight): Add self-contained spotlight binary creation

BYK:


Shubhdeep12:
working fine in linux 🚀

packages/sidecar/package.json
```diff
@@ -52,6 +52,6 @@
     "extends": "../../package.json"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=20"
```
Lms24, line 55: are we dropping support for node 18? This would be a breaking change because Node 18 isn't EOL yet. afaik setting `engines` will only produce a warning in conventional `npm install` runs but if we actually use node 20+ only features, we might still run into incompatibilities.

.changeset/violet-trees-help.md
```diff
@@ -0,0 +1,6 @@
+---
```
Lms24, line 1: if we drop support for Node 18, we need to make this a major release.

Also I noticed we made changes in more packages than listed in the changeset.

Lms24 (commented):
Would you mind adding a small PR description? Looks good to me mostly on first glance but I don't have context around package binary signing and related stuff. Had a q about Node support though.

BYK:
@Lms24 thanks for the review! Updated the PR desc and responded to your question.

packages/sidecar/package.json
Lms24: I'm happy to let you make the call on that! Just wanted to mention that users who are currently using spotlight in a node 18 project (for example, by calling `Spotlight.init()`) might get warnings about spotlight being incompatible if they update their spotlight packages, due to `engines` sepcifying Node 20+. Depending on their npm/package manager settings, this could also throw an error. In SDK-land, we treat this as a breaking change.

BYK: Oh wait, I think I can just revert this but keep the main repo Node version at 22 for many things. I'll see if I can set up a test matrix.

BYK: All fair calls btw, I'm just treating Spotlight a bit looser than SDKs as it is quite new, not widely adopted, and the stakes are much lower compared to SDKs.

.changeset/violet-trees-help.md
BYK: Removed other changes for now so we should be good.

packages/sidecar/package.json
Lms24: That's also fair, no worries!

Lms24 (approved):

PR #563, opened 2024-11-19
feat(span): Use JsonViewer for JSON-like queries such as MongoDB

### LINKED ISSUE #562: MongoDB queries are not formatted
MongoDB queries are valid JSON so we should use our nice JSON viewer when we detect a query is valid JSON.

BYK:
Fixes #562.

### DIFF
--- a/packages/overlay/src/integrations/sentry/components/traces/spans/SpanDetails.tsx
+++ b/packages/overlay/src/integrations/sentry/components/traces/spans/SpanDetails.tsx
@@ -22,15 +22,16 @@ function DBSpanDescription({ desc, dbType }: { desc: string; dbType?: string }) 
     }
   }
 
+  let description = desc;
   if (desc.match(/^(SELECT|INSERT|UPDATE|DELETE|TRUNCATE|ALTER) /i)) {
     try {
-      desc = formatSQL(desc.replace(/([\s,(])(%[a-z])([\s,)])/gim, '$1?$3'), { language: dbType || 'sql' });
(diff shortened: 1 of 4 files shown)


Shubhdeep12:
Hi @BYK can you please add an envelope under overlay/_fixtures with mongodb queries. that would be helpful.
thanks

packages/overlay/src/integrations/sentry/components/traces/spans/SpanDetails.tsx
```diff
@@ -9,16 +9,28 @@ import { getDuration } from '../../../utils/duration';
...
+    } catch (_err) {
+      // pass
+    }
+  }
 
-function formatSpanDescription(desc: string) {
   if (desc.match(/^(SELECT|INSERT|UPDATE|DELETE|TRUNCATE|ALTER) /i)) {
     try {
-      return formatSQL(desc.replace(/([\s,(])(%[a-z])([\s,)])/gim, '$1?$3'));
+      desc = formatSQL(desc.replace(/([\s,(])(%[a-z])([\s,)])/gim, '$1?$3'), { language: dbType || 'sql' });
```
Shubhdeep12, line 27: although its fine, but i think we should not mutate the arguments(desc).

Shubhdeep12:
> Hi @BYK can you please add an envelope under overlay/_fixtures with mongodb queries. that would be helpful. thanks

we can do this later. accepting this now.

Shubhdeep12 (approved):

Shubhdeep12 (approved):

Issue #477, opened 2024-08-06
Surface Sentry SDK client status reports in Spotlight


BYK:
It would be great to surface SDK client reports such as events getting dropped due to size etc.

Shubhdeep12:
how can we get this information?

BYK:
@Shubhdeep12 It's just a different type of event: https://develop.sentry.dev/sdk/telemetry/client-reports/

Shubhdeep12:
Okay. 
I never read about this one. 

Also, is there a way I can trigger this event manually? 
Looks like there is no built-in function to trigger this.

BYK:
Yeah, I agree it's tricky to trigger. I think easist would be to install a `beforeSend` hook and keep dropping events there to trigger a report. I saw that's one of the categories.

PR #592, opened 2024-11-28
feat: Restructured sentry integration tabs

### LINKED ISSUE #82: SDKs should become supplemental information
This is a carry over from hackweek when it was fully Sentry-coupled. Its still valuable information, but maybe we can combine SDKs with some kind of Meta tab that also includes integration information and versions?

e.g.

```
[Integrations]

Sentry - 1.0
├── @sentry/node - 3s ago - blah blah
└── @sentry/browser - blah blah
Astro - 1.0
```

Current display:

image

### LINKED ISSUE #581: To have Tabs in spotlight's sentry integration almost similar to Sentry dashboard
Image

Ideally Spotlight's sentry integration should have tabs somewhat close to sentry dashboard, so proposing this structure:

- Errors/Issues
- Explore
  - Traces
  - Profiles  #565 
- Insights
  - Queries
  - Web Vitals
  - Resources
- Performance - this will list all transactions and links back to trace view, profiles, etc.
- SDKs
- Developer Info

Shubhdeep12:
Fixes #581, fixes #82.

 - Restructured tabs
 - added performance tabs to list all transactions.

New Performance tab - includes transactions info.
image
image


Shubhdeep12:
Should I add a Profiles tab under Explore as well?
For now, we can show a list of profiles and linked transactions with the profiles.

As for the flamegraph, we can try a PoC later in another PR using some libraries to render the data.

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
```diff
@@ -0,0 +1,159 @@
...
+      [TRANSACTION_SUMMARY_SORT_KEYS.transaction]: (a, b) => {
+        if (a.event_id! < b.event_id!) return -1;
+        if (a.event_id! > b.event_id!) return 1;
+        return 0;
+      },
+      [TRANSACTION_SUMMARY_SORT_KEYS.traces]: (a, b) => {
+        if (a?.contexts?.trace?.trace_id < b?.contexts?.trace?.trace_id) return -1;
+        if (a?.contexts?.trace?.trace_id > b?.contexts?.trace?.trace_id) return 1;
+        return 0;
+      },
```
BYK, line 53: I feel like we can get rid of the non-null assertions and the optional chains if we just typed these comparator functions properly. What do you think? Non-null is not a big deal but the optional chaining definitely takes a toll on performance.

Finally (this is a side node), do you think sorting by random trace ids is actually useful? I'd much rather we have sorting via tracing timestamps which is something we keep getting asked.

BYK (changes requested):
Okay I'm not gonna go too deep into code bits as this should be mostly a rearrangement of the UI so I trust you there.

I really like the direction but there are a few things I was thrown off and I'd love to get them fixed before moving forward:

1. There's only one tab under "Explore" which is just transactions and it looks like a slightly different (better) listing of transactions under Performance. I feel like we should only have "Explore" here at this point?
2. I think we should move SDKs and Developer info under insights or "Explore" as that is complementary information. This is already brought up in #82 so we can aim to resolve that too here. "Explore" tab also seems like a good place to have a subtab for #477 once we implement it.
3. I keep seeing the "x other transactions are hidden from other sessions" in almost _all_ tabs: Explore, Performance, Insights... And in some of them, it doesn't seem to be doing anything. I feel like it needs to be a higher-level filter and should only be shown once and easily toggled (once you click reveal, it's over -- you cannot go back). This is probably a job for a follow-up PR.
4. As I noted with inline comments, Transaction sorting should be done by timestamp, not random IDs as that's definitely not useful.

Shubhdeep12:
> There's only one tab under "Explore" which is just transactions and it looks like a slightly different (better) listing of transactions under Performance. I feel like we should only have "Explore" here at this point?

ok so this is what implemented right now:
- Explore tab:  Traces tab, SDKs, Developer Info.
- Performance tab: List of all transactions recorded yet.
  - Since there can be multiple transaction events in a single trace. So this transactions list will help user to debug and view what type of transactions are recorded yet. These transactions are grouped by name of transaction.
  - Under every transaction summary- we will display Transaction event id , start timestamp (added after your comment), total duration (added after your comment), and a trace id in which that particular transaction was recorded.
 
Let me know if you still think that its fine to just have explore tab.
All this references are taken from Sentry dashboard.


> I think we should move SDKs and Developer info under insights or "Explore" as that is complementary information. This is already brought up in https://github.com/getsentry/spotlight/issues/82 so we can aim to resolve that too here. "Explore" tab also seems like a good place to have a subtab for https://github.com/getsentry/spotlight/issues/477 once we implement it.

Done, added SDKs tab and Developer Info tab under Explore tab for now.

> I keep seeing the "x other transactions are hidden from other sessions" in almost all tabs: Explore, Performance, Insights... And in some of them, it doesn't seem to be doing anything. I feel like it needs to be a higher-level filter and should only be shown once and easily toggled (once you click reveal, it's over -- you cannot go back). This is probably a job for a follow-up PR.

yeah, lets do this in another PR, by having a single button to toggle this with a shared state with all sentry integration's tab.

> As I noted with inline comments, Transaction sorting should be done by timestamp, not random IDs as that's definitely not useful

Done.
- Removed unwanted sorting of IDs
- added timestamp and total duration wherever needed with sorting.




Some screenshots for reference - 
image
image

Shubhdeep12:
> > Should I add a Profiles tab under Explore as well?
> > For now, we can show a list of profiles and linked transactions with the profiles.
> 
> If you think that adds at least _some_ value (like if we can surface some high-level data from profile information) let's do it. Otherwise we can wait till we have proper/full profiles support.
> 
> > As for the flamegraph, we can try a PoC later in another PR using some libraries to render the data.
> 
> I want to do 2 things with profiles:
> 
> 1. Fill in missing instrumentation data in transactions or at least provide more details under some spans. I think this is easier (at least it is what I'm told) and would immediately bring value.
> 2. Actually have the flamegraph etc for deeper debugging. I was referred to this component which we can easily port: https://github.com/getsentry/sentry/blob/master/static/app/components/profiling/flamegraph/flamegraphPreview.tsx

Cool, lets add this in Profiles PR seperately.

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
Shubhdeep12: Removed these. Because as you mentioned, its not of much value to sort IDs

Shubhdeep12:
Or maybe, we can add Traces, Transactions, Profiles, Client Reports, etc all under Explore tab. wdyt @BYK ?

BYK:
>Or maybe, we can add Traces, Transactions, Profiles, Client Reports, etc all under Explore tab. wdyt @BYK ?

Kind of makes sense. The only thing is I think "transactions" is a subset of "traces" so feel like they should be joint maybe? Or we just have traces where we also put the errors under traces with filters through checkboxes for "only errors" or "only transactions"?

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
```diff
@@ -41,14 +43,16 @@ export default function TransactionDetail({ showAll }: { showAll: boolean }) {
       return [];
     }
     const COMPARATORS: Record<TransactionsSortTypes, TransactionsInfoComparator> = {
-      [TRANSACTION_SUMMARY_SORT_KEYS.transaction]: (a, b) => {
-        if (a.event_id! < b.event_id!) return -1;
-        if (a.event_id! > b.event_id!) return 1;
+      [TRANSACTION_SUMMARY_SORT_KEYS.timestamp]: (a, b) => {
+        if (a.start_timestamp < b.start_timestamp) return -1;
+        if (a.start_timestamp > b.start_timestamp) return 1;
```
BYK, line 48: At this point you should be able to do `(a, b) => a.start_timestamp - b.start_timestamp`

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
```diff
@@ -41,14 +43,16 @@ export default function TransactionDetail({ showAll }: { showAll: boolean }) {
...
         return 0;
       },
-      [TRANSACTION_SUMMARY_SORT_KEYS.traces]: (a, b) => {
-        if (a?.contexts?.trace?.trace_id < b?.contexts?.trace?.trace_id) return -1;
-        if (a?.contexts?.trace?.trace_id > b?.contexts?.trace?.trace_id) return 1;
+      [TRANSACTION_SUMMARY_SORT_KEYS.duration]: (a, b) => {
+        const aDuration = a.timestamp - a.start_timestamp;
+        const bDuration = b.timestamp - b.start_timestamp;
+        if (aDuration < bDuration) return -1;
+        if (aDuration > bDuration) return 1;
```
BYK, line 55: Similar to above:
```ts
(a, b) => (a.timestamp - a.start_timestamp) - (b.timestamp - b.start_timestamp)
```

Which then becomes:
```ts
(a, b) => a.timestamp + b.start_timestamp - a.start_timestamp - b.timestamp
```

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
```diff
@@ -16,7 +18,7 @@ export default function TransactionDetail({ showAll }: { showAll: boolean }) {
   const helpers = useSentryHelpers();
 
   const [sort, setSort] = useState({
-    active: TRANSACTION_SUMMARY_SORT_KEYS.transaction,
+    active: '',
```
BYK, line 21: We should probably sort by timestamp by default, wdyt?

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
```diff
@@ -131,19 +140,27 @@ export default function TransactionDetail({ showAll }: { showAll: boolean }) {
...
                 <tr key={txn.event_id} className="hover:bg-primary-900">
-                  <td className="text-primary-200 w-3/5 truncate whitespace-nowrap px-6 py-4 text-left text-sm font-medium">
+                  <td className="text-primary-200 w-2/5 truncate whitespace-nowrap px-6 py-4 text-left text-sm font-medium">
                     {/* Ref: https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa */}
                     <Link
                       className="truncate hover:underline"
                       to={`/explore/traces/${txn.contexts?.trace?.trace_id}/spans/${txn.contexts?.trace?.span_id}`}
                     >
-                      {txn.event_id}
+                      {txn.event_id.substring(0, 8)}
```
BYK, line 149: This `.substring(0, 8)` thing should become a utility function so we don't have to repeat the same logic and magic number (`8`) everywhere.

BYK (changes requested):
Looking good! I'll do another pass, did a light one this time. LMK if you need a deeper dive before your next update.

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
Shubhdeep12: correct.
 feeling noob 😆

Shubhdeep12:
> Kind of makes sense. The only thing is I think "transactions" is a subset of "traces" so feel like they should be joint maybe? 

I think we should keep the transactions tab separate because Traces is collection of Transactions. And i think Transactions should have separate detail page as well.
Later we can add contexts, and links to profiles etc inside every transaction as well. right now we just added links of traces.

I think this is same reason why transactions list is shown under Performance tab on Sentry dashboard.


> Or we just have traces where we also put the errors under traces with filters through checkboxes for "only errors" or "only transactions"?

I think we are adding all events and their spans already in Trace. And yes, similar to Errors Tab, we can have Transactions tab under Trace detail as well.

packages/overlay/src/integrations/sentry/components/explore/ExploreTabDetails.tsx
```diff
@@ -0,0 +1,52 @@
...
+    },
+    {
+      id: 'sdks',
+      title: 'SDKs',
+      active: activeTab === 'sdks',
+      onSelect: () => setActiveTab('sdks'),
+    },
+    {
+      id: 'devInfo',
+      title: 'Developer Info',
```
BYK, line 32: Shall we name this "Raw Events" or something now?

packages/overlay/src/integrations/sentry/components/insights/InsightsTabDetails.tsx
```diff
@@ -0,0 +1,68 @@
...
+      active: activeTab === 'webvitals',
+      onSelect: () => setActiveTab('webvitals'),
+    },
+    {
+      id: 'resources',
+      title: 'Resources',
+      active: activeTab === 'resources',
+      onSelect: () => setActiveTab('resources'),
+    },
+  ];
```
BYK, line 41: So much repetition in these tab definitions (same with the previous file). Maybe we can also have a `tab(id, title, extra)` helper to reduce repetition here and over there?

packages/overlay/src/integrations/sentry/components/performance/PerformanceTabDetails.tsx
```diff
@@ -1,68 +1,44 @@
...
-  const [activeTab, setActiveTab] = useState('');
   const [showAll, setShowAll] = useState(!context.experiments['sentry:focus-local-events']);
 
-  const hiddenItemCount = allSpans.length - localSpans.length;
+  const allTransactions = events.filter(e => e.type === 'transaction');
+  const filteredTransactions = showAll
+    ? allTransactions
+    : allTransactions.filter(
+        e => (e.contexts?.trace?.trace_id ? helpers.isLocalToSession(e.contexts?.trace?.trace_id) : null) !== false,
+      );
```
BYK, line 22:
```suggestion
    : allTransactions.filter(
        t => { const id = e.contexts?.trace?.trace_id; return !id || helpers.isLocalToSession(id);}
      );
```

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
```diff
@@ -0,0 +1,168 @@
...
+      prev.active === type
+        ? {
+            active: type,
+            asc: !prev.asc,
+          }
+        : {
+            active: type,
+            asc: false,
+          },
+    );
```
BYK, line 37: Looks like this entire thing can be a hook itself?

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  type TransactionsInfoComparator = (a: SentryTransactionEvent, b: SentryTransactionEvent) => number;
+  type TransactionsSortTypes = (typeof TRANSACTION_SUMMARY_SORT_KEYS)[keyof typeof TRANSACTION_SUMMARY_SORT_KEYS];
```
BYK, line 40: Move to `constants.ts`?

packages/overlay/src/integrations/sentry/components/performance/TransactionsList.tsx
```diff
@@ -0,0 +1,149 @@
...
+  type TransactionsInfoComparator = (
+    a: [string, GroupedTransactionsValue],
+    b: [string, GroupedTransactionsValue],
+  ) => number;
+  type TransactionsSortTypes = (typeof TRANSACTIONS_SORT_KEYS)[keyof typeof TRANSACTIONS_SORT_KEYS];
+
+  const transactionsList: [string, GroupedTransactionsValue][] = useMemo(() => {
+    const COMPARATORS: Record<TransactionsSortTypes, TransactionsInfoComparator> = {
+      [TRANSACTIONS_SORT_KEYS.count]: (a, b) => a[1].transactions.length - b[1].transactions.length,
+      [TRANSACTIONS_SORT_KEYS.count]: (a, b) => a[1].lastSeen - b[1].lastSeen,
```
BYK, line 50: Pretty sure keys are incorrect here. Guessing the second one should be `lastSeen` or something?
```suggestion
      [TRANSACTIONS_SORT_KEYS.count]: (a, b) => a[1].transactions.length - b[1].transactions.length,
      [TRANSACTIONS_SORT_KEYS.lastSeen]: (a, b) => a[1].lastSeen - b[1].lastSeen,
```

packages/overlay/src/integrations/sentry/components/performance/TransactionsList.tsx
```diff
@@ -0,0 +1,149 @@
...
+    };
+
+    const compareTransactions = COMPARATORS[sort.active] || COMPARATORS[TRANSACTIONS_SORT_KEYS.count];
```
BYK, line 53: Default should be `lastSeen` here too, right?
```suggestion
    const compareTransactions = COMPARATORS[sort.active] || COMPARATORS[TRANSACTIONS_SORT_KEYS.lastSeen];
```

packages/overlay/src/integrations/sentry/components/performance/TransactionsList.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const allTransactions = events.filter(e => e.type === 'transaction');
+    const filteredTransactions = showAll
+      ? allTransactions
+      : allTransactions.filter(
+          e => (e.contexts?.trace?.trace_id ? helpers.isLocalToSession(e.contexts?.trace?.trace_id) : null) !== false,
```
BYK, line 58: Same with the above suggestion. Getting `const id = e.contexts?.trace?.trace_id` first would make this much more readable.

packages/overlay/src/integrations/sentry/components/performance/TransactionsList.tsx
```diff
@@ -0,0 +1,149 @@
...
+        );
+
+    const sortedTransactions = filteredTransactions.sort((a, b) => a.start_timestamp - b.start_timestamp);
```
BYK, line 61: Should use an existing comparator?

BYK (changes requested):
Almost there! Identified a few blocker issues (sort types repeated). It is also quite challenging to review due to its size.

Once you make those fixes, I'll just check the latest changes and give it a manul click through before approving.

packages/overlay/src/integrations/sentry/components/explore/ExploreTabDetails.tsx
Shubhdeep12: yeah we can rename this to "Raw Events" OR "Events" OR "All Events"
Lets go with "Raw  Events" for now.

packages/overlay/src/integrations/sentry/utils/misc.ts
```diff
@@ -0,0 +1 @@
+export const truncateId = (eventId = '') => eventId.substring(0, 8);
```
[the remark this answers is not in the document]
Shubhdeep12: yeah

packages/overlay/src/integrations/sentry/components/performance/TransactionsList.tsx
Shubhdeep12: yup lastSeen would be better

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
Shubhdeep12: yup

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
Shubhdeep12: Lets keep it here for now. Will move all these in types file later for all pages.

BYK (approved):
This just looks amazing, thanks so much!

PR #596, opened 2024-11-29
fix: Traces should use the same context view as errors

### LINKED ISSUE #595: Errors and Traces are not using the same context view tab

BYK:
Fixes #595

Sample: 
image


Shubhdeep12 (commented):
Hey @BYK 

I'm a bit confused 😅.

We use EventContexts.tsx to display the context of Sentry events, which can be a transaction, error, or any other type of event.

When it comes to the context of a trace, it's essentially a collection of one or more transactions [(reference)](https://docs.sentry.io/concepts/key-terms/tracing/distributed-tracing/#traces). This is why we use TraceContext.tsx to represent the context of the trace as a whole.

Here in Spotlight, we name the trace after the root transaction, and if there are multiple transactions, we don't display the trace name.

Is there a specific reason for showing the root (or the first transaction) context as the trace context in this PR?

BYK:
What I'm after is using the same rich context logic between all events. Looks like we have a lot of information being sent by transaction events, including the top level `request` object (there's also a `response` object which we probably should add to the mix). I thought the best way to get that is through the `rootTransaction` or if none exists, the first transaction in the list. Is this an issue? If yes how can we fix that?

Regarding the earlier "General" tab, the information it shows is mostly useless as all of that is already available in multiple places. The most important one is the duration I'd say, which is not directly visible in the table but we show that in the top level transaction list _and_ on the main transaction details page.

.changeset/eleven-spies-thank.md
```diff
@@ -0,0 +1,6 @@
+---
+'@spotlightjs/overlay': patch
+'@spotlightjs/spotlight': patch
```
BYK, line 3: This is actually a big improvement and adds new functionality so marking it as a minor 
```suggestion
'@spotlightjs/overlay': minor
'@spotlightjs/spotlight': minor
```

Shubhdeep12:
> What I'm after is using the same rich context logic between all events. Looks like we have a lot of information being sent by transaction events, including the top level `request` object (there's also a `response` object which we probably should add to the mix). I thought the best way to get that is through the `rootTransaction` or if none exists, the first transaction in the list. Is this an issue? If yes how can we fix that?

No, my query was just that in the Trace context why we are showing root transaction context(when there can be more than 1 txns)?

Also, we show that context details on spandetails page when the root transaction's root span is clicked.

But,
For now we can display root transaction context details.
Because most of the important details about trace lies in root transaction only.


We can get users reviews and I'll also think what we can show in a trace context and how it's done in the sentry dashboard.

Shubhdeep12 (approved):

BYK:
@Shubhdeep12 
>But, for now we can display root transaction context details.
Because most of the important details about trace lies in root transaction only.
>
>We can get users reviews and I'll also think what we can show in a trace context and how it's done in the sentry dashboard.

Sounds good and thanks for unblocking this PR! Should we open a ticket to follow up on your comment? I'd need more details to understand what you have in mind (or, as usual, feel free to just come up with a PR :D)

PR #599, opened 2024-12-02
feat: Add method name to trace list

### LINKED ISSUE #594: More visibility of HTTP Method on traces
I mainly use the Traces tab in Spotlight, and find it really helpful to see what is going on in my application. One thing I find frustrating, though, is that it doesn't show the HTTP method anywhere; in my application, the same URL could be a GET or a POST, and it would be really nice to be able to see that at a glance.

I've hacked together a quick example of the sort of thing I'd find helpful!

For example:
Current:
Current
Current

Proposed example
Proposed Example
 Proposed Example

BYK:
Fixes #594.

With Astro where we _may_ get the method name in the description already:

image

With Django SDK where we don't get method name in transaction name:

image

### DIFF
--- a/packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
+++ b/packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
@@ -1,30 +1,27 @@
 import { useState } from 'react';
 import type { Trace } from '../../types';
 import { Link } from 'react-router-dom';
-import Badge from '~/ui/Badge';
 import CardList from '../../../../components/CardList';
 import TimeSince from '../../../../components/TimeSince';
 import classNames from '../../../../lib/classNames';
 import { useSpotlightContext } from '../../../../lib/useSpotlightContext';
 import { useSentryHelpers } from '../../data/useSentryHelpers';
 import { useSentryTraces } from '../../data/useSentryTraces';
+import Badge from '../../../../ui/Badge';
+import Tag from '../../../../ui/Tag';
 import { getDuration } from '../../utils/duration';
 import HiddenItemsButton from '../HiddenItemsButton';
 import TraceIcon from './TraceIcon';
 
 function TransactionName({ trace }: { trace: Trace }) {
   const method = String(
-    trace.rootTransaction?.contexts?.trace.data?.method || trace.rootTransaction?.request?.method || '(GET)',
+    trace.rootTransaction?.contexts?.trace.data?.method || trace.rootTransaction?.request?.method || '',
   );
-  const name = trace.rootTransactionName.startsWith(method)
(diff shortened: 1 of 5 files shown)


BYK:
@Shubhdeep12 I stole your fancy tag-pill style here but if you think plain is better, LMK. @adnweedon also appreciate your input on how these look as the original poster for the issue.

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -11,6 +12,21 @@ import { getDuration } from '../../utils/duration';
 import HiddenItemsButton from '../HiddenItemsButton';
 import TraceIcon from './TraceIcon';
 
+function TransactionName({ trace }: { trace: Trace }) {
+  const method = String(
+    trace.rootTransaction?.contexts?.trace.data?.method || trace.rootTransaction?.request?.method || '(GET)',
```
Shubhdeep12, line 17: @BYK i think we should only show method in trace list when there is `request.method` or `contexts.trace.data.method` present.
we should not set this as default value.

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -11,6 +12,21 @@ import { getDuration } from '../../utils/duration';
...
+  );
+  const name = trace.rootTransactionName.startsWith(method)
+    ? trace.rootTransactionName.slice(method.length + 1)
+    : trace.rootTransactionName;
+  return (
+    <div className="border-primary-300 bg-primary-900 divide-x-primary-300 text-smflex inline-flex divide-x whitespace-nowrap rounded-full border font-mono">
```
Shubhdeep12, line 23: the design of this is almost same as of Tag Component.
How about using it?

BYK: Yup, I can make it generalized.

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
Shubhdeep12: ok how about these?

Option 1
image


Option 2
image

Option 3
image

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
BYK: For sure! I think my favorite is option 2. And I'm also okay with option 3 of you prefer that. Didn't like option 1 much 😅

Shubhdeep12:
hey @BYK 
we should add method in Trace detail page as well in TraceHeader.

PR #600, opened 2024-12-02
fix: Contexts inside Trace detail is breaking in case of no child transactions.

Shubhdeep12:
Before opening this PR:


BYK (approved):
Thank you!

PR #611, opened 2024-12-04
fix(overlay): Use a maintained version of rjv

### LINKED ISSUE #609: Breaking CSP with `new Function()`
There is a CSP violation in our tests due to an invocation of `new Function("return this")` (this is a legacy alternative to `globalThis`). We've traced this to `@spotlightjs/overlay` having a dependency on `react-json-view`, which in turn has a dependency `istanbul-lib-instrument` which contains the problematic code.

`react-json-view` was [updated](https://github.com/mac-s-g/react-json-view/issues/316) in version 1.20.3 to conditionally import istanbul when `NODE_ENV=test` only, which I assume is why this CSP violation does not happen in development and production. `istanbul-lib-instrument` itself was [updated](https://github.com/istanbuljs/istanbuljs/pull/734) in version 6.0.1 to mitigate the problem, but `react-json-view` still uses version 4.0.3.

Note also that your version of react-json-view is no longer maintained, but there are replacements such as https://github.com/microlinkhq/react-json-view.

We've worked around this in our code base, but would be nice if Spotlight would update its dependencies so that this would go away.

BYK:
Fixes #609.

### DIFF
--- a/.changeset/cold-maps-type.md
+++ b/.changeset/cold-maps-type.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/overlay': patch
+---
+
+Use a maintained version of ReactJsonViewer
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,6 +14,7 @@
     "getsentry",
     "iife",
     "imagetools",
+    "microlink",
     "nextjs",
     "notarytool",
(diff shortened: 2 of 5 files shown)


Shubhdeep12 (approved):
🚀

PR #615, opened 2024-12-06
fix(overlay): Do not modify body styles on injection

### LINKED ISSUE #613: Remix/React hydration warning for style on body
Spotlight appears to be injecting styles onto the body client side. `style="overflow: auto;"`

### Environment

"@spotlightjs/spotlight": "2.7.0",
"@remix-run/node": "2.15.0"

### Steps to Reproduce

Follow instructions to initialise in development mode

### Expected Result

No warnings logged to console.

### Actual Result

"Warning: Extra attributes from the server: style"

BYK:
Fixes #613.

We used to add `overflow: auto` on injection, no matter what. This is bad for a few reasons:

1. Our default is already the closed state, so we should not do anything on initial injection
2. We don't store or know the original value of this style we are touching, which may cause breakage
3. It causes hydration issues as highlighted in #613

This PR injects a special class into the DOM using [adoptedStyleSheets API](https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets) on injection and just toggles this class on open/close.

### DIFF
--- a/packages/overlay/src/App.tsx
+++ b/packages/overlay/src/App.tsx
@@ -9,6 +9,7 @@ import { log } from './lib/logger';
 import useKeyPress from './lib/useKeyPress';
 import { connectToSidecar } from './sidecar';
 import type { NotificationCount, SpotlightOverlayOptions } from './types';
+import { SPOTLIGHT_OPEN_CLASS_NAME } from './constants';
 
 type AppProps = Omit<SpotlightOverlayOptions, 'debug' | 'injectImmediately'> &
   Required<Pick<SpotlightOverlayOptions, 'sidecarUrl'>>;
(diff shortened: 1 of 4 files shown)


packages/overlay/src/App.tsx
```diff
@@ -217,10 +218,10 @@ export default function App({
   useEffect(() => {
     if (!isOpen) {
       spotlightEventTarget.dispatchEvent(new CustomEvent('closed'));
-      document.body.style.overflow = 'auto';
```
Shubhdeep12, line 220: We just need to remove this, right? 

But I think it's good to have another stylesheet layer added with classname

Shubhdeep12 (approved):

packages/overlay/src/App.tsx
BYK: Removing wouldn't fix as we need a way of undoing our "hidden" override when a user toggles Spotlight closed

PR #620, opened 2024-12-10
feat(spotlight): Add Sentry SDK so we can monitor

### LINKED ISSUE #510: Add Sentry reporting to all possible packages
Right now we only have Sentry reporting on the `electron` package. Ideally, we should have this on all available packages. It should be straightforward to add this to the dedicated web UI in the `@spotlightjs/spotlight` package and the various `sidecar` packages, including the docker image, we should be careful with the overlay itself as it gets injected into other pages that already has a Sentry client.

BYK:
Fixes #510.

Also allows you to debug Spotlight, with Spotlight:

image


packages/overlay/src/utils/instrumentation.ts
```diff
@@ -0,0 +1,62 @@
...
+          sentryTraceParent[name] = description;
+        }
+      }
+    }
+  }
+  const hasTraceParent = Object.keys(sentryTraceParent).length === 2;
+  const integrations = [
+    // See docs for support of different versions of variation of react router
+    // https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/
+    Sentry.reactRouterV6BrowserTracingIntegration({
```
Lms24, line 22: h: how would we keep the traces created from this integration isolated from the trace on the page? Have you tested if they interfere?

packages/overlay/src/utils/instrumentation.ts
```diff
@@ -0,0 +1,62 @@
...
+      instrumentPageLoad: !hasTraceParent,
```
Lms24, line 23: can we always default to `false` here? Further below we're specifically calling `startBrowserTracingPageLoadSpan` anyway? Again, I'm very concerned about interference so this would only be relevant if we don't interfere at all.

packages/overlay/src/utils/instrumentation.ts
```diff
@@ -0,0 +1,62 @@
...
+    // plus for 100% of sessions with an error
+    // Learn more at
+    // https://docs.sentry.io/platforms/javascript/session-replay/configuration/#general-integration-configuration
+    replaysSessionSampleRate: 0.01,
+    replaysOnErrorSampleRate: 1.0,
+    ...options,
+  });
+
+  if (hasTraceParent && sentryClient) {
+    Sentry.startBrowserTracingPageLoadSpan(sentryClient, { name: initialTab }, sentryTraceParent);
```
Lms24, line 60: `startBrowserTracingPageLoadSpan` starts an incative idle span but sets it active on the current scope. Which again brings up the interference question. Which SDK sets the current scope?

packages/sidecar/src/main.ts
```diff
@@ -337,6 +362,8 @@ function startServer(
         server.listen(port);
         logger.info(`Port ${port} in use, retrying...`);
       }, 5000);
+    } else {
+      captureException(e);
```
Lms24, line 366: h: Catching errors is very tricky in CLI environments. They will include user paths in stack traces. In the wizard and in our bundler plugins, we basically only capture very generic error messages instead of the full error object to avoid leaking any user data into Sentry.

Lms24 (commented):
I'm a bit concerned with adding Sentry SDKs to the spotlight packages:

- In the overlay, we need to ensure that we don't interfere with the SDK installed in the host app. Starting pageload/navigation spans seems a bit dangerous. Specifically, this would be very problematic if the SDK version used in spotlight is equal to the SDK version used by the host app. 
- In the sidecar or any other CLI, we need to be careful about not catching user data. For instance stacktrace filenames. Also source context lines might be problematic. 
- I'd recommend taking a look at the wizard which is set up in a way to avoid leaking data as good as possible. 
- I'd also recommend [this develop guide](https://develop.sentry.dev/sdk/expected-features/setup-wizards/#telemetry) about telemetry collection in our CLI tools. Let's ensure we're transparent about it.

packages/overlay/src/utils/instrumentation.ts
BYK: This is not relevant as we only enable Sentry in the overlay when it is loaded from the Sidecar as a full-blown app in full page mode.

packages/overlay/src/utils/instrumentation.ts
BYK: I'm guessing this is safe when Spotlight is _the_ page? (as I took this from the docs 😅 )

packages/overlay/src/utils/instrumentation.ts
BYK: I'm also guessing the concern is when we are in overlay mode. In that case we don't even call this function so we should be safe.

The reason this is conditional is due to some browser compat edge case where my `server-timings` hack may not work so we may not be able to get a trace parent.

Lms24: yup, that was the concern. I think this can be disregarded now that I know we only enable sentry in standalone mode

packages/overlay/src/utils/instrumentation.ts
Lms24: yup, as discussed, this is fine

PR #631, opened 2024-12-13
fix: Binary to UTF-8 encoding shenanigans

### LINKED ISSUE #628: Carsh on certain replay envelopes
Sentry Issue: [SPOTLIGHT-ELECTRON-2D](https://sentry.sentry.io/issues/5738737682/?referrer=github_integration)

```
SyntaxError: Unexpected token '�', "����UDe���"... is not valid JSON
  at JSON.parse (<anonymous>)
  at init (../../../../../node_modules/.pnpm/@sentry+electron@5.3.0/node_modules/@sentry/electron/src/renderer/sdk.ts:93:1)
  at init (../../../../../node_modules/.pnpm/@sentry+electron@5.3.0/node_modules/@sentry/electron/src/renderer/sdk.ts:93:1)
  at init (../../../../../node_modules/.pnpm/@sentry+electron@5.3.0/node_modules/@sentry/electron/src/renderer/sdk.ts:93:1)
  at init (../../../../../node_modules/.pnpm/@sentry+electron@5.3.0/node_modules/@sentry/electron/src/renderer/sdk.ts:93:1)
...
(1 additional frame(s) were not displayed)
```

BYK:
We've been corrupting binary envelopes both on Sidecar and Overlay due to implicit utf-8 encodings all around. This patch fixes #628 for reals by keeping envelope data as binary until it reaches the SSE transport, and then uses a weird byte-by-byte conversion to Uint8Array on client side from the UTF-8 encoded version instead of using the `TextEncoder()` (as it also uses utf-8 instead of binary. The ideal solution to this would be to simply encode envelopes with `base64` but that seems like a breaking change, so postponing for now with proper comments in the code.


BYK:
@Lms24 if you think using `base64` encoding for the data between sidecar and the overlay wouldn't be a breaking change, I'd to it in a heartbeat. If they use "matching" versions, they would work but any new versions would not work with old versions of each other (old overlay won't work with new sidecar and vice versa)

Lms24 (approved):

PR #656, opened 2025-01-15
fix: Show something to account for empty strings

dcramer:


dcramer:
Ive still yet to diagnose _what_ events are failing to render, but its a better experience with some kind of placeholder.

image

BYK (approved):

PR #686, opened 2025-02-03
feat: Less indentation for deep trees with single child

BYK:
Before:
image

After:
image

Before: 

image

After:

image

### DIFF
--- a/.changeset/plenty-foxes-destroy.md
+++ b/.changeset/plenty-foxes-destroy.md
@@ -1,0 +1,8 @@
+---
+'@spotlightjs/spotlight': minor
+'@spotlightjs/electron': minor
+'@spotlightjs/overlay': minor
+'@spotlightjs/astro': minor
+---
+
+Flatter tree view for deeply nested traces with 1 child at each level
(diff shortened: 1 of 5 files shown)


dcramer:
I dont LOVE the solution, but I'm also struggling to come up w/ a better idea. We still have this issue on production Sentry too and its kind of a disaster...

BYK:
TBH this is quite a hacky solution and I certainly don't like it much. However until we can do better design and span grouping, this solution is the best bang for the buck.

With the upcoming profiling work, it is a necessity as profiles can go deep fast with 1-child trees.

dcramer:
I'm good w/ this - @vuluongj20 mentioned we tried an approach with a dashed line that would simply collapse a lot of these single child spans. I htink something akin to that is the right direction, but the dashed implementation IMO is still not obvious enough.

dcramer (approved):

vuluongj20:
Here's a screenshot for reference
image
Agree that it's not super obvious what the dashed line means. We can consider other visual cues.

BYK:
Created #686 for a proper follow up.

PR #714, opened 2025-03-03
fix(overlay): Process profiles without thread_metadata correctly

rvanvelzen:
Before opening this PR:

Per https://github.com/getsentry/sentry/blob/55d233c4ac771a876b1dc17a0a0dd330bc7179c8/static/app/types/profiling.d.ts#L119 the `thread_metadata` property is optional, and locally Spotlight now crashes because it's missing for the PHP integration.


BYK (approved):
Ouch, sorry and thanks!

PR #716, opened 2025-03-03
Move download actions into relevant docs

dcramer:
Its not overly useful to suggest folks download the app when there are additional steps to get Spotlight working.

- Add download actions and a bit of explainer to the about page.
- Add new headless environments setup guide.


dcramer:
image

PR #730, opened 2025-03-07
feat: Added Search support in Trace details

Shubhdeep12:
Before opening this PR:

- Added back Search input
- used context for search query(related https://github.com/getsentry/spotlight/pull/424#discussion_r1655526098)
  - we can extend this and make a context only for trace detail.
- added a hook for search input

### DIFF
--- a/packages/overlay/src/integrations/sentry/hooks/useSearchInput.ts
+++ b/packages/overlay/src/integrations/sentry/hooks/useSearchInput.ts
@@ -1,0 +1,32 @@
+import { FormEvent, useEffect, useState } from 'react';
+import useDebounce from './useDebounce';
+
+export default function useSearchInput(onSearch: (value: string) => void, delay = 500) {
+  const [inputValue, setInputValue] = useState('');
+  const [showReset, setShowReset] = useState(false);
+
+  const debouncedSearch = useDebounce(onSearch, delay);
+
+  useEffect(() => {
+    debouncedSearch(inputValue);
+  }, [inputValue, debouncedSearch]);
+
+  const handleChange = (e: FormEvent<HTMLInputElement>) => {
+    const value = e.currentTarget.value;
+    setInputValue(value);
+    setShowReset(!!value);
+  };
+
+  const handleReset = () => {
+    setInputValue('');
+    setShowReset(false);
+    onSearch('');
+  };
+
+  return {
+    inputValue,
+    showReset,
+    handleChange,
+    handleReset,
+  };
+}
--- a/packages/overlay/src/integrations/sentry/hooks/useSearchInput.ts
+++ b/packages/overlay/src/integrations/sentry/hooks/useSearchInput.ts
@@ -14,7 +14,7 @@ export default function useSearchInput(onSearch: (value: string) => void, delay 
   const handleChange = (e: FormEvent<HTMLInputElement>) => {
     const value = e.currentTarget.value;
     setInputValue(value);
-    setShowReset(!!value);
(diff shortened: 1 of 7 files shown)


packages/overlay/src/integrations/sentry/hooks/useSearchInput.ts
```diff
@@ -0,0 +1,32 @@
...
+  const debouncedSearch = useDebounce(onSearch, delay);
+
+  useEffect(() => {
+    debouncedSearch(inputValue);
+  }, [inputValue, debouncedSearch]);
+
+  const handleChange = (e: FormEvent<HTMLInputElement>) => {
+    const value = e.currentTarget.value;
+    setInputValue(value);
+    setShowReset(!!value);
```
BYK, line 17:
```suggestion
    setShowReset(Boolean(value));
```

packages/overlay/src/integrations/sentry/hooks/useSearchInput.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export default function useSearchInput(onSearch: (value: string) => void, delay = 500) {
+  const [inputValue, setInputValue] = useState('');
+  const [showReset, setShowReset] = useState(false);
+
+  const debouncedSearch = useDebounce(onSearch, delay);
+
+  useEffect(() => {
+    debouncedSearch(inputValue);
+  }, [inputValue, debouncedSearch]);
```
BYK, line 12: Maybe `useCallback` here? Not exactly sure what we are after to be honest

packages/overlay/src/integrations/sentry/hooks/useSearchInput.ts
```diff
@@ -0,0 +1,32 @@
...
+  const handleChange = (e: FormEvent<HTMLInputElement>) => {
+    const value = e.currentTarget.value;
+    setInputValue(value);
+    setShowReset(!!value);
+  };
+
+  const handleReset = () => {
+    setInputValue('');
+    setShowReset(false);
+    onSearch('');
```
BYK, line 23: Shouldn't this get triggered automatically with the state changes?

BYK (approved):
I wanna follow up with some UX improvements:

1. Should make the search bar sticky at the top for long/deep trees.
2. We should have an actual filtering functionality, not just highlight
3. We may wanna look into creating a query language of sorts or at least chaining filters with negation etc.

All that said, I think this is a great start and addition. Thanks and thanks for the speedy patch!

packages/overlay/src/integrations/sentry/hooks/useSearchInput.ts
Shubhdeep12: This is just calling the denounced function every time inputvalue changes.

packages/overlay/src/integrations/sentry/hooks/useSearchInput.ts
Shubhdeep12: No this is a function that will reset the values.
I don't think it will run when state changes and component rerenders

Shubhdeep12:
> I wanna follow up with some UX improvements:
> 
> 1. Should make the search bar sticky at the top for long/deep trees.
> 2. We should have an actual filtering functionality, not just highlight
> 3. We may wanna look into creating a query language of sorts or at least chaining filters with negation etc.
> 
> All that said, I think this is a great start and addition. Thanks and thanks for the speedy patch!

added these in a new issue.
Closing this PR for now.

Thanks for the review.

PR #739, opened 2025-03-22
fix: node filename format was wrong

tylersayshi:
Very minor quirk noticed with the stacktrace... node internals were being formatted incorrectly by `formatFilename`, so this change just returns them as is, since they are formatted well to begin with.

before: 

image

after: 

image


BYK (approved):
Great catch, thanks!

PR #740, opened 2025-03-22
Trace Detail Page improvements

Shubhdeep12:
Before opening this PR:

- Trace Detail Improvements
  - Removed Context tab 
  - Added Context on Trace details.
  - Divided Trace Details page in 2 Sections - Span Tree and Context of Trace.
  - Added a generic Resizer and extend its usage to SpanResizer and TreeResizer.
  - Fixed Overflow styling across overlay.

image
image


packages/overlay/src/integrations/sentry/components/explore/traces/TraceDetails/components/TraceTreeview.tsx
```diff
@@ -41,35 +89,8 @@ function TraceTreeviewContent({ traceId }: TraceTreeViewProps) {
...
-          tree={trace.spanTree}
-          startTimestamp={startTimestamp}
-          totalDuration={totalDuration}
-          totalTransactions={(trace.transactions || []).length}
-          spanNodeWidth={spanNodeWidth}
-          setSpanNodeWidth={setSpanNodeWidth}
-        />
-      </div>
+
+      <RenderTraceTree trace={trace} startTimestamp={startTimestamp} totalDuration={totalDuration} />
```
BYK, line 93: Like what you are doing here but `RenderTraceTree` is not a great component name (as it feels more like a function name). Why not just `TraceTree` or `TraceTreeWithSearch` or `TraceTreeContents`? I'm even okay with `TraceTreeInner` rather than `RenderTraceTree`

.changeset/spotty-emus-allow.md
```diff
@@ -0,0 +1,7 @@
+---
+'@spotlightjs/overlay': minor
+---
+
+- #731: Search Bar sticky and fixed overflow across the overlay.
+
+- Made Resizer a generic component and extended in SpanResizer.
```
BYK, line 7: Don't think this should be in the changelog as it is not relevant to end users?

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
```diff
@@ -76,7 +76,7 @@ export default function TransactionDetail({ showAll }: { showAll: boolean }) {
             <h1 className="truncate text-2xl font-bold">{atob(name!)}</h1>
           </div>
           <table className="divide-primary-700 w-full table-fixed divide-y">
-            <thead>
+            <thead className="bg-primary-950 sticky top-0 z-20">
```
BYK, line 79: Based on all this repetition, it feels like this should be a `DetailTable` component (or similar name). We can do this in a follow up.

BYK (approved):
Works great.

One thing I noticed: when holding the mouse down for resize, the mouse cursor does not stay in `resize` but goes back to the default arrow. Probably an existing bug.

Shubhdeep12:
> Works great.
> 
> One thing I noticed: when holding the mouse down for resize, the mouse cursor does not stay in `resize` but goes back to the default arrow. Probably an existing bug.

Improved the experience there @BYK.

packages/overlay/src/integrations/sentry/components/performance/TransactionDetail.tsx
Shubhdeep12: Added a generic table. 
you can review this now.
thanks

.changeset/spotty-emus-allow.md
Shubhdeep12: Removed

packages/overlay/src/integrations/sentry/components/explore/traces/TraceDetails/components/TraceTreeview.tsx
Shubhdeep12: Yeah, i've updated it.

BYK (approved):

packages/overlay/src/integrations/sentry/components/insights/Queries.tsx
```diff
@@ -58,8 +59,8 @@ const Queries = ({ showAll }: { showAll: boolean }) => {
 
   if (queriesData?.length) {
     return (
-      <table className="divide-primary-700 w-full table-fixed divide-y">
-        <thead>
+      <Table variant="detail">
+        <Table.Header>
```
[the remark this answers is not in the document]
Shubhdeep12: are you talking about having variant like queries, query summary, resources etc that will define QUERIES_HEADERS etc for the table?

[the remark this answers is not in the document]
BYK: Not a new variant but a way to standardize these and passing in a config object. This should allow the user of the `<Table>` component to use `Table.Header` and `Table.Body` sub-components. I also think we should rename this to `<DetailTable>` or something along those lines as it is easy to confuse with the native `<table>`.

In an ideal world, I'd just do something like the following:

```ts
<DetailTable headers={QUERIES_HEADERS}>
<tr>
  <td> .... </td>
</tr>
```

So `DetailTable` takes care of the `<thead>` and `<tbody>` parts for me.

Makes sense?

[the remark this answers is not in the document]
Shubhdeep12: Yeah thanks for the detailed response.
I was also thinking on the same lines.
I'll update this.

PR #741, opened 2025-03-23
Fix/trace improvements gh issues

### LINKED ISSUE #738: Span hover highlight does not highlight the duration panel on the right
When hovering over the span list, we only highlight the left panel where the description is (leaving out the right panel where we show the duration). This makes it hard to find the span to dig into on a long trace. We should highlight both panels.

Image

Not immediately clear from the above image but look how the left side is fully highlighted where nothing changes from the neutral state on the right panel.

Shubhdeep12:
Before opening this PR:

image
fixes #738

### DIFF
--- a/.changeset/two-students-sip.md
+++ b/.changeset/two-students-sip.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/overlay': patch
+---
+
+Fixed #738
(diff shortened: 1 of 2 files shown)


BYK (approved):
Works beautifully, thanks!

Issue #731, opened 2025-03-08
UX improvements in Search


Shubhdeep12:
- [x] Should make the search bar sticky at the top for long/deep trees. (#740)
- [x] We should have an actual filtering functionality, not just highlight (#747)
- [ ] #779

_Originally posted by @BYK in https://github.com/getsentry/spotlight/pull/730#pullrequestreview-2668226664_

Shubhdeep12:
Can we allow both highlighting and filtering of matched spans, giving users an option to toggle "**_Show Only Matched_**" after search?

Both approaches serve the purpose of searching in different ways.

BYK:
Agreed, one will be filter and the other would be search. Related: #745

BYK:
Closing as the 2 major things are done. Will create more issues for further stuff, such as #779.

PR #747, opened 2025-03-24
feat: filter spans on search

betegon:
Related to #731 
>We should have an actual filtering functionality, not just highlight

These changes add filtering to span search
- uplifted search functionality from `SpanItem` to `SpanTree`.
- Implemented DFS to handle nested search.
- Also removed highlighting from matching spans.

The results display not only the direct matches but also their hierarchical context up to the root: 

Screenshot 2025-03-24 at 22 24 09
Screenshot 2025-03-24 at 22 24 31
Screenshot 2025-03-24 at 22 23 58

Since spans are time-sorted, it would be useful to add a feature like `+<#items>` between results. This would allow you to jump directly to the result you’re looking for when debugging. I’ve been using Spotlight for the past few days, and this addition would have been very helpful.

Screenshot 2025-03-24 at 22 30 05

Before opening this PR:

### DIFF
--- a/packages/overlay/src/integrations/sentry/components/explore/traces/spans/SpanTree.tsx
+++ b/packages/overlay/src/integrations/sentry/components/explore/traces/spans/SpanTree.tsx
@@ -1,3 +1,5 @@
+import { useMemo } from 'react';
+import { useSearch } from '~/integrations/sentry/context/SearchContext';
 import classNames from '../../../../../../lib/classNames';
 import type { Span, TraceContext } from '../../../../types';
 import SpanItem from './SpanItem';
@@ -21,10 +23,25 @@ export default function SpanTree({
   spanNodeWidth: number;
   setSpanNodeWidth?: (val: number) => void;
 }) {
+  const { query } = useSearch();
+
+  const matchesQuery = (span: Span): boolean | undefined => {
+    return span.span_id.includes(query) || span.op?.includes(query) || span.description?.includes(query);
+  };
+
+  const hasMatchingDescendant = (span: Span): boolean => {
+    if (matchesQuery(span)) return true;
+    if (!span.children) return false;
+    return span.children.some(hasMatchingDescendant);
+  };
+
+  const filteredTree = useMemo(() => (query ? tree.filter(span => hasMatchingDescendant(span)) : tree), [query, tree]);
+
   if (!tree || !tree.length) return null;
+
   return (
     <ul className={classNames(tree.length > 1 && 'deep', 'tree')}>
-      {tree.map(span => {
+      {filteredTree.map(span => {
         return (
           <SpanItem
             key={span.span_id}
--- a/.changeset/sweet-poets-reflect.md
+++ b/.changeset/sweet-poets-reflect.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/overlay': patch
+---
+
+Include filter functionality when searching spans
--- a/packages/overlay/src/integrations/sentry/components/explore/traces/spans/SpanItem.tsx
+++ b/packages/overlay/src/integrations/sentry/components/explore/traces/spans/SpanItem.tsx
@@ -1,7 +1,6 @@
-import { useMemo, useRef, useState } from 'react';
+import { useRef, useState } from 'react';
 import { Link, useParams } from 'react-router-dom';
 import { ReactComponent as ChevronIcon } from '~/assets/chevronDown.svg';
-import { useSearch } from '~/integrations/sentry/context/SearchContext';
 import classNames from '../../../../../../lib/classNames';
 import type { Span, TraceContext } from '../../../../types';
 import { getFormattedDuration, getSpanDurationClassName } from '../../../../utils/duration';
@@ -29,7 +28,6 @@ const SpanItem = ({
   setSpanNodeWidth?: (val: number) => void;
 }) => {
   const { spanId } = useParams();
-  const { query } = useSearch();
   const containerRef = useRef<HTMLLIElement>(null);
   const childrenCount = span.children ? span.children.length : 0;
   const [isItemCollapsed, setIsItemCollapsed] = useState(
@@ -51,17 +49,12 @@ const SpanItem = ({
       setSpanNodeWidth(newLeftWidth);
     }
   };
-  const isQueried = useMemo(() => {
-    if (!query) return false;
-    return span.span_id.includes(query) || span.op?.includes(query) || span.description?.includes(query);
-  }, [query, span.span_id, span.op, span.description]);
 
   return (
     <li key={span.span_id} ref={containerRef}>
       <Link
         className={classNames(
           'hover:bg-primary-700 group flex rounded-sm text-sm',
-          isQueried ? 'bg-primary-200 bg-opacity-20' : '',
           spanId === span.span_id ? 'bg-primary-900' : '',
           span.tags?.source === 'profile' ? 'text-lime-500' : '',
         )}
@@ -73,7 +66,6 @@ const SpanItem = ({
         <div
           className={classNames(
             'node group-hover:bg-primary-700 rounded-sm',
-            isQueried ? 'bg-transparent' : '',
             span.status && span.status !== 'ok' ? 'text-red-400' : '',
             spanId === span.span_id ? 'bg-primary-900' : 'bg-primary-950',
           )}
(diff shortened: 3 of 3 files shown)


Shubhdeep12:
https://github.com/getsentry/spotlight/issues/731#issuecomment-2750165973

packages/overlay/src/integrations/sentry/components/explore/traces/spans/SpanTree.tsx
```diff
@@ -21,10 +23,25 @@ export default function SpanTree({
   spanNodeWidth: number;
   setSpanNodeWidth?: (val: number) => void;
 }) {
+  const { query } = useSearch();
+
+  const matchesQuery = (span: Span): boolean | undefined => {
+    return span.span_id.includes(query) || span.op?.includes(query) || span.description?.includes(query);
+  };
```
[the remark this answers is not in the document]
betegon: Done! I also made sure that it returns a boolean:

```typescript
const matchesQuery = useCallback(
    (span: Span): boolean => {
      const q = context.query.toLowerCase();
      return (
        (span.span_id.toLowerCase().includes(q) ||
          span.op?.toLowerCase().includes(q) ||
          span.description?.toLowerCase().includes(q)) ??
        false
      );
    },
    [context.query],
  );
  ```
  
P.S. I added case-insensitive matching. I believe it’s one of the first things anyone would expect from a search feature.

File: packages/overlay/src/integrations/sentry/components/explore/traces/spans/SpanTree.tsx [hunk 2]
```diff
@@ -21,10 +23,25 @@ export default function SpanTree({
...
+
+  const hasMatchingDescendant = (span: Span): boolean => {
+    if (matchesQuery(span)) return true;
+    if (!span.children) return false;
+    return span.children.some(hasMatchingDescendant);
+  };
```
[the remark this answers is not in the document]
betegon: Good call! I’ve memoized each of the spans we process now.  
```typescript
if (showOnlyMatched) {
  const spanMemo = new Map<string, boolean>();
  const hasMatchingDescendant = (span: Span): boolean => {
    if (spanMemo.has(span.span_id)) return spanMemo.get(span.span_id)!;
    const result = matchesQuery(span) || (span.children?.some(child => hasMatchingDescendant(child)) ?? false);
    spanMemo.set(span.span_id, result);
    return result;
  };
  return tree.filter(span => hasMatchingDescendant(span));
}
```

betegon:
Thanks for the detailed feedback, @BYK  I’ve implemented your suggestions, including:
- Moved `matchesQuery` inside a `useCallback` in `useSearch`.
- Added case-insensitive matching to the search.
- Memoized spans to avoid processing each one more than once for each query.
- Added a toggle to switch between highlighting spans and showing only matches . I didn’t find a similar pattern in the Overlay, so I added the toggle inline to avoid taking up extra vertical space. cc: @Shubhdeep12 https://github.com/getsentry/spotlight/issues/731#issuecomment-2750165973

I think we’re in a better spot now, happy to keep iterating if you have more suggestions 😉

Here’s a quick screenshot video showing how the toggle works.

https://github.com/user-attachments/assets/13270f47-1fa0-4bb0-b328-aa0720f56cc1

packages/overlay/src/integrations/sentry/components/traces/TraceDetails/components/TraceTreeview.tsx
```diff
@@ -24,32 +24,43 @@ function TraceTreeWithSearch({
...
+        </div>
+        <button
+          className={`rounded border px-2 py-1 text-sm ${
+            showOnlyMatched
+              ? 'bg-primary-600 border-white text-white'
+              : 'bg-primary-800 text-primary-300  border-primary-500 hover:bg-primary-700'
+          }`}
+          onClick={() => setShowOnlyMatched(!showOnlyMatched)}
+        >
+          Only Show Matches
```
BYK, line 62: I think it would be better to switch this into a checkbox and make the text shorter (or even use an icon with a very short label)

packages/overlay/src/integrations/sentry/components/traces/spans/SpanItem.tsx
```diff
@@ -51,10 +51,8 @@ const SpanItem = ({
       setSpanNodeWidth(newLeftWidth);
     }
   };
-  const isQueried = useMemo(() => {
-    if (!query) return false;
-    return span.span_id.includes(query) || span.op?.includes(query) || span.description?.includes(query);
-  }, [query, span.span_id, span.op, span.description]);
+
+  const isQueried = !showOnlyMatched && query && matchesQuery(span);
```
BYK, line 55: I think renaming this to `shouldHighlight` makes more sense with the new approach.
```suggestion
  const shouldHighlight = !showOnlyMatched && query && matchesQuery(span);
```

packages/overlay/src/integrations/sentry/components/traces/spans/SpanTree.tsx
```diff
@@ -21,10 +23,29 @@ export default function SpanTree({
   spanNodeWidth: number;
   setSpanNodeWidth?: (val: number) => void;
 }) {
+  const { query, matchesQuery, showOnlyMatched } = useSearch();
+
+  const filteredTree = useMemo(() => {
+    if (!query) return tree;
```
BYK, line 29: Early return here with `showOnlyMatched` would make it easier to follow.
```suggestion
    if (!query || !showOnlyMatched) return tree;
```

packages/overlay/src/integrations/sentry/components/traces/spans/SpanTree.tsx
```diff
@@ -21,10 +23,29 @@ export default function SpanTree({
...
+    if (showOnlyMatched) {
+      const spanMemo = new Map<string, boolean>();
```
BYK, line 31: This implementation is not bad but I don't think the cache would be shared across children/siblings. I think this cache should be moved to `useSearch()` level and get reset when `query` changes.

packages/overlay/src/integrations/sentry/components/traces/spans/SpanTree.tsx
```diff
@@ -21,10 +23,29 @@ export default function SpanTree({
...
+        spanMemo.set(span.span_id, result);
+        return result;
+      };
+
+      return tree.filter(span => hasMatchingDescendant(span));
+    }
+    return tree;
+  }, [query, tree, showOnlyMatched, matchesQuery]);
+
   if (!tree || !tree.length) return null;
```
BYK, line 44: Shouldn't we replace the `tree` here with `filteredTree`?
```suggestion
  if (!filteredTree|| !tree.filteredTree) return null;
```

packages/overlay/src/integrations/sentry/context/SearchContext.tsx
```diff
@@ -1,22 +1,43 @@
...
-  return context;
+
+  const matchesQuery = useCallback(
+    (span: Span): boolean => {
+      const q = context.query.toLowerCase();
+      return (
+        (span.span_id.toLowerCase().includes(q) ||
+          span.op?.toLowerCase().includes(q) ||
+          span.description?.toLowerCase().includes(q)) ??
+        false
```
BYK, line 36: I think it would be more efficient if we create a:

```js
const queryMatcher = new RegExp(RegExp.escape(query), 'i');
```

and then do `queryMatcher.test(span.op)` etc. afterwards. You may need to polyfill [`RegExp.escape`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape) as it doesn't seem to have wide browser support without flags.

Finally, a bit of a nitpick, we may wanna limit matching to only when the query is more than 2 characters as I don't want to match on `a` or even `ab` probably.

packages/sidecar/src/contextlines.ts
```diff
@@ -1,8 +1,8 @@
+import { LEAST_UPPER_BOUND, TraceMap, originalPositionFor, sourceContentFor } from '@jridgewell/trace-mapping';
```
BYK, line 1: Let's not change any other files outside of the context of this patch. I'll revert this for you.

BYK (approved):
I think we may wanna expand all trees when there's a query set and "only show matches" is selected as otherwise it gets a bit confusing when trying to figure out why a top-level span was not filtered (because it has some child matching the query but is collapsed so the child is not visible).

I also have all these notes but I think the PR is a net improvement so gonna approve, merge, and cut a new release and push the improvements to future releases.

PR #748, opened 2025-03-24
feat (sidecar): show url for existing sidecar

betegon:
Small DX improvement: sidecar now always logs the Spotlight URL, even if it was already running. Makes it easy to _cmd/ctrl+click_ and open in browser.

Previous behavior:
Screenshot 2025-03-23 at 20 18 37

New behavior (notice the final line):
Screenshot 2025-03-23 at 20 19 23

Before opening this PR:

- [ x] I added a [Changeset Entry](https://spotlightjs.com/contribute/changesets/) with `pnpm changeset:add`

### DIFF
--- a/.changeset/perfect-laws-burn.md
+++ b/.changeset/perfect-laws-burn.md
@@ -1,0 +1,6 @@
+---
+'@spotlightjs/sidecar': patch
+---
+
+Improve DX by always showing the Spotlight URL, even if the sidecar was already running. Makes it easy to cmd/ctrl+click
+and open in browser.
--- a/.changeset/serious-donkeys-try.md
+++ b/.changeset/serious-donkeys-try.md
@@ -1,0 +1,5 @@
+---
+'@spotlightjs/overlay': patch
+---
+
+format node internals correctly in stacktrace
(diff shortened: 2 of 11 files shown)


Shubhdeep12:
hey @betegon thanks for the contribution.

just wondering what if there is custom `filesToServe` while setupSidecar. @BYK any thoughts?

otherwise LGTM.

BYK (changes requested):
Good point @Shubhdeep12!

@betegon - Sidecar can be used standalone where it may not ship with the built-in overlay UI. In that case that link would not work. Do you think you can detect that case easily and avoid printing a URL for that?

If not we may wanna punt on this change.

betegon:
Wuuups! My bad, didn’t account for that.

I went with checking if `index.html` actually exists, since just doing `filesToServe || basePath` felt too broad; it might log the URL even if `filesToServe` is defined without the UI, or `basePath` is set but the files aren’t actually there.

```typescript
const hasSpotlightUI = (filesToServe && '/src/index.html' in filesToServe) || (!filesToServe && basePath);
```

Happy to throw this out the window if it feels too fragile or unnecessary, just thought it might make it a bit easier for users.

BYK (approved):
Love the solution!

PS: basePath might be obsolete but I'll check that later.

PR #751, opened 2025-03-25
Fix: Parsing of envelope data

Shubhdeep12:
Before opening this PR:

We should parse the envelope data to string while showing raw envelope data.

Earlier
image

Now
image


BYK (approved):
Ah, this has been annoying me so much but then got off my mind. Thanks!

Issue #757, opened 2025-04-02
Trace & Error count in tabs don't update accurately


BYK:
### Steps to Reproduce

1. Launch Spotlight in dev mode: `pnpm dev`
2. Open the browser for Spotlight web UI
3. Go to fixtures directory: `cd packages/overlay/_fixtures`
4. Inject all the sample envelopes: `./send_to_sidecar.cjs`

### Expected Result

**Explore** tab should show `17` and **Errors** tab should show `7`.

### Actual Result

**Explore** and **Errors** tabs show `13` and `3` respectively. They update to correct values upon any kind of navigation.

BYK:
I think we should wrap tab generation code from integrations into a `useMemo` here: https://github.com/getsentry/spotlight/blob/6aea29bd34b836e86e08ceb040003cb8cb89f622/packages/overlay/src/components/Overview.tsx#L20-L29

I tried that but using `integrationData` as the dependencies for `useMemo` doesn't really work as the object doesn't change but it's values do. So I actually got worse results. Not sure what's the best fix for this TBH.

Shubhdeep12:
@BYK yeah looks like data is not updated in sentry-integration before rendering the UI causing mismatch in the count.

We can move the tabs to a state and wait for integration to finalized the updates before rendering on the UI.

BYK:
@Shubhdeep12 oh, so you are saying we need to refactor how we do integrations. That actually makes sense. We probably need to allow them to pass a "on data updated" hook function upstream. And I think switching to zustand fist would make that a lot easier?

PR #764, opened 2025-04-05
Fix: integration data update

Shubhdeep12:
Before opening this PR:

#757 

Added support for waiting for data updates from integration

### DIFF
--- a/.changeset/slimy-poets-sparkle.md
+++ b/.changeset/slimy-poets-sparkle.md
@@ -1,0 +1,6 @@
+---
+'@spotlightjs/overlay': minor
+'@spotlightjs/website': minor
+---
+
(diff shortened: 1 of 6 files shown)


BYK (changes requested):
I guess this solves the problem but it adds a lot of complexity and seems to aim only at the initial load.

I'll have to think more about how to propagate all the changes through regular React states etc rather than these explicit blocking mechanism.

I'm guessing the main issue is how we constructed sentryDataCache singleton.

Shubhdeep12:
Yeah @BYK 
As discussed let's try moving from this simple singleton state to a reactive approach( zustand)  in sentry integration. 
Closing this PR for now.

Shubhdeep12:
Thanks for the review

PR #776, opened 2025-04-17
feat(sidecar): use nanosecond timestamp for captured filenames

### LINKED ISSUE #775: bug: sidecar capture filename conflict
### Environment

- node v22.14.0
- spotlight v2.12.0
- sidecar v1.11.2

### Steps to Reproduce

1. run sidecar in capture mode (`SPOTLIGHT_CAPTURE=1`)
2. send a burst of events (i.e. test suite)

### Expected Result

each capture should be saved to a unique file.

### Actual Result

some files are written multiple times with different content, leading to data loss due to overwrites.

Image

cruzdanilo:
fixes #775

### DIFF
--- a/packages/sidecar/src/main.ts
+++ b/packages/sidecar/src/main.ts
@@ -179,7 +179,7 @@ const streamRequestHandler = (buffer: MessageBuffer<Payload>, incomingPayload?: 
         }
 
         if (process.env.SPOTLIGHT_CAPTURE || incomingPayload) {
-          const timestamp = new Date().getTime();
+          const timestamp = BigInt(Date.now()) * 1_000_000n + (process.hrtime.bigint() % 1_000_000n);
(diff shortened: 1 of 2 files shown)


packages/sidecar/src/main.ts
```diff
@@ -179,7 +179,7 @@ const streamRequestHandler = (buffer: MessageBuffer<Payload>, incomingPayload?:
         }
 
         if (process.env.SPOTLIGHT_CAPTURE || incomingPayload) {
-          const timestamp = new Date().getTime();
+          const timestamp = BigInt(Date.now()) * 1_000_000n + (process.hrtime.bigint() % 1_000_000n);
```
BYK, line 182: I'd love to use [uuidv7](https://uuid7.com/) for this but what you do works and is dependency-free so 🤷🏻‍♂️

BYK (approved):
Thanks so much!

packages/sidecar/src/main.ts
cruzdanilo: it also keeps the files chronologically sorted

Issue #694, opened 2025-02-05
Add flamegraphs for profiles


BYK:
We now have support for profiles (#565) and will have support for continuous profiling (#567) so we'd better have an aggregate profile view, where a flamegraph is the industry standard. These will ideally have their dedicated tab.

This component may help: [getsentry/sentry@master/static/app/components/profiling/flamegraph/flamegraphPreview.tsx](https://github.com/getsentry/sentry/blob/master/static/app/components/profiling/flamegraph/flamegraphPreview.tsx?rgh-link-date=2024-11-29T09%3A35%3A29Z)

BYK:
We should also try a sunburst chart: https://github.com/vasturiano/react-sunburst-chart

BYK:
@tylersayshi shared this library: https://github.com/antfu/nanovis and it looks awesome.

[![Image](https://github.com/user-attachments/assets/f6462e00-9c3e-49d5-a2ff-b83cafb4e3e3)](https://node-modules.dev/chart/sunburst#selected=@spotlightjs/spotlight@2.12.0&install=@spotlightjs/spotlight)

It even supports on-the-fly switching between sunburst, flamegraph, _and_ a tree map!

Shubhdeep12:
hey @BYK ,

I think [nanovis](https://github.com/antfu/nanovis) is a great option for visualisation.

I did a quick PoC that parses the profile and maps it to a Flamegraph and other visualisations:

https://github.com/user-attachments/assets/95ad4cd0-2fe2-4557-8f6d-4b7f67afb3b5


The data shown here is from a single profile, it's not an aggregated Flamegraph.
this is the direction I was thinking of for integrating profile visualisation. 
the UI is still a rough draft and will be improved in the final version.

cc @betegon

Shubhdeep12:
hey @BYK @betegon did you get a chance to check this?

betegon:
@Shubhdeep12  had a busy week, let's see how this looks like ;)

Shubhdeep12:
That's great!
I'll improve it a little and raise a PR then

thanks for the review.

PR #772, opened 2025-04-15
feat: Refactor Sentry integration to use Zustand for state management

Shubhdeep12:
Before opening this PR:

- [x] I added a [Changeset Entry](https://spotlightjs.com/contribute/changesets/) with `pnpm changeset:add`
- [x] I referenced issues that this PR addresses
#771 
To use Zustand in sentry-integration

PR #786, opened 2025-04-24
feat(overlay): Profiles view

betegon:
##  Adds “Profiles” tab view

Related to #694 

Screenshot 2025-04-24 at 14 25 18

### Added
- Introduces a new **Profiles** tab to inspect the most time-consuming functions.
- Displays function names alongside a visual duration bar (`TimeBar` component).

### Highlights
- Flamegraph support will be added in a future iteration, but this view already makes it easy to spot performance hotspots.
- Implements a reusable `TimeBar` component (props: `value`, `maxValue`, optional `title`/`className`), _thanks to @BYK’s suggestion_,  for use across other _Insights_ tabs (e.g., Queries).
-  `TimeBar` percentages are based on the longest function, so as you can see in the screenshot above, `<unknown>` is assigned 100% and all other bars scale down relative to it.

Before opening this PR:


Shubhdeep12:
Hey @betegon, can you please rebase it and resolve the merge conflicts?
We just merged this PR - https://github.com/getsentry/spotlight/pull/772
where we moved to the Zustand for the Sentry integration data store.

betegon:
Done @Shubhdeep12! 
I took some time diving into your Zustand PR to get a better grasp, great work on it! 😉

quick q:  you’re calling the `SentryStoreActions` via `useSentryStore.getState()` rather than through the hook (e.g `const {getAction} = useSentryStore()`. Is that intentional to avoid subscribing and prevent any unnecessary re-renders (even though they probably wouldn’t happen)?

Shubhdeep12:
Hey @betegon 
main intent of using .getState() was to use it where the file is not a react component. Because as you mentioned it's a hook and I think it can only be used in the react component.

.changeset/fluffy-cougars-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spotlightjs/overlay': minor
+---
+
+add a Profiles tab
```
BYK, line 5:
```suggestion
Add a Profiles tab under insights
```

packages/overlay/src/integrations/sentry/components/insights/FunctionProfiles.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return 0;
+  },
+};
+
+function FunctionProfiles() {
+  const { sort, toggleSortOrder } = useSort({ defaultSortType: FUNCTION_PROFILES_SORT_KEYS.timeSpent });
+
+  const functionProfiles = useMemo(() => {
+    const profiles = useSentryStore.getState().getFunctionProfiles();
+    const compareProfileInfo = COMPARATORS[sort.active] || COMPARATORS[FUNCTION_PROFILES_SORT_KEYS.timeSpent];
```
BYK, line 35: Future: We should limit the values `sort.active` can take by making `useSort` use a template so we don't have to do this unnecessary fallback.

packages/overlay/src/integrations/sentry/components/insights/FunctionProfiles.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    return profiles.sort((a, b) => {
+      return sort.asc ? compareProfileInfo(a, b) : compareProfileInfo(b, a);
+    });
+  }, [sort]);
+
+  if (!functionProfiles?.length) {
```
BYK, line 42:
```suggestion
  if (!functionProfiles.length) {
```

packages/overlay/src/integrations/sentry/components/insights/FunctionProfiles.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return <p className="text-primary-300 px-6 py-4">No profiles found.</p>;
+  }
+
+  // Calculate max time for bar visualization (100%, scaling form here)
+  const maxTime = Math.max(...functionProfiles.map(profile => profile.totalTime));
+
+  return (
+    <table className="divide-primary-700 w-full table-fixed divide-y">
```
BYK, line 50: We'll need to switch to `<Table>` from `import Table from '~/ui/Table';` for this.

packages/overlay/src/integrations/sentry/components/insights/FunctionProfiles.tsx
```diff
@@ -0,0 +1,111 @@
...
+            <td className="text-primary-200 w-[15%] whitespace-nowrap px-6 py-4 text-right text-sm font-medium">
+              <span className={getSpanDurationClassName(profile.totalTime)}>
+                {getFormattedDuration(profile.totalTime)}
+              </span>
+            </td>
+            <td className="text-primary-200 w-[15%] whitespace-nowrap px-6 py-4 text-right text-sm font-medium">
+              {profile.samples}
+            </td>
+            <td className="text-primary-200 w-[15%] whitespace-nowrap px-6 py-4 text-right text-sm font-medium">
+              {profile.traceId.substring(0, 8)}
```
BYK, line 102: We should probably create a `<TraceLink>` helper and standardize this (plus making it a link)

packages/overlay/src/integrations/sentry/components/insights/index.tsx
```diff
@@ -51,6 +53,7 @@ export default function InsightsTabDetails() {
           {/* Default tab */}
           <Route path="queries" element={<Queries showAll={showAll} />} />
           <Route path="*" element={<Navigate to="/insights/queries" replace />} />
+          <Route path="functionprofiles" element={<FunctionProfiles />} />
```
BYK, line 56: I'll rename this to just Profiles for now.

packages/overlay/src/integrations/sentry/components/shared/TimeBar.tsx
```diff
@@ -0,0 +1,21 @@
...
+  className?: string;
+};
+
+export function TimeBar({ value, maxValue, title, className }: TimeBarProps) {
+  const percentage = (value / maxValue) * 100;
+
+  return (
+    <div className={classNames('relative text-left text-sm font-medium', className)}>
+      {title && <div className="truncate text-lime-500">{title}</div>}
+      <div className={classNames('bg-primary-800 h-1 w-full overflow-hidden rounded-full', title ? 'mt-1' : '')}>
```
BYK, line 16: Love how simple this component is yet how beautiful it looks

packages/overlay/src/integrations/sentry/constants.ts
```diff
@@ -1,5 +1,36 @@
...
+    sortKey: FUNCTION_PROFILES_SORT_KEYS.timeSpent,
+  },
+  {
+    id: 'samples',
+    title: 'Samples',
+    sortKey: FUNCTION_PROFILES_SORT_KEYS.samples,
+  },
+  {
+    id: 'profile',
+    title: 'Profile',
```
BYK, line 29: This should probably say `Trace`?

packages/overlay/src/integrations/sentry/types.ts
```diff
@@ -158,6 +158,14 @@ export type SentryProcessedProfile = SentryProfile & {
   samples: ProcessedProfileSample[];
 };
 
+export type FunctionProfile = {
```
BYK, line 161:
```suggestion
export type AggregateCallProfile = {
```

packages/overlay/src/integrations/sentry/data/sentryStore.ts
```diff
@@ -410,6 +412,60 @@ const useSentryStore = create<SentryStoreState & SentryStoreActions>()((set, get
...
+
+      // Go over each profile sample
+      for (let sampleIdx = 0; sampleIdx < profile.samples.length - 1; sampleIdx++) {
+        const sample = profile.samples[sampleIdx];
+        const nextSample = profile.samples[sampleIdx + 1];
+        const duration = nextSample.start_timestamp - sample.start_timestamp;
+
+        const stackId = sample.stack_id;
+        const frameIndices = profile.stacks[stackId];
+        if (!frameIndices?.length) continue;
```
BYK, line 430: This indicates a data error so I'd rather crash

packages/overlay/src/integrations/sentry/data/sentryStore.ts
```diff
@@ -410,6 +412,60 @@ const useSentryStore = create<SentryStoreState & SentryStoreActions>()((set, get
...
+
+        // go over each stack frame (frame = 1 func/meth call)
+        for (const frameIdx of frameIndices) {
+          const frame = profile.frames[frameIdx];
+          if (!frame) continue;
```
BYK, line 435: Same as above, this indicates a data error which I think we should just crash on

packages/overlay/src/integrations/sentry/data/sentryStore.ts
```diff
@@ -410,6 +412,60 @@ const useSentryStore = create<SentryStoreState & SentryStoreActions>()((set, get
...
+
+          const funcName =
+            frame.function ||
+            (frame.module
+              ? `${frame.module}:<anonymous>`
+              : frame.filename
+                ? `${frame.filename}:${frame.lineno || '?'}`
+                : '<unknown>');
+
```
BYK, line 444: We should standardise this.

packages/overlay/src/integrations/sentry/data/sentryStore.ts
```diff
@@ -410,6 +412,60 @@ const useSentryStore = create<SentryStoreState & SentryStoreActions>()((set, get
...
+          const existing = functionProfiles.get(funcName) || {
+            name: funcName,
+            totalTime: 0,
+            samples: 0,
+            frames: [],
+            traceId, //
```
BYK, line 450: We gotta revisit this as a function may be present in multiple traces. This also requires some updates to the UI

packages/overlay/src/integrations/sentry/data/sentryStore.ts
```diff
@@ -410,6 +412,60 @@ const useSentryStore = create<SentryStoreState & SentryStoreActions>()((set, get
...
+          };
+
+          existing.totalTime += duration;
+          existing.samples += 1;
+          if (!existing.frames.includes(frame)) {
+            existing.frames.push(frame);
+          }
```
BYK, line 457: Why not use a `Set()` which does exactly the same thing but way more efficiently? Also not sure why we are actually storing the frames? I also see an opportunity to optimize things a lot more. I'll see if I can get these done.

BYK (changes requested):
I'll work on the suggestions I made and then merge this. Looks great, thanks!

packages/overlay/src/integrations/sentry/data/sentryStore.ts
```diff
@@ -410,6 +414,45 @@ const useSentryStore = create<SentryStoreState & SentryStoreActions>()((set, get
...
   },
 
+  getAggregateCallData(): AggregateCallData[] {
+    const aggregateCalls = new Map<string, AggregateCallData>();
+    for (const [traceId, profile] of this.profilesByTraceId) {
+      for (let sampleIdx = 0; sampleIdx < profile.samples.length - 1; sampleIdx++) {
+        const sample = profile.samples[sampleIdx];
+        const nextSample = profile.samples[sampleIdx + 1];
+        // TODO: Handle the case where nextSample is undefined -- use the end of the profile or associated trace
+        const duration = nextSample.start_timestamp - sample.start_timestamp;
```
Shubhdeep12, line 424:
```suggestion
        const duration = Math.max(0, nextSample.start_timestamp - sample.start_timestamp)
```

just a safety clamp in case of any issue in sync

packages/overlay/src/integrations/sentry/data/sentryStore.ts
```diff
@@ -410,6 +414,45 @@ const useSentryStore = create<SentryStoreState & SentryStoreActions>()((set, get
...
+        // TODO: Keep a running average based on continuous samples -- as in where we keep seeing the same
+        //       function name / frame back to back
+
+        const stackId = sample.stack_id;
+        const frameIndices = profile.stacks[stackId];
```
Shubhdeep12, line 429:
```suggestion
        const frameIndices = profile.stacks[stackId] || [];
```

packages/overlay/src/integrations/sentry/components/shared/TimeBar.tsx
```diff
@@ -0,0 +1,21 @@
...
+  title?: string;
+  className?: string;
+};
+
+export function TimeBar({ value, maxValue, title, className }: TimeBarProps) {
+  const percentage = (value / maxValue) * 100;
+
+  return (
+    <div className={classNames('relative text-left text-sm font-medium', className)}>
+      {title && <div className="truncate text-lime-500">{title}</div>}
```
Shubhdeep12, line 15:
```suggestion
      {title && <div className="truncate text-lime-500" title={title}>{title}</div>}
```

Shubhdeep12:
Can we create a hook to retrieve `getAggregateCallData` from local or all traces
It looks like we are showing profiles from all traces right now.
image

betegon:
appreciate you stepping in! The changes look great 📈

packages/overlay/src/integrations/sentry/data/sentryStore.ts
BYK: Don't think we should do this as samples should always be sorted. If not I'd prefer to know about that somehow rather than masking it with 0.

packages/overlay/src/integrations/sentry/data/sentryStore.ts
BYK: Similar to others: if we hit this fallback case, I'd rather us throw an exception to see the faulty data.

BYK:
@Shubhdeep12 
>Can we create a hook to retrieve getAggregateCallData from local or all traces
>It looks like we are showing profiles from all traces right now.

We are showing profiles from all traces as this is intentional. I think keeping it this way for now is better. I think the useful bits would be per-trace specific analysis and total analysis OR an analysis matching a search query which we probably should add (including a local trace selector somehow)

BYK (approved):
Let's go!

BYK:
@betegon @Shubhdeep12 we should follow up this by aligning the Queries tab with the Profiles tab:

1. We should use the new TimeBar component there
2. Rename "Time Spent" to "Total Time"

I wonder if we can use `TimeBar` on other pages such as "Resources"

Another follow up would be to add a "Function/Frame detail" page (or just a new column?) where we list all the traces it appears and then link to that specific span in the right trace (finding that span id might be tricky but should be possible).

Issue #773, opened 2025-04-16
to have getLocalTraces Hook in sentry-integration


Shubhdeep12:
I think it might be worthwhile to provide a `getLocalTraces()` helper and refactor other places to show a "Local" badge to use something else. (this is for a follow up, not for this PR)

_Originally posted by @BYK in https://github.com/getsentry/spotlight/pull/772#discussion_r2044318712_

PR #789, opened 2025-04-26
feat(sentry): Refactored and added slices in Sentry store

Shubhdeep12:
Before opening this PR:


.changeset/famous-dryers-sing.md
```diff
@@ -0,0 +1,6 @@
+---
+'@spotlightjs/overlay': minor
```
BYK, line 2: This is a refactor and doesn't really add anything new, right?

```suggestion
'@spotlightjs/overlay': patch
```

.changeset/famous-dryers-sing.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+- Sentry Store Slices
+- fixes: #773: Added a helper for getting localTrces and removed redundant isTraceLocal helper
```
BYK, line 6: Remember that this is the user-facing side of changelogs, so maybe keep it simpler?

```suggestion
Refactor the data store for Sentry integration with slices
```

packages/overlay/src/integrations/sentry/components/events/EventList.tsx
```diff
@@ -14,16 +14,16 @@ import { EventSummary } from './Event';
...
+  const { isTraceLocal } = useSentryStore();
 
   const matchingEvents = events.filter(isErrorEvent);
 
   const [showAll, setShowAll] = useState(!context.experiments['sentry:focus-local-events']);
   const filteredEvents = showAll
     ? matchingEvents
     : matchingEvents.filter(
-        e => (e.contexts?.trace?.trace_id ? helpers.isLocalToSession(e.contexts?.trace?.trace_id) : null) !== false,
+        e => (e.contexts?.trace?.trace_id ? isTraceLocal(e.contexts?.trace?.trace_id) : null) !== false,
```
BYK, line 26: What I meant in #773 was to have a helper that automatically filters and only returns the local traces, hence the name `getLocalTraces()`.

What do you think, would it add too much complexity?

packages/overlay/src/integrations/sentry/data/useSentryHelpers.tsx
```diff
@@ -1,14 +1,13 @@
 import { useContext } from 'react';
+import useSentryStore from '../store';
 import { SentryEventsContext } from './sentryEventsContext';
-import useSentryStore from './sentryStore';
 
 export const useSentryHelpers = () => {
   useContext(SentryEventsContext);
-  const isTraceLocal = useSentryStore(state => state.isTraceLocal);
+  const { isTraceLocal, getTraces } = useSentryStore();
```
BYK, line 7: Kind of confused about why `isTraceLocal` living in `store`. Should it be something like `store-helpers`?

packages/overlay/src/integrations/sentry/store/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import { createEventsSlice } from './slices/eventsSlice';
+import { createProfilesSlice } from './slices/profilesSlice';
+import { createSDKsSlice } from './slices/sdksSlice';
+import { createSettingsSlice } from './slices/settingsSlice';
+import { createSharedSlice } from './slices/sharedSlice';
+import { createSubscriptionsSlice } from './slices/subscriptionsSlice';
+import { createTracesSlice } from './slices/tracesSlice';
+import type { SentryStore, SharedSliceActions } from './types';
+
+const useSentryStore = create<SentryStore & SharedSliceActions>()((...a) => ({
```
BYK, line 12: What is `a`? Can we use something a bit more descriptive? I guess this stands for `args`?

packages/overlay/src/integrations/sentry/store/utils.ts
```diff
@@ -0,0 +1,9 @@
+export function toTimestamp(date: string | number) {
+  if (typeof date === 'string') return new Date(date).getTime();
+  return date * 1000;
+}
+
+export function relativeNsToTimestamp(startTs: number, ns: number | string) {
+  const nsStr = ns.toString();
+  return nsStr.length > 3 ? startTs + Number.parseInt(nsStr.slice(0, -3), 10) / 1000 : startTs;
+}
```
BYK, line 9: I think `isLocalTrace` also belongs here.

BYK (changes requested):
Haven't look into every slice as I'm assuming it is just extracted code from the old store code. Have a bit of a confusion regarding code organization for `getLocalTraces` and `isLocalTrace`.

Am I right to assume that with this change, we would have less rerenders as we access and mutate the store more precisely?

Other than these, looks great thanks!

Shubhdeep12:
> Am I right to assume that with this change, we would have less rerenders as we access and mutate the store more precisely?
> 


Slicing in Zustand is just cosmetic. It's just splitting the big store into multiple slices, not isolating the slices.
For less rerenders -> there is a specific way to use Zustand store by just selecting/subscribing the data needed like
`const traces = useSentryStore(state => state.traces)`
instead of `const {traces} = useSetnryStore()`

This way, the component will only render if traces are changed.

packages/overlay/src/integrations/sentry/store/index.ts
Shubhdeep12: this is just passing down all attributes to slices. 
just took a reference from docs - https://zustand.docs.pmnd.rs/guides/slices-pattern#:~:text=Combining%20all%20the%20stores%20together%20is%20the%20same%20as%20before.

Shubhdeep12: mainly set and get methods

packages/overlay/src/integrations/sentry/data/useSentryHelpers.tsx
Shubhdeep12: yeah will fix this

packages/overlay/src/integrations/sentry/components/events/EventList.tsx
Shubhdeep12: let me check if we can use getLocalTraces here

BYK (approved):

BYK:
Great work, thanks!

Issue #650, opened 2025-01-11
Highlight slow queries


BYK:
Current queries view is useful but it doesn't even sort queries by duration by default. We should highlight slow queries by sorting them in reverse for time taken and marking the ones over a threshold with a different color.

PR #791, opened 2025-04-27
Feat(overlay): Added Timebar in queries listing

Shubhdeep12:
… to totalTime

Before opening this PR:

- [x] I added a [Changeset Entry](https://spotlightjs.com/contribute/changesets/) with `pnpm changeset:add`
- [x] I referenced issues that this PR addresses

- Addressing #788 by adding a timebar to the query listing.
- Added threshold color on total time spent values - maybe fixes #650 (took a ref from Profiles Tab)

image

PR #793, opened 2025-04-28
feat: Align other Insights pages with Profiles

### LINKED ISSUE #788: Align Queries tab with Profiles tab using TimeBar, rename Time Spent
we should follow up this by aligning the Queries tab with the Profiles tab:

1. We should use the new TimeBar component there
2. Rename "Time Spent" to "Total Time"

I wonder if we can use `TimeBar` on other pages such as "Resources"

_Originally posted by @BYK in https://github.com/getsentry/spotlight/issues/786#issuecomment-2831468981_

BYK:
Fixes #788


packages/overlay/src/integrations/sentry/components/shared/TimeBar.tsx
```diff
@@ -3,19 +3,19 @@ import classNames from '~/lib/classNames';
...
+  title?: string | null;
+  children?: React.ReactNode;
   className?: string;
 };
 
-export function TimeBar({ value, maxValue, title, text, className }: TimeBarProps) {
-  const percentage = maxValue !== 0 ? Math.round(value / maxValue) * 100 : 100;
-  const tooltip = title ?? text ?? '';
+export function TimeBar({ value, maxValue, title, children, className }: TimeBarProps) {
+  const percentage = maxValue !== 0 ? Math.round((value / maxValue) * 100) : 100;
```
BYK, line 12: This was a bug I introduced into @betegon's earlier patch.

Shubhdeep12:
Hey @BYK 
Looks like this PR covers the changes done in these
https://github.com/getsentry/spotlight/pull/790
https://github.com/getsentry/spotlight/pull/791

Can you please check so that we can cancel both of these.

BYK:
Oops, sorry @Shubhdeep12 didn't see those before starting work on this. Merged #790 both to give you the credit _and_ to make the fix in a dedicated PR (the _right way_ to things 😅 )

I'd say this definitely covers most of the changes you made in #791 but it also has some more so I'd love to merge this one and then refactor #791 to add the 3 specific changes in:

1. Span coloring that also fixes #650 
2. The keyboard navigation fixes you made, thanks! 🙏🏻 
3. The entire-row clickability (vs cell click right now)

Do you think this is okay?

packages/overlay/src/integrations/sentry/components/shared/TimeBar.tsx
```diff
@@ -4,18 +4,17 @@ type TimeBarProps = {
...
+export function TimeBar({ value, maxValue, title, children, className }: TimeBarProps) {
   const percentage = maxValue !== 0 ? Math.round((value / maxValue) * 100) : 100;
-  const tooltip = title ?? text ?? '';
 
   return (
-    <div title={tooltip} className={classNames('relative text-left text-sm font-medium', className)}>
-      {text && <div className="truncate text-lime-500">{text}</div>}
-      <div className={classNames('bg-primary-800 h-1 w-full overflow-hidden rounded-full', text ? 'mt-1' : '')}>
+    <div title={title} className={classNames('relative text-left text-sm font-medium', className)}>
+      {children && <div className="truncate text-lime-500">{children}</div>}
```
Shubhdeep12, line 16: @BYK should we render children directly instead of wrapping them in a `<div>`?
This would give us more flexibility- for example, allowing just a progress bar without inserting an unnecessary empty div when no text is needed.

But nevertheless, it's good with this too!

Shubhdeep12 (approved):

packages/overlay/src/integrations/sentry/components/shared/TimeBar.tsx
BYK: Good thinking, removed it along with the default `text-lime-500` class as I think that should be reserved for Profiles for now (if at all).

PR #824, opened 2025-06-03
fix(overlay): Fix scroll and resizer overflow issues

BYK:


packages/overlay/src/components/Overview.tsx
```diff
@@ -49,16 +49,14 @@ export default function Overview({
   }, [notificationCountSum, setTriggerButtonCount]);
 
   return (
-    <div className="flex flex-1">
+    <div className="flex overflow-hidden">
```
dcramer, line 52: without flex-1 the logs panel i think ends up not being full-width (thats why i added it)

id worry about hiding overflow as usually it masks other bugs but i can take a look

BYK: Quickly checked all tabs, all seem fine. Without `overflow-hidden`, the width usually goes off-viewport.

The `flex-1` here was unnecessary and was breaking limiting the total heigth to the viewport (so not triggering the scroll).

I think I've fixed all these issues while also keeping filter sections sticky.

LMK if you bump into any issues and I'll dig again.

PR #829, opened 2025-06-05
fix: Span details view

BYK:
image


packages/overlay/src/integrations/sentry/tabs/TracesTab.tsx
```diff
@@ -109,6 +99,10 @@ export default function TracesTab() {
           path="/:traceId/spans/:spanId"
           element={<TraceSplitViewLayout aiMode={aiMode} onToggleAIMode={handleToggleAIMode} />}
         />
+        <Route
```
BYK, line 102: This was the main fix. Without this route definition, we were immediately going back to the main "trace view" URL due to the catch all below (`/:traceId/*`)

PR #831, opened 2025-06-06
feat(overlay): Trace details layout change

betegon:
## Layout and minor changes:
- Now trace title shows up at the top when you click on a trace. Because of this, now there's not Trace header on the trace details view as the info is already at the top.
- Traces that contain AI now  have an AI tag for better discoverability.

https://github.com/user-attachments/assets/9458148f-1626-4eb2-85f1-a56632620f47

### Notes
- Now there's no X button on trace details. Not sure if that's a good idea. Now I'm used to closing the trace details by clicking on the trace header but may no be intuitive for newbies.

Before opening this PR:


packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
```diff
@@ -0,0 +1,98 @@
+import { Link } from "react-router-dom";
```
betegon, line 1: I would like some feedback about this.

We had some code duplication between the trace  detail header and trace list item. They basically are the same. 

I created this component to encapsulate that but ended up having too much logic inside due to handling the 2 cases.

not sure if the best option is to create the two components (trace item, trace header) and have code duplication as both components will evolve (or I hope so 😆 )

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -1,149 +1,92 @@
...
+  onTraceSelect?: (traceId: string) => void;
+  selectedTraceId?: string;
+  traceData: {
+    filtered: Trace[];
+    all: Trace[];
+    visible: Trace[];
+  };
+  displayConfig: {
+    aiMode: boolean;
+    hideSelectedInline?: boolean;
```
betegon, line 21: Encapsulated these two topics into objects so they are easily extendable without ending up with 100 props. thought i would make sense

packages/overlay/src/integrations/sentry/components/traces/TraceDetails/index.tsx
```diff
@@ -77,10 +78,6 @@ export default function TraceDetails({ traceId, onClose, aiMode, onToggleAI }: T
   };
 
   return (
-    <div className="flex h-full flex-col">
-      <TraceDetailHeader trace={trace} onClose={onClose} hasAI={hasAI} aiMode={aiMode} onToggleAI={onToggleAI} />
-
-      {aiMode && hasAI ? renderAITraceView() : renderNormalTraceView()}
```
BYK, line 83: My only feedback (from earlier revisions too): instead of having these `renderX` functions, let's extract these parts as separat components in a follow-up.

packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
BYK: The solution is having 2 components bu then splitting them into smaller hooks and components and reuse those to avoid repetition 😉

packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+        {content}
+      </Link>
+    );
+  }
+
+  return (
+    <div className={wrapperClassName} onClick={onClick}>
+      {content}
+    </div>
+  );
```
BYK, line 97: Easier to follow as a ternary:
```ts
return asLinkg && href ? (
...
) : (
...
)
```

Actually, it should _always_ be a link. When we work with URLs and React Router, we don't need to keep track of "selectedTrace" etc as its already in the URL. It gives you simpler and more consistent code along with a more consistent user experince. Also, when we finally make actual URL-based history (not memory) in React Router, this URLs will be very useful.

See my earlir patch where I removed all this `onClick` logic and replaced them with URLs: [packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx](https://github.com/getsentry/spotlight/pull/829/files#diff-2e0eee345a8ad0d0fc846a6371bdb4be667703874fc3c75756f56108f822619f)

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -1,149 +1,92 @@
...
 
-  const [searchQuery, setSearchQuery] = useState("");
-  const [activeFilters, setActiveFilters] = useState<string[]>([]);
-
-  const { TRACE_FILTER_CONFIGS, filteredTraces } = useTraceFiltering(visibleTraces, activeFilters, searchQuery);
-
-  // Auto-scroll to selected trace
   useEffect(() => {
-    if (traceId && selectedTraceRef.current) {
+    if (selectedTraceId && selectedTraceRef.current) {
```
BYK, line 39: Yeah, you are essentially reverting my URL-based changes. Please go back to the URL-based approach and stop keeping track of `selectedTraceId`. It should come from URL params.

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -1,149 +1,92 @@
...
       selectedTraceRef.current.scrollIntoView({
         behavior: "smooth",
-        block: "start",
+        block: "nearest",
```
BYK, line 42: Why?

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -1,149 +1,92 @@
...
+  displayConfig,
+  onShowAll,
+}: TraceListProps) {
   const getTraceById = useSentryStore(state => state.getTraceById);
   const selectedTraceRef = useRef<HTMLDivElement>(null);
 
-  const [showAll, setShowAll] = useState(!context.experiments["sentry:focus-local-events"]);
-  const visibleTraces = showAll ? allTraces : localTraces;
-  const hiddenItemCount = allTraces.length - visibleTraces.length;
+  const hiddenItemCount = traceData.all.length - traceData.visible.length;
```
BYK, line 36: I'd say this information should also come from `traceData`? Like `traceData.hiddenItemCount`

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -1,149 +1,92 @@
...
-              const traceData = getTraceById(trace.trace_id);
-              const isAITrace = traceData ? hasAISpans(traceData) : false;
+          {hiddenItemCount > 0 && <HiddenItemsButton itemCount={hiddenItemCount} onClick={onShowAll} />}
+          {traceData.filtered.map(trace => {
+            const isSelected = selectedTraceId === trace.trace_id;
+            const traceData = getTraceById(trace.trace_id);
+            const isAITrace = traceData ? hasAISpans(traceData) : false;
 
-              // TODO: For this #<traceId> link to work as intended, we need to do something like this:
-              //       https://dev.to/mindactuate/scroll-to-anchor-element-with-react-router-v6-38op
```
BYK, line 115: Get my comment and URLs back please :)

packages/overlay/src/integrations/sentry/tabs/TracesTab.tsx
```diff
@@ -91,25 +230,93 @@ export default function TracesTab() {
     setAiMode(prev => !prev);
   }, []);
 
+  const context = useSpotlightContext();
+  const { allTraces, localTraces } = useSentryTraces();
+  const [showAll, setShowAll] = useState(() => !context.experiments["sentry:focus-local-events"]);
+  const visibleTraces = showAll ? allTraces : localTraces;
+  const [searchQuery, setSearchQuery] = useState("");
+  const [activeFilters, setActiveFilters] = useState<string[]>([]);
+  const { TRACE_FILTER_CONFIGS, filteredTraces } = useTraceFiltering(visibleTraces, activeFilters, searchQuery);
```
BYK, line 239: These all should probably be combined in a single hook that simply returns `traceData`. It would have `all`, `local`, `filtered`, `searchQuery`, `activeFilters` etc. Not sure about the name `hiddenTraceCount` as it is actually `nonLocalTraceCount`. We would then move the `showAll` logic to somewhere else (potentially a global option through the URL like `?showNonLocalTraces=1`)

BYK (commented):
Good one and it looks very neat thanks!

My biggest gripe is around you undoing my URL-based bookkeeping stuff. I'm guessing you had a merge conflict and opted to go with your changes rather than looking into what I changed (and why :)).

Other is us needing to split down these component. Some of them are way too large at this point and hard to follow. Add some more hooks we extract and I think we will be in a pretty good shape!

PR #832, opened 2025-06-08
docs: Add Laravel setup guide for Spotlight

HazAT:
Vibed hard, need to check preview

### DIFF
--- a/packages/website/src/content/docs/setup/laravel.mdx
+++ b/packages/website/src/content/docs/setup/laravel.mdx
@@ -1,0 +1,61 @@
+---
+title: Using Spotlight with Laravel
+description: A guide on how to setup Spotlight for Laravel
+---
+
+import { Content as InstallCommand } from "../../../components/InstallCommand.mdx";
+import { Tabs, TabItem } from "@astrojs/starlight/components";
+
+This guide gets Spotlight up and running in Laravel with Sentry SDK.
+
+## 1. Installation
+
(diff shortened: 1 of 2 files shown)


HazAT:
Why is it not in the sidebar 🤔 
The url is accessible 
https://spotlightjs-git-docs-add-laravel-setup-guide.sentry.dev/setup/laravel/

packages/website/src/content/docs/setup/laravel.mdx
```diff
@@ -0,0 +1,61 @@
...
+</Tabs>
+
+## 2. Enable Spotlight
+
+### Configure Sentry
+
+In your `config/sentry.php` file, uncomment or add the spotlight configuration:
+
+```php
+'spotlight' => env('SENTRY_SPOTLIGHT', false),
```
BYK, line 38: I think this is the default now so we should probably remove this section: https://github.com/getsentry/sentry-php/pull/1789

BYK (approved):
Love it, just one question before smashing that merge button!

packages/website/src/content/docs/setup/laravel.mdx
HazAT: Not in laravel https://github.com/getsentry/sentry-laravel/blob/7dda008a8812f7b3747499ec85eefac04de55dfa/config/sentry.php#L14
for whatever reason 🤷‍♂️

PR #835, opened 2025-06-09
ref(overlay): Restore routing for traces and spans

betegon:
I completely destroyed @BYK routing done in #831 . This should restore it.

Added a short cursor rule for routing. I haven't used it yet (I thought about it once I created the PR) but it should be better than not having one.  We could keep adding rules for different topics. 

Next I'll create a hook for traceData as per https://github.com/getsentry/spotlight/pull/831#discussion_r2132878766 to clean up  `TracesTab`

Before opening this PR:

### DIFF
--- a/.cursor/rules/.cursor/rules/routing.mdc
+++ b/.cursor/rules/.cursor/rules/routing.mdc
@@ -1,0 +1,15 @@
+---
+description: 
+globs: 
+alwaysApply: false
+---
+## Routing
+Use React Router v6 (with MemoryRouter) for all routing logic.
+
+Details:
+	•	Use <Routes> and <Route> to define all navigable views.
+	•	Use <Link> or useNavigate for navigation — do not manually manage routing state like selectedTrace or similar in local component state.
+	•	When possible, extract state from the URL using useParams or useSearchParams instead of passing props or using global state.
+	•	Treat the URL as the source of truth for navigation-related state (e.g. selected items, tabs, filters, etc).
+
+Examples of routing: [TraceItem.tsx](mdc:packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx) [index.tsx](mdc:packages/overlay/src/integrations/sentry/components/traces/TraceDetails/index.tsx) 
--- a/packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
+++ b/packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
@@ -68,6 +68,8 @@ export default function TraceItem({ trace, isSelected = false, className }: Trac
     </>
   );
 
+  // TODO: For this #<traceId> link to work as intended, we need to do something like this:
+  //       https://dev.to/mindactuate/scroll-to-anchor-element-with-react-router-v6-38op
   return (
     <Link
       className={classNames(
--- a/packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
+++ b/packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
@@ -1,4 +1,4 @@
-import { Link } from "react-router-dom";
+import { Link, useParams } from "react-router-dom";
 import { Badge } from "~/ui/badge";
 import TimeSince from "../../../../components/TimeSince";
 import classNames from "../../../../lib/classNames";
@@ -13,20 +13,12 @@ import TraceIcon from "./TraceIcon";
 type TraceItemProps = {
   trace: Trace;
   isSelected?: boolean;
-  onClick?: () => void;
   className?: string;
-  href?: string;
-  asLink?: boolean;
 };
 
-export default function TraceItem({
-  trace,
-  isSelected = false,
-  onClick,
-  className,
-  href,
-  asLink = false,
-}: TraceItemProps) {
+export default function TraceItem({ trace, isSelected = false, className }: TraceItemProps) {
+  const { spanId } = useParams<{ spanId: string }>();
+
   const duration = getFormattedSpanDuration(trace);
   const truncatedId = truncateId(trace.trace_id);
   const isLocal = isLocalTrace(trace.trace_id);
@@ -76,23 +68,16 @@ export default function TraceItem({
     </>
   );
 
-  const wrapperClassName = classNames(
-    "hover:bg-primary-900 flex cursor-pointer items-center gap-x-4 px-6 py-2",
-    isSelected && "bg-primary-800",
-    className,
-  );
-
-  if (asLink && href) {
-    return (
-      <Link className={wrapperClassName} to={href}>
-        {content}
-      </Link>
-    );
-  }
-
   return (
-    <div className={wrapperClassName} onClick={onClick}>
+    <Link
+      className={classNames(
+        "hover:bg-primary-900 flex cursor-pointer items-center gap-x-4 px-6 py-2",
+        isSelected && "bg-primary-800",
+        className,
+      )}
+      to={isSelected && !spanId ? `../#${trace.trace_id}` : `/traces/${trace.trace_id}/context`}
+    >
       {content}
-    </div>
+    </Link>
   );
 }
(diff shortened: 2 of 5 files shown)


packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -9,12 +10,11 @@ import TraceTreeview from "./TraceDetails/components/TraceTreeview";
 import TraceItem from "./TraceItem";
 
 type TraceListProps = {
-  onTraceSelect?: (traceId: string) => void;
-  selectedTraceId?: string;
   traceData: {
     filtered: Trace[];
     all: Trace[];
     visible: Trace[];
+    hiddenItemCount: number;
```
betegon, line 17: Added `hiddenItemCount` as part of `traceData`. @BYK you mentioned we could change the name to something like `nonLocalTraceCount`. I leave that for later as we have this naming for logs, spans and events so I didn't want to change it only for this.

packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
```diff
@@ -76,23 +68,16 @@ export default function TraceItem({
...
-
   return (
-    <div className={wrapperClassName} onClick={onClick}>
+    <Link
+      className={classNames(
+        "hover:bg-primary-900 flex cursor-pointer items-center gap-x-4 px-6 py-2",
+        isSelected && "bg-primary-800",
+        className,
+      )}
+      to={isSelected && !spanId ? `../#${trace.trace_id}` : `/traces/${trace.trace_id}/context`}
```
betegon, line 79: this is basically the line restoring routing

.cursor/rules/.cursor/rules/routing.mdc
```diff
@@ -0,0 +1,15 @@
...
+description: 
+globs: 
+alwaysApply: false
+---
+## Routing
+Use React Router v6 (with MemoryRouter) for all routing logic.
+
+Details:
+	•	Use <Routes> and <Route> to define all navigable views.
+	•	Use <Link> or useNavigate for navigation — do not manually manage routing state like selectedTrace or similar in local component state.
```
BYK, line 11: This one sounds eerily familiar 🤣

packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
```diff
@@ -13,20 +13,12 @@ import TraceIcon from "./TraceIcon";
...
-export default function TraceItem({
-  trace,
-  isSelected = false,
-  onClick,
-  className,
-  href,
-  asLink = false,
-}: TraceItemProps) {
+export default function TraceItem({ trace, isSelected = false, className }: TraceItemProps) {
+  const { spanId } = useParams<{ spanId: string }>();
```
BYK: You should not need `isSelected` here as you can get `traceId` from `useParams()` and check if that matches `trace.trace_id`?

packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
BYK, line 79: Please bring back [my note](https://github.com/getsentry/spotlight/pull/831/files#r2132873195) about the anchor scroll behavior todo:
```ts
// TODO: For this #<traceId> link to work as intended, we need to do something like this:
//       https://dev.to/mindactuate/scroll-to-anchor-element-with-react-router-v6-38op
```

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -23,49 +23,37 @@ type TraceListProps = {
...
+export default function TraceList({ traceData, displayConfig, onShowAll }: TraceListProps) {
+  const { traceId } = useParams<{ traceId: string }>();
   const getTraceById = useSentryStore(state => state.getTraceById);
   const selectedTraceRef = useRef<HTMLDivElement>(null);
 
-  const hiddenItemCount = traceData.all.length - traceData.visible.length;
-
   useEffect(() => {
-    if (selectedTraceId && selectedTraceRef.current) {
+    if (traceId && selectedTraceRef.current) {
```
BYK, line 32: This would probably go away if we implement the general solution for anchor scrolling I mentioned above? (we may need to add the achor like `/trace/<trace_id>#<trace_id>` which is not great but I'd say it's better than passing around `selectedTraceRef`?)

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -23,49 +23,37 @@ type TraceListProps = {
...
+            <HiddenItemsButton itemCount={traceData.hiddenItemCount} onClick={onShowAll} />
+          )}
           {traceData.filtered.map(trace => {
-            const isSelected = selectedTraceId === trace.trace_id;
+            const isSelected = traceId === trace.trace_id;
             const traceData = getTraceById(trace.trace_id);
             const isAITrace = traceData ? hasAISpans(traceData) : false;
 
+            // TODO: For this #<traceId> link to work as intended, we need to do something like this:
+            //       https://dev.to/mindactuate/scroll-to-anchor-element-with-react-router-v6-38op
```
BYK: Wrong place? 😅 
```suggestion
```

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
```diff
@@ -23,49 +23,37 @@ type TraceListProps = {
...
-                  <TraceItem trace={trace} isSelected={isSelected} onClick={() => onTraceSelect(trace.trace_id)} />
-                ) : (
-                  <TraceItem
-                    trace={trace}
-                    isSelected={isSelected}
-                    asLink={true}
-                    href={`/traces/${trace.trace_id}/context`}
-                  />
-                )}
+                <TraceItem trace={trace} isSelected={isSelected} />
```
BYK: Again, we should get rid of the entire `isSelected` thing.

packages/overlay/src/integrations/sentry/components/traces/TraceItem.tsx
betegon: I left that back in the TraceList 🤦 . Fixed.

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
betegon: exactly 😆

packages/overlay/src/integrations/sentry/components/traces/TraceList.tsx
betegon: Having a deeper look. It actually doesn't make sense to have an scroll (for this) right now. It was for placing the trace item at the top, before removing the trace detail header and making just one big header. 

**Before:** when clicking on a trace, we will scroll until the trace gets at the top and below it's tree and then more traces. Beware of that could be traces above and below. So you can scroll up or down.

**Now:** wehn clicking on a trace, we will show a trace header at the top with its tree below, and below, **the complete list of traces**. This means, that you can only scroll down past the trace tree to see the full list of traces until eventually seeing again the selected trace item.

Untitled-2025-06-10-0124



https://github.com/user-attachments/assets/187103c7-62d4-431c-a2d7-f9a2630149d4


So I'm not sure it makes sense to use scroll for this, but I find the usage a bit weird right now. Any thoughts?

BYK: Eh, I did not notice the change in the earlier patch. Being able to scroll past the selected trace but not backwards is not ideal. I'd leave this in for now and try to fix that behavior -- instead of detaching the header etc. I'll need to dig deeper for that apparently.

I also want to remove the header for span details and embed that information into the trace header to save space and for visual consistency. Will get to that.

BYK (approved):
This is good stuff, rolling forward.

PR #856, opened 2025-07-19
fix(overlay) : routes persistence

### LINKED ISSUE #811: Keep routing state when "minimizing" overlay
> each time i clicked out of spotlight i had to navigate back to the thing [...] we kind of wanna minimize if you will, bring it back to original state, vs "close" - @dcramer 

Imo this would be great. I think it has some relation with the ability to develop/debug your product while having spotlight aside, so you are using spotlight "live".

Shubhdeep12:
Fixes #811

### DIFF
--- a/packages/overlay/src/App.tsx
+++ b/packages/overlay/src/App.tsx
@@ -172,9 +172,9 @@ export default function App({
   const contextId = sidecarUrl || (typeof window !== "undefined" ? window.location.origin : "default");
 
   // helper to get valid panel routes
-  const getValidRoutes = () => {
+  const getValidRoutes = useCallback(() => {
     return getPanelsFromIntegrations(integrations, integrationData).map(panel => `/${panel.id}`);
-  };
+  }, [integrationData, integrations]);
 
   const clearEvents = useCallback(async () => {
     try {
--- a/packages/overlay/src/App.tsx
+++ b/packages/overlay/src/App.tsx
@@ -169,11 +169,11 @@ export default function App({
   const location = useLocation();
 
   // contextId for namespacing of routes in sessionStorage
-  const contextId = sidecarUrl || (typeof window !== "undefined" ? window.location.origin : "default");
+  const contextId = sidecarUrl;
 
   // helper to get valid panel routes
   const getValidRoutes = useCallback(() => {
-    return getPanelsFromIntegrations(integrations, integrationData).map(panel => `/${panel.id}`);
+    return new Set(getPanelsFromIntegrations(integrations, integrationData).map(panel => `/${panel.id}`));
   }, [integrationData, integrations]);
 
   const clearEvents = useCallback(async () => {
@@ -210,11 +210,14 @@ export default function App({
         try {
           const lastRoute = sessionStorage.getItem(getRouteStorageKey(contextId));
           const validRoutes = getValidRoutes();
-          if (lastRoute && validRoutes.includes(lastRoute) && lastRoute !== location.pathname) {
+          if (lastRoute && lastRoute !== location.pathname && validRoutes.has(lastRoute)) {
             navigate(lastRoute);
           }
-        } catch {
-          // ignore path error
+        } catch (error) {
+          log("Failed to retrieve or navigate to the last route from browser storage", {
+            error,
+            currentPath: location.pathname,
+          });
         }
       }
     },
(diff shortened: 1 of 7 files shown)


betegon:
@sentry review

betegon:
Hey @Shubhdeep12 , I want to try out sentry review here. I'll review after that!

packages/overlay/src/App.tsx
```diff
@@ -164,6 +166,16 @@ export default function App({
...
   // See https://github.com/remix-run/react-router/issues/7634
   const navigate = useNavigate();
+  const location = useLocation();
+
+  // contextId for namespacing of routes in sessionStorage
+  const contextId = sidecarUrl || (typeof window !== "undefined" ? window.location.origin : "default");
+
+  // helper to get valid panel routes
+  const getValidRoutes = useCallback(() => {
+    return getPanelsFromIntegrations(integrations, integrationData).map(panel => `/${panel.id}`);
```
BYK: Let's convert this to a `Set()` as the main operation we do is a look up:
```suggestion
    return new Set(getPanelsFromIntegrations(integrations, integrationData).map(panel => `/${panel.id}`));
```

packages/overlay/src/App.tsx
```diff
@@ -192,9 +204,21 @@ export default function App({
...
       log("Open");
       setOpen(true);
-      if (e.detail.path) navigate(e.detail.path);
+      if (e.detail.path) {
+        navigate(e.detail.path);
+      } else {
+        try {
+          const lastRoute = sessionStorage.getItem(getRouteStorageKey(contextId));
+          const validRoutes = getValidRoutes();
+          if (lastRoute && validRoutes.includes(lastRoute) && lastRoute !== location.pathname) {
```
BYK: Based on the above, we need to use `.has` instead of `.includes`. Moreover, it makes sense to sort the comparisons in terms of cost. In this case we should check for valid route as the last step as that's the most expensive check:
```suggestion
          if (lastRoute && lastRoute !== location.pathname && validRoutes.has(lastRoute)) {
```

packages/overlay/src/App.tsx
```diff
@@ -192,9 +204,21 @@ export default function App({
...
+            navigate(lastRoute);
+          }
+        } catch {
+          // ignore path error
```
BYK: We may wanna log this into the console?

packages/overlay/src/App.tsx
```diff
@@ -164,6 +166,16 @@ export default function App({
   // as our <Route>s are scattered around a bit
   // See https://github.com/remix-run/react-router/issues/7634
   const navigate = useNavigate();
+  const location = useLocation();
+
+  // contextId for namespacing of routes in sessionStorage
+  const contextId = sidecarUrl || (typeof window !== "undefined" ? window.location.origin : "default");
```
BYK: Using `window.location.origin` is not really useful as session storage is already segregated by origin. Shall we just go with `sidecarUrl`?

packages/overlay/src/components/Overview.tsx
```diff
@@ -11,30 +13,28 @@ export default function Overview({
...
   integrations: Integration[];
   integrationData: IntegrationData<unknown>;
   setTriggerButtonCount: (count: NotificationCount) => void;
   setOpen: (value: boolean) => void;
   isOnline: boolean;
   showClearEventsButton: boolean;
+  contextId?: string;
 }) {
   const [notificationCountSum, setNotificationCountSum] = useState<NotificationCount>({ count: 0, severe: false });
+  const location = useLocation();
```
BYK, line 28: This (using `useLocation()`) causes a full render on each path change. We should be mindful about this. I wonder if we can hook into some higher-level React Router callback that gets triggered on each path change?

packages/overlay/src/components/Overview.tsx
```diff
@@ -11,30 +13,28 @@ export default function Overview({
...
+    try {
+      sessionStorage.setItem(getRouteStorageKey(contextId), location.pathname);
+    } catch {
+      // ignore path error
     }
-    return [];
-  });
+  }, [location.pathname, contextId]);
+
+  const panels = getPanelsFromIntegrations(integrations, integrationData);
```
BYK, line 41: Love the refactor here

packages/overlay/src/components/Overview.tsx
Shubhdeep12: yeah makes sense, i have an idea to resolve this issue - let me implement and push it in sometime

Shubhdeep12: Solving this only in App.js will not help us much, as code in Overview.js and other Parent components uses these React Router hooks.
lets solve this separately

BYK (approved):

PR #857, opened 2025-07-22
feat(overlay): Profile Tree

Shubhdeep12:
Before opening this PR:


packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,236 @@
...
+  sampleCount: number;
+}
+export interface FlamegraphUtilOptions {
+  getColor?: (frame: EventFrame, depth: number, parent?: NanovisTreeNode) => ColorValue;
+  getLabel?: (frame: EventFrame, depth: number, parent?: NanovisTreeNode) => string;
+}
+
+// FIXME: need to work on condition here
+// need to figure out how sentry sets is_application as true after parsing data from sdk.
+function isApplicationFrame(frame: EventFrame): boolean {
```
Shubhdeep12: we need to figure this out @betegon 
I think this condition is not valid for every framework/language

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+  frameId: number;
+  sampleCount: number;
+}
+
+interface FlamegraphUtilOptions {
+  getColor?: (frame: EventFrame, depth: number, platform?: string, parent?: NanovisTreeNode) => ColorValue;
+  getLabel?: (frame: EventFrame, depth: number, parent?: NanovisTreeNode) => string;
+}
+
+export const isApplicationFrame = (frame: EventFrame, platform?: string) => {
```
Shubhdeep12: hey @BYK @betegon added checks for the platforms that we support today in spotlight to check applicate frames.

packages/overlay/src/integrations/sentry/constants/profile.ts
```diff
@@ -0,0 +1,16 @@
+export const FRAME_COLOR = {
```
BYK, line 1: No way to put these into CSS?

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+    stacks: profile.stacks,
+    platform: profile.platform,
+  };
+}
+
+// Sentry color scheme
+// ref: https://docs.sentry.io/product/explore/profiling/flame-charts-graphs/
+function getFrameColors(frame: EventFrame, platform?: string): ColorValue {
+  if (!frame.abs_path) return FRAME_COLOR.unknown;
+  if (isApplicationFrame(frame, platform)) {
```
BYK: I can easily see us adding more frame types like "framework" so my suggestion is to refactor this to following pieces:

1. A utility `getFrameType`
2. A map that maps frame types to details like label, colors etc.

Not required in this PR but would be great if it can be done.

packages/overlay/src/integrations/sentry/utils/profileTree.ts
BYK: This should definitely live in a separate helper as we can (and probably will) use it in more places like where we graft profiles into traces. This also includes the logic around getting the frame type and colors, labels etc.

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+      id: "empty",
+      text: "No profile data",
+      subtext: "",
+      sizeSelf: 0,
+      size: 0,
+      children: [],
+      color: "#6b7280",
+      frameId: -1,
+      sampleCount: 0,
+    };
```
BYK: Let's move this into a contstant, frozen object?

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+    typeof options.getLabel === "function" ? options.getLabel : (frame: EventFrame) => frame.function || "anonymous";
+  const totalDuration =
+    samples.length > 0 ? samples[samples.length - 1].start_timestamp - samples[0].start_timestamp : 0;
+
+  type TreeNodeWithMap = NanovisTreeNode & { childrenMap?: Map<number, TreeNodeWithMap> };
+
+  const root: TreeNodeWithMap = {
+    id: "root",
+    text: `${platform || "unknown"} Profile`,
+    subtext: `${totalDuration.toFixed(2)}ms total`,
```
BYK: We may wanna use `formatDuration` (should be somewhere in the code?) for this to make durations consistent -- what if it was 2.1s which would currently show as 2100ms

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+    !path.includes("lib/python");
+
+  if (!platform) {
+    return fallback;
+  }
+
+  if (platform.startsWith("javascript")) {
+    return !(
+      path.includes("node_modules") ||
+      func.includes("node::") ||
```
BYK:
```suggestion
      func.includes("node:") ||
```

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+      path.includes("webpack-internal") ||
```
BYK:
```suggestion
      path.includes("webpack-internal:") ||
```

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+  if (!profile || !Array.isArray(profile.samples) || !Array.isArray(profile.frames) || !Array.isArray(profile.stacks)) {
+    return null;
+  }
+  return {
+    samples: profile.samples,
+    frames: profile.frames,
+    stacks: profile.stacks,
+    platform: profile.platform,
+  };
+}
```
BYK: Why is this helper necessary? Because all fields are optional?

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+    color: "#fff",
+    frameId: -1,
+    sampleCount: 0,
+    childrenMap: new Map<number, TreeNodeWithMap>(),
+  };
+
+  for (const sample of samples) {
+    const stackId = sample.stack_id;
+    const stack = stacks[stackId];
+    if (!Array.isArray(stack) || stack.length === 0) continue;
```
BYK: Do we ever hit this condition? I'd fail instead of hiding this data problem

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+    let currentNode: TreeNodeWithMap = root;
+    root.sampleCount++;
+    const reversedStack = [...stack].reverse();
```
BYK: I think this thould be optional? Also, let's try [`.toReversed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) and see if we have support for that:
```suggestion
    const reversedStack = stack.toReversed();
```

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+          sizeSelf: 0,
+          size: 0,
+          children: [],
+          frame: frame,
+          frameId: frameId,
+          sampleCount: 1,
+          color: getColor(frame, depth, platform, currentNode),
+          childrenMap: new Map<number, TreeNodeWithMap>(),
+        };
+        currentNode.children.push(childNode);
```
BYK, line 132: Shouldn't this line be outside of the if block? We want to add `childNode` to the `children` array no matter what (whether the `currentNode` record existed or not)

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+    for (let depth = 0; depth < reversedStack.length; depth++) {
+      const frameId = reversedStack[depth];
+      const frame = frames[frameId];
+      if (!frame) continue;
+      if (!currentNode.childrenMap) {
+        currentNode.childrenMap = new Map<number, TreeNodeWithMap>();
+      }
+      let childNode = currentNode.childrenMap.get(frameId);
+      if (!childNode) {
+        childNode = {
```
BYK, line 119: It'd be great if you can add a type for this and explain the fields briefly. For instance I'm really confused between `sizeSelf` and `size` and why do we need to keep track of them.

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+        childNode.sampleCount++;
+      }
+      currentNode = childNode;
+    }
+    if (currentNode !== root) {
+      currentNode.sizeSelf++;
+    }
+  }
+
+  function removeChildrenMap(node: TreeNodeWithMap): void {
```
BYK, line 150: A short jsdoc explaining where and why we use this helper would be great.

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+/**
+ * Returns a display name for a Sentry profile (thread name or fallback).
+ */
+export function getProfileDisplayName(profile: SentryProfileWithTraceMeta, idx = 0): string {
+  return profile.thread_metadata?.[profile.active_thread_id]?.name || profile.platform || `Profile ${idx + 1}`;
+}
+
+function sortChildrenBySize(node: NanovisTreeNode): void {
+  if (Array.isArray(node.children)) {
+    node.children.sort((a, b) => b.size - a.size);
```
BYK: Probably makes sense to extract `(a, b) => b.size - a.size` into a constant helper at the top level to avoid creating it over and over in this recursive call.

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+    for (const child of node.children) {
+      sortChildrenBySize(child);
+    }
+  }
+}
+
+function createSentryFrameSubtext(frame: EventFrame): string {
+  const parts: string[] = [];
+  if (frame.abs_path) {
+    const filename = frame.abs_path.split("/").pop() || frame.abs_path;
```
BYK: This pattern is quite expensive when used in loops: ` frame.abs_path.split("/").pop()`. It creates _and_ mutates an array just to get the last item. A more efficient way to do the same thing would be something like:

```ts
const lastDelimIdx = frame.abs_path.lastIndexOf('/');
const filename = lastDelimIdx !== -1 : frame.abs_path.slice(lastDelimIdx) : frame.abs_path;
```

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+    const cleanFilename = filename.replace(/\?v=[a-f0-9]+$/, "");
+    parts.push(cleanFilename);
+  }
+  if (typeof frame.lineno === "number") {
+    parts.push(`L${frame.lineno}`);
+  }
+  if (typeof frame.colno === "number") {
+    parts.push(`C${frame.colno}`);
+  }
```
BYK: A typical notation for this is `filename:<LINE_NO>:<COL_NO>`

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+  return parts.length > 0 ? parts.join(" ") : "unknown location";
+}
+
+function convertSampleCountsToSizes(node: NanovisTreeNode, totalSamples: number): void {
+  const samplePercentage = (node.sampleCount / totalSamples) * 100;
+  if (Array.isArray(node.children)) {
```
BYK: Shall we just make sure `node.children` is always an array (even empty `[]`) and get rid of these checks?

packages/overlay/src/integrations/sentry/utils/profileTree.ts
```diff
@@ -0,0 +1,282 @@
...
+    const originalSubtext = node.subtext;
+    node.subtext = `${samplePercentage.toFixed(1)}% | ${originalSubtext}`;
+  }
+}
+
+/**
+ * Converts a Sentry profile to a normalized nanovis tree node for visualization.
+ * Always normalizes the tree for correct sunburst rendering.
+ */
+export async function convertSentryProfileToNormalizedTree(
```
BYK, line 174: We may wanna memoize this function.

BYK (changes requested):
Overall, great patch thanks so much!

Added a bunch of suggestions, mostly optimizations. If you think they are gonna take too much time we can merge as is and follow up right after that but I'd prefer to get them in with this PR.

packages/overlay/src/integrations/sentry/constants/profile.ts
Shubhdeep12: These are for the package used to render Profile Tree - it accepts the hex string

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: this is for any C++ Node-internal call @BYK i think this will be system frame.
I think we are good here with `node::`

Shubhdeep12: but lets use single colon - otherwise it will be narrow search

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: just a check for bad data with early return and we only need samples, frames, stacks, platform for the trees

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: yeah we can remove isArray check as it is already checked while parsing. 
but this condition will not increase unwanted sampleCount that is why its needed

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: we are checking if childNode exists or not in currentNode using childrenMap.
if exists then we just increase its sampleCount
that is the reason of pushing it in if block

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: ref: https://github.com/antfu/nanovis#:~:text=/**%0A%20%20%20*%20The%20size,calculated%20automatically.%0A%20%20%20*/

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: sure

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: just wondering how this will help? memozing and depending upon profile will not help us

BYK: It was to avoid reprocessing of the same profile but I think you are right.

packages/overlay/src/integrations/sentry/constants.ts
```diff
@@ -356,3 +358,15 @@ export const LOGS_HEADERS = [
...
+  id: "empty",
+  text: "No profile data",
+  subtext: "",
+  sizeSelf: 0,
+  size: 0,
+  children: [],
+  color: "#6b7280",
+  frameId: -1,
+  sampleCount: 0,
+};
```
BYK:
```suggestion
export const SAMPLE_EMPTY_PROFILE_FRAME: NanovisTreeNode = Object.freeze({
  id: "empty",
  text: "No profile data",
  subtext: "",
  sizeSelf: 0,
  size: 0,
  children: [],
  color: "#6b7280",
  frameId: -1,
  sampleCount: 0,
});
```

packages/overlay/src/integrations/sentry/utils/frame.ts
```diff
@@ -0,0 +1,95 @@
...
+ * @returns A normalized profile object or null if the profile is invalid.
+ */
+export const parseSentryProfile = (profile: SentryProfileWithTraceMeta) => {
+  return {
+    samples: profile.samples,
+    frames: profile.frames,
+    stacks: profile.stacks,
+    platform: profile.platform,
+  };
+};
```
BYK: I am still skeptical about the value of this helper. It just creates a new "simpler" object but what's the harm in passing around the larger, original sentry profile?

packages/overlay/src/integrations/sentry/utils/profileTree.ts
BYK: Left to a follow up?

packages/overlay/src/integrations/sentry/utils/profileTree.ts
BYK: what's the difference between `currentNode.children` and `currentNode.childrenMap`? That's the confusing part for me.

packages/overlay/src/integrations/sentry/utils/profileTree.ts
BYK: Left for a follow up?

BYK (changes requested):
Looks great! Only had to block on https://github.com/getsentry/spotlight/pull/857/files#r2245003665 (may not require a code change, I just need to understand what's going on there before merging)

packages/overlay/src/integrations/sentry/utils/frame.ts
Shubhdeep12: yeah we can use original sentry profile, its just a helper that is taking out required values.. 
Let me pass the original profile.

Shubhdeep12: we'll create something like this in future, if we need to parse or do some data setup.

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: Added. 
sorry missed this earlier.

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: Yes, let's resolve this in a follow-up. Also, it would be great if you could add your thoughts on more frame types in an issue.
we can discuss and take this up separately.

packages/overlay/src/integrations/sentry/utils/profileTree.ts
Shubhdeep12: childrenMap is added for finding if childNode of a specific frame is already present in stack or not.

using map it will just take O(1) for lookup

Shubhdeep12: that is why we are removing it in end for cleanup

BYK: Okay sorry I got it now, thanks for humoring me 😅

PR #862, opened 2025-07-31
feat: migrated the sidecar server to honojs

MathurAditya724:
- Shifted to honojs

### DIFF
--- a/packages/sidecar/src/contextlines.ts
+++ b/packages/sidecar/src/contextlines.ts
@@ -1,8 +1,8 @@
 import { readFileSync } from "node:fs";
-import type { IncomingMessage, ServerResponse } from "node:http";
 import * as os from "node:os";
 import * as path from "node:path";
 import { LEAST_UPPER_BOUND, TraceMap, originalPositionFor, sourceContentFor } from "@jridgewell/trace-mapping";
+import type { Handler } from "hono";
 
 type SourceContext = {
   pre_context?: string[];
@@ -124,67 +124,52 @@ function isValidSentryStackFrame(frame: SentryStackFrame): frame is ValidSentryS
   return !!frame.filename && !!frame.lineno && !!frame.colno;
 }
 
-export function contextLinesHandler(req: IncomingMessage, res: ServerResponse) {
-  // We're only interested in handling a PUT request
-  if (req.method !== "PUT") {
-    res.writeHead(405);
-    res.end();
-    return;
+export const contextLinesHandler: Handler = async c => {
+  const requestBody = await c.req.text();
+
+  const stacktrace = parseStackTrace(requestBody);
+
+  if (!stacktrace) {
+    return c.body(null, 500);
   }
 
-  let requestBody = "";
-  req.on("data", chunk => {
-    requestBody += chunk;
-  });
-
-  req.on("end", async () => {
-    const stacktrace = parseStackTrace(requestBody);
-
-    if (!stacktrace) {
-      res.writeHead(500);
-      res.end();
-      return;
+  for (const frame of stacktrace.frames ?? []) {
+    if (
+      !isValidSentryStackFrame(frame) ||
+      // let's ignore dependencies for now with this naive check
+      frame.filename.includes("/node_modules/")
+    ) {
+      continue;
     }
-
-    for (const frame of stacktrace.frames ?? []) {
-      if (
-        !isValidSentryStackFrame(frame) ||
-        // let's ignore dependencies for now with this naive check
-        frame.filename.includes("/node_modules/")
-      ) {
+    const { filename } = frame;
+    // Dirty check to see if this looks like a regular file path or a URL
+    if (filename.includes("://")) {
+      const generatedCode = await getGeneratedCodeFromServer(frame.filename);
+      if (!generatedCode) {
         continue;
       }
-      const { filename } = frame;
-      // Dirty check to see if this looks like a regular file path or a URL
-      if (filename.includes("://")) {
-        const generatedCode = await getGeneratedCodeFromServer(frame.filename);
-        if (!generatedCode) {
-          continue;
-        }
 
-        // Extract the inline source map from the minified code
-        const inlineSourceMapMatch = generatedCode.match(/\/\/# sourceMappingURL=data:application\/json;base64,(.*)/);
+      // Extract the inline source map from the minified code
+      const inlineSourceMapMatch = generatedCode.match(/\/\/# sourceMappingURL=data:application\/json;base64,(.*)/);
 
-        if (inlineSourceMapMatch?.[1]) {
-          const sourceMapBase64 = inlineSourceMapMatch[1];
-          const sourceMapContent = Buffer.from(sourceMapBase64, "base64").toString("utf-8");
-          applySourceContextToFrame(sourceMapContent, frame);
-        }
-      } else if (!filename.includes(":")) {
-        try {
-          const lines = readFileSync(filename, { encoding: "utf-8" }).split(/\r?\n/);
-          addContextLinesToFrame(lines, frame);
-        } catch (err) {
-          if ((err as NodeJS.ErrnoException).code !== "ENOENT") {
-            throw err;
-          }
+      if (inlineSourceMapMatch?.[1]) {
+        const sourceMapBase64 = inlineSourceMapMatch[1];
+        const sourceMapContent = Buffer.from(sourceMapBase64, "base64").toString("utf-8");
+        applySourceContextToFrame(sourceMapContent, frame);
+      }
+    } else if (!filename.includes(":")) {
+      try {
+        const lines = readFileSync(filename, { encoding: "utf-8" }).split(/\r?\n/);
+        addContextLinesToFrame(lines, frame);
+      } catch (err) {
+        if ((err as NodeJS.ErrnoException).code !== "ENOENT") {
+          throw err;
         }
       }
     }
+  }
 
-    const responseJson = JSON.stringify(stacktrace);
+  const responseJson = JSON.stringify(stacktrace);
 
-    res.writeHead(200, { "Content-Type": "application/json" });
-    res.end(responseJson);
-  });
-}
+  return c.json(responseJson);
+};
--- a/packages/sidecar/src/main.ts
+++ b/packages/sidecar/src/main.ts
@@ -1,8 +1,13 @@
-import { createWriteStream, readFileSync } from "node:fs";
-import { type IncomingMessage, type Server, type ServerResponse, createServer, get } from "node:http";
-import { extname, join, resolve } from "node:path";
+import { createWriteStream } from "node:fs";
+import { get } from "node:http";
+import { resolve } from "node:path";
 import { createGunzip, createInflate } from "node:zlib";
-import { addEventProcessor, captureException, getTraceData, startSpan } from "@sentry/node";
+import { type ServerType, serve } from "@hono/node-server";
+import { serveStatic } from "@hono/node-server/serve-static";
+import { addEventProcessor, captureException, startSpan } from "@sentry/node";
+import { type Handler, Hono } from "hono";
+import { cors } from "hono/cors";
+import { streamSSE } from "hono/streaming";
 import launchEditor from "launch-editor";
 import { CONTEXT_LINES_ENDPOINT, DEFAULT_PORT, SERVER_IDENTIFIER } from "./constants.js";
 import { contextLinesHandler } from "./contextlines.js";
@@ -33,7 +38,7 @@ type SideCarOptions = {
    */
   basePath?: string;
 
-  filesToServe?: Record<string, Buffer>;
+  filesToServe?: string[];
 
   /**
    * More verbose logging.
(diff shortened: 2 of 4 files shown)


Shubhdeep12:
Hey @MathurAditya724, I was checking the changes and found that there is an issue with the SSE stream connection

The client (EventSource) is receiving both error and open events. I might be wrong, but is it because streamSSE from Hono doesn't keep the connection alive with the client?

I searched and tried adding a ping in handleStreamRequest under streamGetRequestHandler, something like
```
while (true) {
        await stream.sleep(15000);
        await stream.write(": ping\n\n");
      }
```
this solves the issue.

packages/sidecar/src/main.ts
```diff
@@ -49,269 +54,145 @@ type SideCarOptions = {
...
             }
+
+            return _data;
           };
+
       const sub = buffer.subscribe(([payloadType, data]) => {
         logger.debug("🕊️ sending to Spotlight");
-        res.write(`event:${payloadType}${base64Indicator}\n`);
-        dataWriter(data);
+        stream.writeln(`${payloadType}${base64Indicator}`);
```
Shubhdeep12: can we use `stream.writeSSE` here? will that handles adding prefix like data:, event: as well ?

https://hono.dev/docs/helpers/streaming#streamsse:~:text=await%20stream.writeSSE(%7B

MathurAditya724: I initially did that only, but when dealing with non base64 data the `writeSSE` will not work, as it just send the `"data:"` thing a single time for a single event while we wanna send them multiple times as part of the same event.

And also this was a bug pointed out by Copilot, so just corrected it

MathurAditya724:
> I might be wrong, but is it because streamSSE from Hono doesn't keep the connection alive with the client?

You're right hono does this, I will just create a custom `streamSSE` for hono which resolves this

packages/sidecar/vite.config.ts
```diff
@@ -22,7 +22,7 @@ export default defineConfig({
     outDir: "./dist",
     sourcemap: true,
     rollupOptions: {
-      external: [...dependencies, ...builtinModules.map(x => `node:${x}`)],
+      external: [...dependencies, ...builtinModules.map(x => `node:${x}`), ...builtinModules],
```
BYK: Why do we need this addition?

packages/spotlight/bin/run.js
```diff
@@ -24,22 +21,6 @@ setContext("CLI", {
   argv: process.argv.map(arg => arg.replace(homeDir, "~")),
 });
 
-const withTracing =
-  (fn, spanArgs = {}) =>
-  (...args) =>
-    startSpan({ name: fn.name, attributes: { args }, ...spanArgs }, () => fn(...args));
-
-const readAsset = withTracing(
```
BYK: Quite sure this needs to come back for fossilized binaries?

packages/sidecar/src/main.ts
```diff
@@ -49,269 +54,145 @@ type SideCarOptions = {
...
-      res.write("\n");
-
-      const useBase64 = searchParams?.get("base64") != null;
-      const base64Indicator = useBase64 ? ";base64" : "";
+const streamGetRequestHandler = (buffer: MessageBuffer<Payload>): Handler => {
+  return function handleStreamRequest(c) {
+    // TODO: Check if we still need this
+    // // Send something in the body to trigger the `open` event
+    // // This is mostly for Firefox -- see getsentry/spotlight#376
+    // res.write("\n");
```
BYK: We need to resolve this before merging

packages/sidecar/src/streaming.ts
```diff
@@ -0,0 +1,67 @@
+import type { Context } from "hono";
```
BYK, line 1: Is this module copy/pasted from somewhere? If not why are we concerned about `bun` at all?

packages/sidecar/src/main.ts
```diff
@@ -49,269 +54,145 @@ type SideCarOptions = {
...
         logger.debug("🕊️ sending to Spotlight");
-        res.write(`event:${payloadType}${base64Indicator}\n`);
-        dataWriter(data);
+        stream.writeln(`event:${payloadType}${base64Indicator}`);
+        for (const line of dataWriter(data)) {
+          stream.writeln(line);
+        }
         // This last \n is important as every message ends with an empty line in SSE
-        res.write("\n");
+        stream.write("\n");
```
BYK: We should probably replace all these with `stream.writeSSE({})` if we are using Hono?

packages/sidecar/src/main.ts
```diff
@@ -49,269 +54,145 @@ type SideCarOptions = {
...
-        } else {
-          buffer.put([contentType, body]);
+const streamPostRequestHandler = (
+  buffer: MessageBuffer<Payload>,
+  incomingPayload?: IncomingPayloadCallback,
+): Handler<{ Bindings: HttpBindings }> => {
+  return async function handleStreamRequest(c) {
+    logger.debug("📩 Received event");
+    let stream = c.env.incoming;
+    // Check for gzip or deflate encoding and create appropriate stream
```
BYK, line 120: Does hono not support these out of the box?

BYK (changes requested):
Looks like there's still some work to do on this. I've also noticed you have removed our Sentry tracing which I'd rather keep in.

Also, looking at the current state of things, I'm not sure switching to Hono provides us much value in terms of reduced complexity. Shall we really pursue this conversion or abandon before investing more time in?

packages/sidecar/vite.config.ts
MathurAditya724: For some reason I was getting this error when building the sidecar, this occurs due to `@hono/node-server`
image
By adding `"fs"` and `"path"` as external resolves this issue. This was just added for the time being to test the server first before dealing with this

packages/sidecar/src/streaming.ts
MathurAditya724: Yes, I copy/pasted this from `@hono/mcp`. I will remove it

packages/sidecar/src/main.ts
MathurAditya724: Oh! I just looked at the code for `hono/streaming`, I must have missed the breaking lines part. Will update this.

Also thanks @Shubhdeep12, for spotting this. I got tangled up in the old code 🥲

packages/sidecar/vite.config.ts
BYK: We'd better add a comment here to document this in the code for future reference.

packages/sidecar/src/main.ts
MathurAditya724: I know it works with gzip, but I will need to check for others. Just to make sure

packages/sidecar/src/mcp/index.ts
```diff
@@ -12,28 +11,112 @@ const NO_ERRORS_CONTENT: CallToolResult = {
...
+      title: "Get Local App Errors",
+      description: `Retrieve recent application errors from Spotlight debugger to diagnose failures, crashes, and exceptions across your entire stack - frontend, backend, and edge functions.
+
+**CRITICAL: Call this tool IMMEDIATELY when:**
+- User says "error", "broken", "not working", "failing", "crash", "bug", "issue"
+- You encounter errors during testing or code execution
+- Before and after making code changes to verify no regressions
+- Investigating any unexpected behavior or performance problems
+
+**What you get:**
```
BYK, line 61: It would be ideal if we kept these changes to a separate PR. Won't block for this but thoughts for future PRs :)

packages/sidecar/src/contextlines.ts
```diff
@@ -124,67 +124,52 @@ function isValidSentryStackFrame(frame: SentryStackFrame): frame is ValidSentryS
...
 }
 
-export function contextLinesHandler(req: IncomingMessage, res: ServerResponse) {
-  // We're only interested in handling a PUT request
-  if (req.method !== "PUT") {
-    res.writeHead(405);
-    res.end();
-    return;
-  }
+export const contextLinesHandler: Handler = async c => {
```
BYK: blocker: What is `c`? Let's use more descriptive variable/parameter names. Or feel free to use desctructuring in parameters.

packages/sidecar/src/contextlines.ts
```diff
@@ -124,67 +124,52 @@ function isValidSentryStackFrame(frame: SentryStackFrame): frame is ValidSentryS
...
+  }
 
-    const responseJson = JSON.stringify(stacktrace);
+  const responseJson = JSON.stringify(stacktrace);
 
-    res.writeHead(200, { "Content-Type": "application/json" });
-    res.end(responseJson);
-  });
-}
+  return c.json(responseJson);
```
BYK: Nit: Let's just combine these 2 lines:

```suggestion
  return c.json(JSON.stringify(stacktrace));
```

packages/sidecar/src/main.ts
```diff
@@ -50,360 +52,175 @@ type SideCarOptions = {
...
-      for (const [header, value] of Object.entries(headers)) {
-        res.setHeader(header, value);
-      }
-      if (req.method === "OPTIONS") {
-        res.writeHead(204, {
-          "Cache-Control": "no-cache",
-        });
-        res.end();
-        return;
+const contextId = "some-unique-id";
```
BYK: Use `genUUIDv4` or some proper crypto function please?

packages/sidecar/src/main.ts
```diff
@@ -50,360 +52,175 @@ type SideCarOptions = {
...
+const transport = new StreamableHTTPTransport();
+const mcp = createMcpInstance();
+
+export const app = new Hono<HonoEnv>()
+  .use(contextStorage(), async (c, next) => {
+    c.header("X-Powered-By", SERVER_IDENTIFIER);
+
+    c.set("contextId", contextId);
+    c.set("transport", transport);
```
BYK: Does this apply to all requests? We probably shouldn't do this?

packages/sidecar/src/main.ts
```diff
@@ -50,360 +52,175 @@ type SideCarOptions = {
...
-            // sending binary data as utf-8 over SSE which enforces utf-8
-            // encoding. This is only for backwards compatibility
-            for (const line of data.toString("utf-8").split("\n")) {
-              // This is very important - SSE events are delimited by two newlines
-              res.write(`data:${line}\n`);
-            }
-          };
+      await next();
+    },
+    async c => {
```
BYK: blocker: Same `c` strikes again.

packages/sidecar/src/main.ts
```diff
@@ -50,360 +52,175 @@ type SideCarOptions = {
...
+      const response = await c.get("transport").handleRequest(c, await c.req.json());
+      return response;
+    },
+  )
+  .get("/health", c => c.text("OK"))
+  .use(cors(), async (_c, next) => {
+    await startSpan({ name: "enableCORS", op: "sidecar.http.middleware.cors" }, async () => await next());
+  })
```
BYK: Not sure what we are doing here? Looks like we randomly create a span named `enableCORS` but not anywhere near the actual CORS enabling code?


Aside, just wondering if this would fly if this code was actually useful:

```suggestion
  .use(cors(), (_c, next) => {
    startSpan({ name: "enableCORS", op: "sidecar.http.middleware.cors" }, () => next());
  })
```

packages/sidecar/src/messageBuffer.ts
```diff
@@ -6,7 +6,7 @@ export class MessageBuffer<T> {
   private timeout = 10;
   private readers = new Map<string, (item: T) => void>();
 
-  constructor(size = 100) {
+  constructor(size = 10) {
```
BYK: Similar to my earlier comment: let's scope changes to only Hono-related.

packages/sidecar/src/contextlines.ts
MathurAditya724: `c` stands for context and I am following the usual way of writing it as you can see here in the `hono` docs - https://hono.dev/docs/

If it's better, we can do `ctx`. It's just a convention and we don't destructure it, you can see this pattern in the sentry-mcp too - https://github.com/getsentry/sentry-mcp/blob/main/packages/mcp-cloudflare/src/server/app.ts (just for ref)

BYK: Let's go with `ctx`. Definitely more readable.

packages/sidecar/src/main.ts
```diff
@@ -50,360 +53,237 @@ type SideCarOptions = {
...
+    return ctx.text("Cleared");
+  })
+  .get("/stream", ctx => {
+    const buffer = getBuffer();
+
+    const useBase64 = ctx.req.query("base64") != null;
+    const base64Indicator = useBase64 ? ";base64" : "";
+
+    return streamSSE(ctx, async stream => {
       const sub = buffer.subscribe(([payloadType, data]) => {
```
dcramer, line 102: I think what would be interesting here, is abstraction sub/unsub, and making this an async generator

possibly a future todo, as i dont know how complicated that is either, but it could make this so it "looks" like a terminating loop that cleans itself up, vs the implicit abstractions that are kind of hard to follow with streamSSE and its child run function

BYK (approved):
Unblocking

PR #864, opened 2025-08-01
feat: MCP

MathurAditya724:


packages/sidecar/src/main.ts
```diff
@@ -267,6 +269,18 @@ function handleClearRequest(req: IncomingMessage, res: ServerResponse): void {
   }
 }
 
+function handleMcpRequest(mcp: McpServer, transport: StreamableHTTPServerTransport) {
+  let init = false;
+  return async (req: IncomingMessage, res: ServerResponse) => {
+    if (!init) {
+      await mcp.connect(transport);
```
BYK: Should we not make the `mcp.connect()` call part of the MCP Setup process below?

packages/sidecar/src/mcp.ts
```diff
@@ -0,0 +1,62 @@
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { MessageBuffer } from "./messageBuffer.js";
+import type { Payload } from "./utils.js";
+
+export function createMcpInstance(buffer: MessageBuffer<Payload>) {
+  const mcp = new McpServer({
+    name: "spotlight-mcp",
+    version: "0.0.1",
```
BYK: Let's align this with the sidecar version?

packages/sidecar/src/mcp.ts
```diff
@@ -0,0 +1,62 @@
...
+            },
+          ],
+        };
+      }
+
+      const errors = [];
+
+      for (const [_contentType, data] of envelopes) {
+        const text = data.toString("utf-8");
+        const [_, { type }, payload] = JSON.parse(`[${text.replaceAll("}\n{", "},{")}]`) as any[];
```
BYK: Please do a proper envelope parse here. Feel free to copy/paste the parsing utilities here for now until we abstract them out into a shared package.

Your approach _will_ break so I'll be blocking on this change as we already have the proper parsing functions around.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -66,6 +66,13 @@ export class MessageBuffer<T> {
     this.head = 0;
     this.readers = new Map<string, (item: T) => void>();
   }
+
+  read(): T[] {
+    return [
+      ...this.items.slice(this.head % this.size, this.size).map(item => item[1]),
+      ...this.items.slice(0, this.head % this.size).map(item => item[1]),
+    ].filter(Boolean);
+  }
```
BYK: Do we really need this as `buffer.subscribe` triggers a replay of existing messages already? It also seems rather inefficient, especially at larger buffer sizes.

BYK (changes requested):
Good start!

We'd also need some tests, even if they are cursory like just making sure we can connect to the `/mcp` endpoint. Happy to help with that once we are further down the road with this PR.

packages/sidecar/src/messageBuffer.ts
MathurAditya724: The idea is to have a smaller buffer for the AI, for optimizing the context window and also having them able to focus on the most recent issues which the user has in-countered.

I have optimized the logic which is much better, initially it was just to glue things together.

packages/sidecar/src/mcp/index.ts
```diff
@@ -1,11 +1,14 @@
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
-import { MessageBuffer } from "./messageBuffer.js";
-import type { Payload } from "./utils.js";
+import type { TextContent } from "@modelcontextprotocol/sdk/types.js";
+import packageJson from "../../package.json" with { type: "json" };
```
BYK: Cannot do this due to fossilized binaries. https://github.com/getsentry/spotlight/blob/f5b67113dbdd2802af34092d49fd0b6922cd0b71/packages/spotlight/bin/instrument.js#L6 is a way (which requires embedding that value during the build process)

packages/sidecar/src/mcp/index.ts
```diff
@@ -31,26 +34,35 @@ export function createMcpInstance(buffer: MessageBuffer<Payload>) {
...
       }
 
-      const errors = [];
-
-      for (const [_contentType, data] of envelopes) {
-        const text = data.toString("utf-8");
-        const [_, { type }, payload] = JSON.parse(`[${text.replaceAll("}\n{", "},{")}]`) as any[];
+      const errors = envelopes
+        .map(([contentType, data]) => processEnvelope({ contentType, data }))
+        .sort((a, b) => {
```
BYK, line 39: We can probably move that `event.type === "event"` check here and save some memory and CPU time. It wold also allow us to do

```ts
const content = errors.map(payload => ({...}));
```

which should also be more efficient than `someArray.push()`.

packages/sidecar/src/mcp/parsing.ts
```diff
@@ -0,0 +1,66 @@
+type RawEventContext = {
```
BYK, line 4: Might be good to add a top-level comment about where this code came from and our future plans of splitting this out.

packages/sidecar/src/main.ts
```diff
@@ -91,7 +91,7 @@ const enableCORS = (handler: RequestHandler): RequestHandler =>
         res.end();
         return;
       }
-      return handler(req, res, pathname, searchParams);
+      return (await handler)(req, res, pathname, searchParams);
```
BYK: I do not like `await`ing a handler every single time we get a request. It adds at least one tick to everything. Can we find a better way to do this? I'm happy doing `await mcp.connect()` at the start of the server and delay that once rather than doing this.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -68,10 +68,16 @@ export class MessageBuffer<T> {
   }
 
   read(): T[] {
```
BYK, line 70: This does something _very_ similar to `.stream()` method above so we should align them or make them share some code. Also it might be wise to make `start` an argument but defaulting to `this.head` for different clients being at different places?

BYK (changes requested):
Looking much better. I think we can clear out some of my questions in a meeting and see how we can get this merged ASAP.

packages/sidecar/src/mcp/index.ts
Shubhdeep12: i think instead of `event.type === "event"` we should check something like this
https://github.com/getsentry/spotlight/blob/f5b67113dbdd2802af34092d49fd0b6922cd0b71/packages/overlay/src/integrations/sentry/utils/sentry.ts#L10

because we only need error events here.


and why not check this in the errorBuffer itself?

packages/sidecar/src/main.ts
```diff
@@ -329,8 +336,16 @@ function startServer(
       "/assets/main.js": readFileSync(join(basePath, "assets/main.js")),
     };
   }
+
+  // MCP Setup
+  const transport = new StreamableHTTPServerTransport({
```
Shubhdeep12, line 335: I think we should not create an MCP server instance every time. Instead, we should move the initialization logic to handleMCPRequest. 

also in handleMCPRequest, we should initiate MCP server only in first API call
something like-
```
let mcpServerInstance: McpServer;
let mcpTransport: StreamableHTTPServerTransport;

async function handleMcpRequest(): Promise<RequestHandler> {
  if (!mcpServerInstance) {
    mcpTransport = new StreamableHTTPServerTransport({
      sessionIdGenerator: undefined,
    });
    mcpServerInstance = createMcpInstance(buffer);
    await mcpServerInstance.connect(mcpTransport);
  }
  return (req, res) => mcpTransport.handleRequest(req, res);
}
```
however i think this approach also can cause an issue with concurrent requests - if multiple requests arrive simultaneously before initialization is complete, they could all try to initialize the server at the same time.

BYK (approved):
This addresses all my blockers so unblocking. Great work!

PR #870, opened 2025-08-04
fix: corrected the turbo config

MathurAditya724:
Fix the issue of `pnpm run dev` with turborepo

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -15,7 +15,7 @@
       ]
     },
     "dev": {
-      "dependsOn": ["^build"],
+      "with": ["@spotlightjs/sidecar#dev"],
       "cache": false,
       "persistent": true
     },
--- a/turbo.json
+++ b/turbo.json
@@ -14,11 +14,8 @@
         "APPLE_API_KEY_PATH"
       ]
     },
-    "build#overlay": {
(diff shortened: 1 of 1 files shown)


turbo.json
```diff
@@ -14,11 +14,8 @@
         "APPLE_API_KEY_PATH"
       ]
     },
-    "build#overlay": {
```
dcramer, line 17: ah does this not actually do what we wanted it to do?

turbo.json
```diff
@@ -14,11 +14,8 @@
...
-      "dependsOn": ["build#sidecar"],
-      "outputs": ["dist/**", "dist-bin/**", ".next/**", "!.next/cache/**"]
-    },
     "dev": {
+      "with": ["@spotlightjs/sidecar#dev"],
```
dcramer, line 18: TIL - probably can use this in some other areas where we have turbo setup

dcramer (approved):

turbo.json
MathurAditya724: Yeah, this is doing nothing. Because even if we wanna use it the key is incorrect, it should be `"build#@spotlightjs/overlay"`

PR #879, opened 2025-08-06
feat: MCP Issue Formatting

MathurAditya724:
- Added a formatting function which converts the error object in markdown for the LLM
- Added simple tests for MCP


packages/sidecar/package.json
```diff
@@ -37,7 +37,8 @@
     "@sentry/core": "catalog:",
     "@sentry/node": "catalog:",
     "kleur": "^4.1.5",
-    "launch-editor": "^2.9.1"
+    "launch-editor": "^2.9.1",
```
dcramer, line 42: this appears unused?

MathurAditya724: This is the handler which is using this - https://github.com/getsentry/spotlight/blob/main/packages/sidecar/src/main.ts#L271-L301

PR #880, opened 2025-08-06
fix: to check all items in an envelope for errors

Shubhdeep12:
Before opening this PR:

### DIFF
--- a/packages/sidecar/src/mcp/index.ts
+++ b/packages/sidecar/src/mcp/index.ts
@@ -34,7 +34,7 @@ export function createMcpInstance(buffer: MessageBuffer<Payload>) {
         };
       }
 
-      const errors = envelopes
+      const content = envelopes
         .map(([contentType, data]) => processEnvelope({ contentType, data }))
         .sort((a, b) => {
           const a_sent_at = a.envelope[0].sent_at as string;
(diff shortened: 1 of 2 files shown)


packages/sidecar/src/mcp/index.ts
```diff
@@ -34,34 +34,35 @@ export function createMcpInstance(buffer: MessageBuffer<Payload>) {
...
+      const content = envelopes
         .map(([contentType, data]) => processEnvelope({ contentType, data }))
         .sort((a, b) => {
           const a_sent_at = a.envelope[0].sent_at as string;
           const b_sent_at = b.envelope[0].sent_at as string;
           if (a_sent_at < b_sent_at) return 1;
           if (a_sent_at > b_sent_at) return -1;
           return 0;
+        })
+        .flatMap(processed => {
```
BYK: Let's move `.flatMap()` before sort so we don't spend time sorting filtered items? :)

BYK (approved):
cc @MathurAditya724

packages/sidecar/src/mcp/index.ts
BYK: Actually, moving this later makes sorting proper so it should be part of the patch.

Shubhdeep12: looks like code was updated after this patch. fixed it again.
you can review now
thanks

packages/sidecar/src/mcp/index.ts
```diff
@@ -75,12 +79,21 @@ async function formatErrorEnvelope([contentType, data]: Payload) {
...
   const {
-    envelope: [, [[{ type }, payload]]],
+    envelope: [, items],
   } = event;
 
-  if (type === "event" && isErrorEvent(payload)) {
-    return formatEventOutput(processErrorEvent(payload));
-  }
+  const formattedErrors = items
+    .map(item => {
```
MathurAditya724: Can we just use a for loop, which pushes to an array only when we have the data. Just to remove these 2 loops

packages/sidecar/src/mcp/index.ts
```diff
@@ -42,13 +42,17 @@ export function createMcpInstance(buffer: MessageBuffer<Payload>) {
...
-          if (markdown) {
-            content.push({
-              type: "text",
-              text: markdown,
-            });
+          const formattedErrors = await formatErrorEnvelope(envelope);
+
+          if (formattedErrors?.length) {
+            for (const formattedError of formattedErrors) {
+              if (formattedError) {
```
MathurAditya724: Why do we need the `if` check again over here? We already have removed the `undefined` values in the `formatErrorEnvelope` function.

You can make this a one liner add all the data using `push` in the content array and formatting then with `map`

MathurAditya724 (commented):
Thanks for the PR! 😄 

Just left some small changes/preferences

packages/sidecar/src/mcp/index.ts
Shubhdeep12: thanks, fixed it

packages/sidecar/src/mcp/index.ts
Shubhdeep12: yes we can do that.
thanks

MathurAditya724 (approved):
LGTM! 🚀

PR #890, opened 2025-08-07
fix(docs): corrected the logic for getting the electron version

MathurAditya724:
- Updated the logic to get the latest `@spotlightjs/electron`'s assets

### DIFF
--- a/packages/website/src/components/DownloadButton.astro
+++ b/packages/website/src/components/DownloadButton.astro
@@ -2,19 +2,12 @@
 import { Fragment } from 'react';
 
 let response = await fetch('https://api.github.com/repos/getsentry/spotlight/releases');
-let releases = await response.json();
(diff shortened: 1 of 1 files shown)


packages/website/src/components/DownloadButton.astro
```diff
@@ -2,19 +2,12 @@
 import { Fragment } from 'react';
 
 let response = await fetch('https://api.github.com/repos/getsentry/spotlight/releases');
-let releases = await response.json();
+let releases = await response.json().then(data => data.filter((release:any) => release.name.startsWith('@spotlightjs/electron@')));
```
BYK:
```suggestion
const releases = (await response.json()).filter((release:any) => release.name.startsWith('@spotlightjs/electron@'));
```

PR #892, opened 2025-08-07
Spotlight 2.0

MathurAditya724:
- Removed demos
- Removed tests-e2e
- Removed the standalone astro package


dcramer:
> Just wanna check, are we going to remove the astro package as well 😱?

Yeah - largely unsuccessful. We're going to ignore the original attempt at "what if we could be the new default error page" is its too much of an uphill battle. Instead we'll focus on a really rich local application to help you debug things + all the agentic stuff.

packages/electron/playwright.config.ts
```diff
@@ -0,0 +1,23 @@
...
+export default defineConfig({
+  // Timeout per test
+  timeout: 30 * 1000,
+  // Test directory
+  testDir: "tests",
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
```
BYK, line 13: Why?

package.json
```diff
@@ -14,14 +14,13 @@
...
     "test": "turbo test --filter=./packages/*",
-    "test:demos": "turbo test --filter=./demos/*",
     "test:e2e": "turbo test:e2e",
     "yalc:publish": "yalc publish --push --sig --private",
     "changeset:add": "pnpm changeset",
     "changeset:consume": "pnpm changeset version",
     "changeset:publish": "pnpm changeset publish",
     "clean": "turbo clean",
-    "clean:deps": "rimraf ./packages/*/node_modules demos/*/node_modules node_modules",
+    "clean:deps": "rimraf ./packages/*/node_modules node_modules",
```
BYK, line 24: Please no?

turbo.json
```diff
@@ -4,7 +4,7 @@
   "tasks": {
     "build": {
       "dependsOn": ["^build"],
-      "outputs": ["dist/**", "dist-bin/**", ".next/**", "!.next/cache/**"],
+      "outputs": ["dist/**", "dist-bin/**", ".next/**", "!.next/cache/**", "out/**"],
```
BYK, line 7: Would it be possible to change this `out` to `dist` to align with other projects?

packages/electron/playwright.config.ts
MathurAditya724: Idk, copied from the old playwright config 😅. I will take a look

turbo.json
MathurAditya724: This is coming from the electron package, I will take a look

package.json
MathurAditya724: We don't have any demos now, that's why I removed it

turbo.json
BYK: No blocker, let's merge and look into this later.

packages/electron/playwright.config.ts
BYK: NVM then, I'll investigate later on during CI speed improvement work.

package.json
BYK: Ouch, I misread and thought you added `clean:deps` script anew 🫨 

Ignore :D

PR #893, opened 2025-08-10
feat(sidecar): To get only specific error event when debugging with spotlight

Shubhdeep12:
Before opening this PR:

#892

### DIFF
--- a/.changeset/honest-sides-hide.md
+++ b/.changeset/honest-sides-hide.md
@@ -1,0 +1,5 @@
+---
+"@spotlightjs/sidecar": patch
+---
+
+- Added event id support in get errors mcp server tool to just get the specific error for context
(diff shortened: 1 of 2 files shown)


packages/sidecar/src/mcp/index.ts
```diff
@@ -67,13 +67,68 @@ export function createMcpInstance(buffer: MessageBuffer<Payload>) {
...
+  );
+
   // TODO: Add tool for performance tracing
   // TODO: Add tool for profiling data
 
   return mcp;
 }
 
-async function formatErrorEnvelope([contentType, data]: Payload) {
+async function formatErrorEnvelope([contentType, data]: Payload, eventId?: string) {
```
BYK: Not a fan of overloading the function like this. Shall we at least make the argument name more generic like `filter`? This way we can pass in more things like timestamps etc.

BYK (commented):
Leaving it to @MathurAditya724 for final approval and merge.

Issue #894, opened 2025-08-11
Issue in Buffer clear/reset logic


Shubhdeep12:
When the `/clear` route in Sidecar is hit -> `buffer.clear()` is executed
this also removes all subscribers, so MCP server tools will no longer receive events after that.

Shubhdeep12:
i think it would be great to have soft and hard reset in buffer.
Adding a PR, please feel free to review or update it.
thanks

PR #895, opened 2025-08-11
fix(sidecar): To Add soft reset in message buffer

Shubhdeep12:
Before opening this PR:

#894


packages/sidecar/src/messageBuffer.ts
```diff
@@ -60,13 +60,26 @@ export class MessageBuffer<T> {
...
   }
 
+  /**
+   * hard reset: drops items and subscribers and resets cursors.
+   */
   clear(): void {
     this.items = new Array(this.size);
     this.writePos = 0;
     this.head = 0;
     this.readers = new Map<string, (item: T) => void>();
```
BYK: Thinking on this, I think we should also `.clear()` the map to remove any dangling references?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -60,13 +60,26 @@ export class MessageBuffer<T> {
...
   }
 
+  /**
+   * soft reset: clears buffered items but preserves subscribers
+   * and write cursor. Readers will fast-forward to the new head and continue
+   * streaming from the next write.
+   */
+  reset(): void {
+    this.items = new Array(this.size);
```
BYK, line 79: I'd say move the code in `clear()` here (including the `writePos` and `head` reset), them make `clear()` call `this.reset()` and remove the subscribers.

BYK (changes requested):

packages/sidecar/src/messageBuffer.ts
Shubhdeep12: yeah, instead of initialising a new map for readers, we can just clear the original map like you did in reset :)

Shubhdeep12: this is already fixed, right?

PR #897, opened 2025-08-13
feat: added logs tool in the MCP

MathurAditya724:
- Added new tool in the MCP for getting the logging data


dcramer (approved):
One thiung we should add to the prompt descriptions:

"Get local logs captured with the Sentry SDK in the Spotlight server."

Same thing for errors, but lets do this in a follow up when we refine prompts

Basically make it so it really knows whats going on to help it coerce for other actions.

PR #901, opened 2025-08-14
feat(sidecar): improve debug logging with detailed event information

dcramer:
## Summary
- Implemented EventContainer pattern for lazy parsing of Sentry envelopes
- Enhanced debug logging to show detailed event metadata (SDK, error types, messages)
- Added client identification to track which clients receive events

## Key improvements
- **Zero production impact**: Envelopes are only parsed when debug mode is enabled
- **Lazy parsing**: EventContainer caches parsed results to avoid redundant processing
- **Atomic logging**: All related information is logged in single statements for consistency
- **Client identification**: Tracks clients in format `spotlight-overlay/3.2.0 (Chrome; hostname)`
- **Clean output**: Professional logging format with unicode arrows (→ incoming, ← outgoing)

## Example debug output
```
→ event | sdk: @sentry/browser | id: abc123
  └ error [javascript/production] TypeError: Cannot read property 'foo' of undefined
← event to spotlight-overlay/3.2.0 (Chrome; localhost:3000)
```

## Test plan
- [x] Verified debug logging shows detailed event information
- [x] Confirmed no parsing occurs in production mode (debug=false)
- [x] Tested client identification from overlay
- [x] Verified atomic logging (single log statements)
- [x] Linting and formatting checks pass

### DIFF
--- a/packages/overlay/src/sidecar.ts
+++ b/packages/overlay/src/sidecar.ts
@@ -1,8 +1,5 @@
 import { log } from "./lib/logger";
 
-// Version should match package.json
-const OVERLAY_VERSION = "3.2.0";
-
 /**
  * Extracts browser name from User-Agent
  */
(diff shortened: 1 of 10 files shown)


packages/overlay/src/sidecar.ts
```diff
@@ -1,5 +1,27 @@
 import { log } from "./lib/logger";
 
+// Version should match package.json
+const OVERLAY_VERSION = "3.2.0";
```
dcramer: clearly this is dumb 😂

dcramer:
Prob some cleanup we'll figure out later - this is a first pass that will move us closer to just parsing everything in sidecar => sending to overlay/mcp clients (vs double parsing). For now the parsed data is only in debug mode and used for logging.

PR #902, opened 2025-08-14
feat(sidecar): add trace viewing tools to MCP server

dcramer:
## Summary
- Adds `get_local_traces` tool to list recent traces with summaries
- Adds `get_events_for_trace` tool to show detailed span trees for specific traces
- Implements hierarchical span tree rendering with proper Unicode characters

## Test plan
- [x] Added comprehensive unit tests with inline snapshots for visual verification
- [x] Tests cover trace extraction, span tree building, orphan handling, and rendering
- [x] All tests passing (12 tests total)

## Details
This PR implements trace viewing functionality in Spotlight's MCP server, modeled after the sentry-mcp implementation but adapted to Spotlight's architecture.

### Key Features:
- **Trace extraction**: Parses Sentry transaction envelopes to extract trace data
- **Hierarchical rendering**: Builds parent-child span trees with proper indentation
- **Orphan handling**: Creates synthetic parent nodes for orphaned spans (matching UI behavior)
- **Unicode tree display**: Uses box-drawing characters (├─, └─, │) for clear visualization

### Usage:
1. First, use `get_local_traces` to see available traces:
   ```
   get_local_traces(timeWindow: 300)
   ```
   Returns summaries like:
   ```
   **71a8c5e4** | /api/users | 23ms | 2 spans | 0 errors | 2025-01-06T22:26:40.100Z
   ```

2. Then use `get_events_for_trace` with a trace ID to see the full span tree:
   ```
   get_events_for_trace(traceId: "71a8c5e4")
   ```
   Returns hierarchical tree:
   ```
   missing or unknown parent span [ce75b8fe · orphan · unknown]
      └─ /api/users [99b1b00f · 18ms]
         ├─ GET /external-api/profile [def01234 · http.client · 20ms]
         └─ SELECT * FROM users WHERE id = ? [abc12345 · db.query · 10ms]
   ```

### Implementation Notes:
- Follows the pattern established by the existing error and log tools
- Reuses the envelope parsing infrastructure
- Includes documentation in README for the trace viewing workflow
- Comprehensive test coverage with inline snapshots for tree rendering verification

### DIFF
--- a/packages/sidecar/src/mcp/utils/traces.ts
+++ b/packages/sidecar/src/mcp/utils/traces.ts
@@ -302,6 +302,14 @@ export function buildSpanTree(trace: TraceSummary): SpanNode[] {
     }
   }
 
+  // Helper function to recursively update levels
+  function updateLevels(node: SpanNode, newLevel: number) {
+    node.level = newLevel;
+    for (const child of node.children) {
+      updateLevels(child, newLevel + 1);
+    }
+  }
+
   // Create orphan parent spans for grouped orphans (like the UI does)
   for (const [parentId, orphans] of orphanedByParentId) {
     const orphanParent: SpanNode = {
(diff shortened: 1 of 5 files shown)


packages/sidecar/src/mcp/utils/traces.ts
```diff
@@ -0,0 +1,406 @@
+import type { EventContainer } from "../../eventContainer.js";
+import { processEnvelope } from "../parsing.js";
+
+export interface TraceContext {
```
dcramer, line 4: would be nice if sentry maintained types for these that we could just use

BYK: I think this is the one: https://github.com/getsentry/sentry-javascript/blob/5ee25973e6baf534ac7502b802f658bb469bbb85/packages/core/src/types-hoist/transaction.ts#L4

(from `@sentry/core` package)

Issue #891, opened 2025-08-07
Spotlight 2.0 Plan


dcramer:
We're going to make some big shifts with Spotlight, so I'm collecting this in a master ticket so folks can keep track of what is going on.

1. We're moving away from the embedded Overlay. The Electron app will be the way you consume the UI.
2. Sidecar will need a faster iteration model and will remain deployable outside of the Electron app.
3. Agentic flows will become a near term priority, including the new MCP server and TBD tools to make it easier to debug MCP and Agents.

The MCP is our immediate term focus as theres clear ROI for it existing today. We expose it to local agents (like Claude Code or Cursor) and those agents easily can query debug telemetry while working. This promotes two clear flows:

1. Human hits an issue, wants the agent to help resolve it. No more copy paste.
2. Agent hits an issue, cant get telemetry on its own (think client-side app), or its simply too complex to get it.

This is high value and we have 99% of the tech, so the main shift is building this MCP service into the sidecar to enable this (thus the need for iteration time).

That's also why we're shifting the focus to the desktop app, as Spotlight will begin being system-wide, and there's not a lot of value in us supporting N ways to deploy it.

dcramer:
Tech demo of MCP is already available at current release at http://localhost:8969/mcp

We know we need to add logs to it, improve the tool call definitions so they are picked up more implicitly, and we'll also likely need some behavior around app selection (TBD - but imagine you have multiple apps running locally and Spotlight has data coming from all of them at once).

PR #914, opened 2025-09-03
docs: Remove framework setups and integrations

BYK:
Ref #891


BYK:
>I suggest we use [fumadocs](https://fumadocs.dev/) , I'll happy port the content we can reuse, but we'll need to write new docs for the new setup.

I'm happy with the current design/system with Astro but would not put my foot down for it 😅

PR #917, opened 2025-09-05
Remove integrations and overlay mode

betegon:
Didn't want to make this PR too big, so here I'm just:

**1. removing integrations.** The idea is to have some sort of a primary navigation formed by: telemetry (sentry stuff), AI agents / MCP. 
**2. removing overlay mode.** Now the only mode is a full page application. 

notes:
- I've tried to keep as much code as possible, most of the changes are just moving files from one directory to another.
- Main files to review are `app.tsx` and `index.tsx` inside `packages/overlay/src/telemetry/`. 

Next steps:
- use url based routing.
- change design and add the top navigation mentioned above.
- rename the Overlay package to UI.


BYK:
@Shubhdeep12 standalone stays for now as it also powers electron. We'll see how we evolve from here.

packages/overlay/src/lib/useSpotlightContext.tsx
```diff
@@ -1,11 +1,9 @@
 import type React from "react";
 import { type ReactNode, createContext, useContext } from "react";
+import type { SpotlightContext } from "~/types";
 import { DEFAULT_EXPERIMENTS, DEFAULT_SIDECAR_URL } from "../constants";
-import type { SpotlightContext } from "../integrations/integration";
 
 const Context = createContext<SpotlightContext>({
-  open: () => {},
-  close: () => {},
   experiments: DEFAULT_EXPERIMENTS,
```
BYK, line 7: We should kill the experiements too

packages/overlay/src/telemetry/components/events/EventBreadcrumbs.tsx
```diff
@@ -1,7 +1,7 @@
 import { Fragment } from "react";
-import JsonViewer from "~/components/JsonViewer";
 import { cn } from "~/lib/cn";
-import Time from "../../../../components/Time";
+import JsonViewer from "~/telemetry/components/shared/JsonViewer";
```
BYK, line 3: Don't think these shared components should live under `telemetry` they are quite generic.

packages/overlay/src/telemetry/components/TelemetryTabs.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export default function TelemetryTabs({ tabs, nested = false }: TabsProps) {
+  return (
+    <nav className="flex border-b border-primary-700">
+      {tabs.map(tab => {
+        const tabPath = nested ? `./${tab.id}` : `/${tab.id}`;
+
+        return (
+          <NavLink
+            key={tab.id}
```
BYK, line 17: Pretty sure we can get rid of this dynamic tabs approach at this point? (in a follow up)

packages/overlay/src/telemetry/components/Debugger.tsx
```diff
@@ -0,0 +1,17 @@
...
+  showClearEventsButton,
+  contextId,
+}: {
+  isOnline: boolean;
+  showClearEventsButton: boolean;
+  contextId: string;
+}) {
+  return (
+    <div className="spotlight-debugger from-primary-900 to-primary-950 flex h-full flex-col overflow-hidden bg-gradient-to-br from-0% to-20% font-sans text-white">
+      <TelemetryView isOnline={isOnline} showClearEventsButton={showClearEventsButton} contextId={contextId} />
```
BYK: This higher level `Debugger` component doesn't provide any value. We should fold it into the `TelemetryView` component and reduce a layer of abstraction.

packages/overlay/src/telemetry/store/slices/sharedSlice.ts
```diff
@@ -1,8 +1,8 @@
 import type { StateCreator } from "zustand";
-import type { SentryStore, SharedSliceActions } from "~/integrations/sentry/store/types";
-import type { SentryErrorEvent } from "~/integrations/sentry/types";
-import { getNativeFetchImplementation } from "~/integrations/sentry/utils/fetch";
 import { log } from "~/lib/logger";
+import type { SentryStore, SharedSliceActions } from "~/telemetry/store/types";
+import type { SentryErrorEvent } from "~/telemetry/types";
+import { getNativeFetchImplementation } from "~/telemetry/utils/fetch";
```
BYK, line 5: This and the "local event transport" system should also be yanked as they make no sense in a world where Spotlight is not embedded.

packages/overlay/src/telemetry/utils/traces.ts
```diff
@@ -1,4 +1,4 @@
-import { log } from "../../../lib/logger";
+import { log } from "~/lib/logger";
```
BYK, line 1: Pretty sure we should look into switching to a proper logging library or just use Sentry and Spotlight itself to debug spotlight.

packages/overlay/src/index.tsx
```diff
@@ -73,157 +31,118 @@ export async function sendEvent(contentType: string, data: string | Uint8Array)
...
     activateLogger();
   }
 
+  const sidecarBaseUrl = removeURLSuffix(sidecarUrl, "/stream");
   const finalExperiments = { ...DEFAULT_EXPERIMENTS, ...experiments };
 
-  // Sentry is enabled by default
-  const defaultIntegrations = () => [sentry({ injectIntoSDK: !isLoadedFromSidecar && !fullPage })];
+  const telemetry = initTelemetry({
+    injectIntoSDK: true,
```
BYK, line 85: This option and related code makes no sense when Spotlight is not embedded. Should be yanked.

packages/overlay/src/index.tsx
```diff
@@ -73,157 +31,118 @@ export async function sendEvent(contentType: string, data: string | Uint8Array)
...
+  styleElement.textContent = `${fontStyles}\n${globalStyles}`;
+  document.head.appendChild(styleElement);
 
-  const initialTab = startFrom || (tabs.length ? `/${tabs[0].id}` : "/no-tabs");
+  const initialTab = startFrom || "/traces";
   log("Starting from", initialTab);
 
-  // TODO: Shall we enable this for other cases too such as when we use it through Django SDK?
-  if (fullPage && (isLoadedFromSidecar || document.location.hash.startsWith("#spotlight"))) {
+  const isLoadedFromSidecar = new URL(sidecarUrl).origin === document.location.origin;
```
BYK, line 110: This probably doesn't make any sense as "not loaded from sidecar" was the implication of being embedded. (yes technically you can load the UI from somewhere else but it was not the intention behind these checks -- this was mostly for instrumenting Spotlight itself without clobbering other people's sentry telemetry)

packages/spotlight/src/vite-plugin.ts
```diff
@@ -29,35 +29,26 @@ export type SpotlightIntegration = "sentry" | "viteInspect";
...
-  const integrations = (options.integrationNames || ["sentry"]).map(i => `Spotlight.${i}(${integrationOptions})`);
-  initOptions = `{integrations: [${integrations.join(", ")}], ${initOptions.slice(1)}`;
+  const initOptions = JSON.stringify(clientOptions);
 
   return [
     `import * as Spotlight from ${JSON.stringify(`/@fs${options.importPath || getSpotlightClientModulePath()}`)};`,
     `Spotlight.init(${initOptions});`,
-    "window.createErrorOverlay=function createErrorOverlay(err) { Spotlight.openSpotlight(); };",
+    "window.createErrorOverlay=function createErrorOverlay(err) { console.error('Vite build error:', err); };",
   ].join("\n");
```
BYK, line 52: This injection stuff makes no sense without overlay. Should all be yanked. The only thing that's useful in this plugin is our ability to capture and relay Vite build-time errors to Spotlight.

packages/overlay/src/lib/useSpotlightContext.tsx
betegon: do you mean showing local events by default in the UI then?

packages/overlay/src/telemetry/components/events/EventBreadcrumbs.tsx
betegon: I default to putting them to the highest level where they were used, with the idea of moving them up as we needed, but for now only used them for telemetry data.

will move them though

packages/overlay/src/lib/useSpotlightContext.tsx
betegon: forget my comment, i got it.

.changeset/long-birds-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spotlightjs/overlay": major
+---
+
+Remove integrations architecture in favour of fixed telemetry features
```
BYK, line 5: And remove the overlay too! 😄

Shubhdeep12:
> @Shubhdeep12 
> 
> 1. we're merging this and i'll open another PR with more fixes. the idea is to publish a release mostly focused on the MCP stuff @MathurAditya724 did.
> 
> 2. We were going to explore the url routing. do you want to take a look? not sure the implications with electron though. If we could, we will prefer url routing.
> 
> 
Sure, I'll check because I read sometime back  that Electron renderrer doesn't behave well with url based routing of react router dom, maybe that is fixed..

.changeset/long-birds-brake.md
betegon: Yes! I keep thinking that overlay is the packager, not an "overlay" feature 😆 .

I'll add it in a PR

PR #918, opened 2025-09-09
ref(sidecar): restructured the sidecar server with improvements

MathurAditya724:
- [x] Incapsulated handlers/routes logic and tests in separate folders, instead of a one big file.
- [x] Removed `launch-editor.d.ts`, as it was not required and removed types from the `launchEditor` callback function on the `/open` route
- [x] Resolved the TODO for adding `"server.port"` in the traces
- [x] Updated `tsconfig.json` to include nested files
- [x] Added `vitest.config.ts` to resolve alias
- [x] Shifted all the types in a separate folder

With this the base router logic stays simpler to understand and also make it easier to add more routes and tests for them.


.changeset/ripe-rivers-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spotlightjs/sidecar": minor
```
BYK:
```suggestion
"@spotlightjs/sidecar": patch
```

This would be a patch version as it doesn't introduce anything new?

packages/sidecar/src/handlers/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./fileToServe.js";
```
BYK, line 1: What's the benefit of these kinds of import/export only files?

packages/sidecar/src/routes/contextlines/utils.ts
```diff
@@ -22,7 +20,7 @@ type SentryStackTrace = {
   frames?: SentryStackFrame[];
 };
 
-async function getGeneratedCodeFromServer(filename: string): Promise<string | undefined> {
+export async function getGeneratedCodeFromServer(filename: string): Promise<string | undefined> {
```
BYK, line 23: Don't really see the benefit of splitting this file as nothing else would be using these "utility functions". Why not keep them in the contextlines handler?

packages/sidecar/src/types/utils.ts
```diff
@@ -0,0 +1,48 @@
+export type IncomingPayloadCallback = (body: string) => void;
```
BYK, line 1: Okay what is "types/utils"? It this utils for types or types for utilities? If the latter, why do we have a separate file for utility types? What's so special about them?

PR #920, opened 2025-09-10
feat: shifted the parsing logic in the sidecar

MathurAditya724:
- [x] Shifted all the parsing logic from the overlay to sidecar
- [x] Created a new `/parser` subpath export for using the parsing functions
- [x] Now, we are sending only parsed JSON data to the frontend
- [x] Integrated this with overlay
- [x] Shift the formatting logic in a separate subpath export `/format`

### DIFF
--- a/packages/sidecar/src/routes/stream/index.ts
+++ b/packages/sidecar/src/routes/stream/index.ts
@@ -65,7 +65,7 @@ const router = new Hono<HonoEnv>()
     return ctx.body(container[0].getData(), 200, {
       "Content-Type": container[0].getContentType(),
       "Cache-Control": "no-cache",
-      "Content-Disposition": `attachment; filename="${envelopeId}.txt"`,
+      "Content-Disposition": `attachment; filename="${envelopeId}.bin"`,
       Connection: "keep-alive",
     });
   })


packages/overlay/src/telemetry/components/insights/envelopes/EnvelopeDetails.tsx
```diff
@@ -2,25 +2,18 @@ import type { Envelope } from "@sentry/core";
...
-  const rawEnvelopeData = {
-    ...rawEnvelope,
-    data: typeof rawEnvelope.data === "string" ? rawEnvelope.data : parseStringFromBuffer(rawEnvelope.data),
-  };
-
   const envelopeId: string | unknown = header.__spotlight_envelope_id;
-  const downloadUrl = URL.createObjectURL(new Blob([rawEnvelope.data as BlobPart], { type: rawEnvelope.contentType }));
-  const downloadName = `${envelopeId}-${rawEnvelope.contentType}.bin`;
+  const downloadUrl = URL.createObjectURL(new Blob([rawEnvelope], { type: "application/json" }));
+  const downloadName = `${envelopeId}-application/json.json`;
```
BYK: Errr no? The envelopes are certainly _not_ JSON blobs or valid JSON files.

packages/overlay/src/telemetry/components/insights/envelopes/EnvelopeDetails.tsx
```diff
@@ -2,25 +2,18 @@ import type { Envelope } from "@sentry/core";
 import { useState } from "react";
 import { ReactComponent as Download } from "~/assets/download.svg";
 import JsonViewer from "~/telemetry/components/shared/JsonViewer";
-import { parseStringFromBuffer } from "~/telemetry/utils/bufferParsers";
-import type { RawEventContext } from "~/types";
 import SidePanel, { SidePanelHeader } from "~/ui/sidePanel";
 
-export default function EnvelopeDetails({ data }: { data: { envelope: Envelope; rawEnvelope: RawEventContext } }) {
+export default function EnvelopeDetails({ data }: { data: { envelope: Envelope; rawEnvelope: string } }) {
```
BYK: If we are not doing the parsing in the frontend/UI I don't think shipping the raw envelope string makes sense. Let's just have an endpoint on the sidecar to get the raw envelope for this and change the logic below accordingly.

packages/overlay/src/App.tsx
```diff
@@ -1,40 +1,35 @@
...
 import { processEnvelope } from "./telemetry";
 import TelemetryView from "./telemetry/components/TelemetryView";
 import sentryStore from "./telemetry/store";
-import type { RawEventContext } from "./types";
 
 type AppProps = {
   sidecarUrl: string;
   showClearEventsButton?: boolean;
-  initialEvents?: Record<string, (string | Uint8Array)[]>;
+  initialEvents?: Record<string, string>;
```
BYK: This should be dead with the removal of the overlay? And even if it isn't since you changed the envelope format that the UI expects, this breaks the API contract and will cause issues for things that actually used `initialEvents`.

packages/sidecar/src/format/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from "./errors.js";
 export * from "./logs.js";
+export * from "./traces.js";
```
BYK, line 3: Repeating a comment from an earlier PR: is there an advantage for these "only export" index files or can we remove them. These make tree-shaking more tricky.

packages/sidecar/src/utils/eventContainer.ts
```diff
@@ -62,7 +53,7 @@ export class EventContainer {
     const envelope = this.getParsedEnvelope();
     if (!envelope) return null;
 
-    return envelope.items.map(item => item.header.type).filter(Boolean) as string[];
+    return envelope.event[1].map(item => item[0].type).filter(Boolean) as string[];
```
BYK: This doesn't feel right. An envelope can have multiple events/items and we should iterate over all of them.

If you are confused about the envelope format, here's the spec page: https://develop.sentry.dev/sdk/data-model/envelopes/#serialization-format

packages/overlay/src/App.tsx
MathurAditya724: Yup, so removing this

packages/sidecar/src/format/index.ts
MathurAditya724: This purely for importing directly from the subpaths like `"@spotlightjs/sidecar/format"` and I do understand that barrel files can be tricky to optimize but 3 things -

1. It makes it easier for other to use it and these functions are somewhat intern dependent, so even if we do tree shaking we won't be saving much.
2. Once the sidecar is compiled, it's a single file with small chunks for subpath imports which is good enough IMO
3. Also, we now have 2 new subpaths - `"/parser"` and `"/format"`

This is just my opinion, happy to change it

BYK: Alright, I'm fine with it for now. Given your first reason tho, we may just combine these into a single file if this kind of thing happens again.

packages/overlay/src/telemetry/components/insights/envelopes/EnvelopeDetails.tsx
MathurAditya724: I changed this to not pass the `rawEnvelope`, and added a endpoint for getting the buffer

packages/overlay/src/telemetry/components/insights/envelopes/EnvelopeDetails.tsx
MathurAditya724: I have created a endpoint to get the rawBuffer, but then should I create a blob or let them download that directly?

packages/overlay/src/telemetry/components/insights/envelopes/EnvelopeDetails.tsx
```diff
@@ -1,19 +1,19 @@
...
 
-  const header = envelope[0];
-  const items = envelope[1];
+  const [header, items] = envelope;
 
   const envelopeId: string | unknown = header.__spotlight_envelope_id;
-  const downloadUrl = URL.createObjectURL(new Blob([rawEnvelope], { type: "application/json" }));
-  const downloadName = `${envelopeId}-application/json.json`;
+
+  const downloadUrl = getSidecarUrl(`/stream/${envelopeId}`);
```
BYK:
```suggestion
  const downloadUrl = getSidecarUrl(`/envelope/${envelopeId}`);
```

packages/sidecar/src/routes/stream/index.ts
```diff
@@ -52,6 +52,22 @@ const router = new Hono<HonoEnv>()
       });
     });
   })
+  .get("/stream/:id", ctx => {
```
BYK:
```suggestion
  .get("/envelope/:id", ctx => {
```

packages/sidecar/src/routes/stream/index.ts
```diff
@@ -52,6 +52,22 @@ const router = new Hono<HonoEnv>()
...
+    const buffer = getBuffer();
+
+    const envelopeId = ctx.req.param("id");
+    const container = buffer.read({ envelopeId });
+
+    if (container.length === 0) {
+      return ctx.notFound();
+    }
+
+    return ctx.body(container[0].getData(), 200, {
```
BYK, line 65: Need a [`Content-Disposition`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#as_a_response_header_for_the_main_body) header here for a proper download experience.

packages/sidecar/src/utils/extras.ts
```diff
@@ -60,3 +60,13 @@ export const isSidecarRunning = withTracing(
     }),
   { name: "isSidecarRunning", op: "sidecar.server.collideCheck" },
 );
+
+export function generateUuidv4(): string {
```
BYK: Since we are on Node side now we can probably use something like this: https://stackoverflow.com/a/40191779/90297

Or maybe even https://www.npmjs.com/package/uuidv7

packages/sidecar/src/logger.ts
```diff
@@ -1,10 +1,12 @@
 import type { SidecarLogger } from "./types/index.js";
 
+const SPOTLIGHT_PREFIX = "🔎 [Spotlight]";
```
BYK, line 3: 🤯

packages/sidecar/src/utils/extras.ts
MathurAditya724: Added uuidv7 package

packages/sidecar/src/logger.ts
MathurAditya724: You are looking at a 10x engineer job

packages/sidecar/src/routes/stream/index.ts
```diff
@@ -65,6 +65,7 @@ const router = new Hono<HonoEnv>()
     return ctx.body(container[0].getData(), 200, {
       "Content-Type": container[0].getContentType(),
       "Cache-Control": "no-cache",
+      "Content-Disposition": `attachment; filename="${envelopeId}.txt"`,
```
BYK:
```suggestion
      "Content-Disposition": `attachment; filename="${envelopeId}.bin"`,
```

PR #921, opened 2025-09-11
feat(overlay): Updated Routing in overlay and removed sentry-integration

Shubhdeep12:
- Simplified the App component by utilizing React Router for navigation.
- Adjusted links and navigation to reflect the new routing structure.

This refactor aims to improve code maintainability and clarity in the telemetry section of the overlay.

Before opening this PR:

#891 

- Updated Router in Overlay - BrowserRouter for url-based routing
- prefixed telemetry routes with /telemetry (default)
- removed sentry-integration -> this was useful in case of embedding. But not useful in the case of electron and standalone.
    - In case of Electron - this was tracking the telemetry data from Sentry Electron SDK used in the Spotlight Electron App.

### DIFF
--- a/packages/overlay/src/index.tsx
+++ b/packages/overlay/src/index.tsx
@@ -68,11 +68,7 @@ export async function init(initOptions: SpotlightInitOptions = {}) {
   const styleElement = document.createElement("style");
   styleElement.textContent = `${fontStyles}\n${globalStyles}`;
   document.head.appendChild(styleElement);
-
-  const isLoadedFromSidecar = new URL(sidecarUrl).origin === document.location.origin;
-  if (isLoadedFromSidecar || document.location.hash.startsWith("#spotlight")) {
-    initSentry();
-  }
+  initSentry();
 
   const context = {
     sidecarUrl: sidecarBaseUrl,


packages/overlay/src/telemetry/components/TelemetrySidebar.tsx
```diff
@@ -59,13 +59,15 @@ export default function TelemetrySidebar({
   };
 
   const isActive = (path: string) => {
-    if (path === "/traces") {
-      return pathname === "/" || pathname.startsWith("/traces");
+    // case for primary path
+    if (path === "traces") {
+      return pathname === "/telemetry" || pathname.endsWith("/telemetry") || pathname.startsWith("/telemetry/traces");
```
BYK: This part doesn't make sense to me: `pathname === "/telemetry" || pathname.endsWith("/telemetry")`
should it be `pathname.startsWith(`/telemetry`) instead?

Shubhdeep12: We can remove `pathname.endsWith("/telemetry")`, but `pathname === "/telemetry" || pathname.startsWith("/telemetry/traces")` is necessary.

This is because we open the traces tab when the route is only `/telemetry`.

packages/overlay/src/telemetry/sentry-integration.ts
Shubhdeep12, line 1: @BYK @betegon i hope we are good here?
i think now we don't need this anymore?

packages/overlay/src/telemetry/index.ts
```diff
@@ -1,251 +0,0 @@
...
-        off("sentry:showError", onRenderError as EventListener);
-      };
-    },
-
-    processEvent: processEnvelope,
-
-    panels: () => {
-      const store = useSentryStore.getState();
-
-      const errorCount = store.getEvents().reduce((sum, e) => sum + Number(isErrorEvent(e)), 0);
```
Shubhdeep12, line 69: hey @betegon, in the recent PR, I saw that you changed something here.
But I guess there's no use in setting up telemetry, and I guess we are not using panels' function as well.

So i think it's better this to be removed.

packages/overlay/src/telemetry/sentry-integration.ts
betegon: i think so. like before we sent the spotlight events to spotlight but now se should get them using the electron sdk if im' not wrong.

packages/overlay/src/telemetry/index.ts
betegon: the only thing i'm concern about that is that we used it to get the notification count in the sidebar, so if you plan to maintain that in any way, ok!

Shubhdeep12: i think you already added the logic to count tracesCount and errorCount in TelemetryView.tsx

packages/overlay/src/telemetry/sentry-integration.ts
Shubhdeep12: Yeah, Electron already has the Sentry SDK added for crash reporting.
This was a different use case, it was about Spotlight integration for older versions of the Sentry SDK, where Spotlight support wasn’t included yet.
In other words, we were adding this integration to browser SDKs, where Spotlight was embedded by pulling the client from the browser.

betegon: perfect then!

packages/overlay/src/telemetry/index.ts
betegon: you're right, then I forgot to remove that. Thanks for doing so. I copied the errorCount from that file but thought i removed it. Thanks!

.changeset/spicy-candies-move.md
```diff
@@ -0,0 +1,6 @@
+---
+"@spotlightjs/overlay": patch
```
BYK:
```suggestion
"@spotlightjs/overlay": minor
"@spotlightjs/electron": patch
```

This is what we discussed with @betegon 

>I need to understand. Based on Semver, on the MINOR, it says:
>when you add a functionality...
>does it mean a user feature? I'm thinking of Shubhdeep's PR about url routing. That's a >functionality, although it's not a user facing feature. So for me it would be:
>Minor for overlay as we add url routing, that implies they could be copy/pasting url.
>Patch for electron as user's cannot see the routing in the app

And I responded:

>You are absolutely right!
>Joking aside, you are right and I thought the same way. Just didn't wanna make a big fuss about it as we're gonna release a major any way
>The way to think about this is:
> - Patch: you are fixing things and unless people need those specific fixes, they should be able to move forward and backward in patch versions safely
> - Minor: you are adding something new and once people upgrade to this and start using that new thing, they cannot go back safely. Moving forward in a minor should always be safe though
> - Major: you're changing how things work, everything is fair game and moving forward or backward across majors is not safe

.changeset/spicy-candies-move.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+- Replace MemoryRouter with BrowserRouter in overlay for url based routing.
+- Removed sentry-integration
```
BYK: Feel like this belongs to another PR?

packages/overlay/src/index.tsx
```diff
@@ -31,13 +30,6 @@ export async function onSevereEvent(cb: (count: number) => void) {
   on("severeEventCount", e => cb((e as CustomEvent).detail?.count ?? 1));
 }
 
-/**
- * Sentry telemetry configuration for Spotlight.
- */
-export function sentry(options: { injectIntoSDK?: boolean } = {}) {
-  return initTelemetry(options);
-}
```
BYK, line 39: I think this (an related stuff like `if (telemetry.setup)` below) should stay? Didn't see it in any other file as a newly added change.

packages/overlay/src/telemetry/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { connectToSidecar } from "../sidecar";
+import TelemetryView from "./components/TelemetryView";
+import useSentryStore from "./store";
+
+export function setSidecarUrl(url: string) {
+  const store = useSentryStore.getState();
+  const baseSidecarUrl = removeURLSuffix(url, "/stream");
+  store.setSidecarUrl(baseSidecarUrl);
+  log("Set sidecar URL for telemetry:", baseSidecarUrl);
+}
```
betegon: thanks for the refactor! Before it was all one big `initTelemetry` function

packages/overlay/src/index.tsx
betegon: do we really need that? par of the that change is a simplification to just init it by setting the sidecar url directly:

```typescript
useEffect(() => {
    if (!isInitialized) {
      setSidecarUrl(sidecarUrl);
      setIsInitialized(true);
    }
  }, [isInitialized, sidecarUrl]);
```


p.s. ok this was a response to @BYK comment on the same topic, I don't know why GH put it apart from it when i just answered on BYK thread.

.changeset/spicy-candies-move.md
betegon: i'd say maybe @Shubhdeep12 can improve the phrase but he removed some legacy code from integrations we don't really need now. (i'm referring to some of the simplified approach in [‎packages/overlay/src/telemetry/index.ts](https://github.com/getsentry/spotlight/pull/921/files/6974e0cf4e7eb8dc89c6fef43aef0f3ba921ecf9#diff-e10d786f89709a99fb937a31d47b3845a40fa46a509cb5d367d5402c45ba5ca3))

p.s. ok this was a response to @BYK comment on the same topic, I don't know why GH put it apart from it when i just answered on BYK thread.

packages/overlay/src/telemetry/index.tsx
Shubhdeep12: yeah it was 
- setting sidecar URL in store - DONE
- initiating spotlight-integration for sentry sdk where overlay was embedded - Not Needed
- Adding some functions to open up error page directly incase of error - Not Needed with standalone/electron app

packages/overlay/src/index.tsx
Shubhdeep12: yeah @BYK same question

- Are we planning to give flexibility to the users to enable/disable telemetry? OR to customise telemetry? And in future, other features like MCP, AI, etc separately?

Right now, Telemetry needs two props - `SidecarUrl: string` and `showClearEventsButton: boolean` and both depends on Overlay init props.

Shubhdeep12:
hey @betegon @BYK, regarding changest and a minor fix in telemetry event listeners please check - https://github.com/getsentry/spotlight/pull/921/commits/e2724f9a110a05ef2d1f78296aa3ae7bb7e423ff

packages/overlay/src/telemetry/index.tsx
```diff
@@ -1,6 +1,5 @@
 import type { Envelope } from "@sentry/core";
 import { useCallback, useEffect, useMemo, useState } from "react";
-import { useNavigate } from "react-router-dom";
 import { removeURLSuffix } from "~/lib/removeURLSuffix";
 import * as db from "../lib/db";
```
BYK: Pretty sure this module is also obsolete now

packages/overlay/src/telemetry/components/TelemetrySidebar.tsx
BYK: `pathname.startsWith('/telemetry')` should cover both cases, right?

packages/overlay/src/index.tsx
BYK: That's not my point. The initTelemetry here is the telemetry we send to Sentry using Sentry SDK. Sorry btw I posted this comment earlier but it was left in "pending" state as part of a general review

BYK: I got confused sorry. `initSentry` was somewhere else so resolving this comment.

packages/electron/electron-builder.js
```diff
@@ -21,6 +21,7 @@ let afterSign = "scripts/notarize.js";
 
 if (!process.env.CSC_LINK || !process.env.CSC_KEY_PASSWORD) {
   mac = {
+    ...mac,
```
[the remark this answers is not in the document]
betegon: merging the PR, but you may want to revisit this in a follow up @Shubhdeep12

betegon:
thanks @Shubhdeep12 , I tested it locally just in case and It works. merging!

PR #922, opened 2025-09-11
feat: added auto updater in electron

MathurAditya724:
- [x] Added auto updater
- [x] Removed unused code

### DIFF
--- a/packages/electron/index.html
+++ b/packages/electron/index.html
@@ -1,50 +1,51 @@
-<!doctype html>
+<!DOCTYPE html>
 <html>
+  <head>
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="script-src 'self'; worker-src 'self' blob:"
+    />
+    <title>Spotlight</title>
+    <style>
+      html,
+      body {
+        margin: 0;
+        padding: 0;
+        height: 100%;
+        width: 100%;
+      }
 
-<head>
-  <meta http-equiv="Content-Security-Policy" content="script-src 'self'; worker-src 'self' blob:" />
-  <title>Spotlight</title>
-  <style>
-    html,
-    body {
-      margin: 0;
-      padding: 0;
-      height: 100%;
-      width: 100%;
-    }
+      .error {
+        color: #fff;
+        text-shadow: 0 0 5px #fff;
+        text-align: center;
+        font-family: system-ui, sans-serif;
+      }
 
-    .error {
-      color: #fff;
-      text-shadow: 0 0 5px #fff;
-      text-align: center;
-      font-family: system-ui, sans-serif;
-    }
+      .header {
+        letter-spacing: 6px;
+        font-weight: 500;
+        font-size: 3rem;
+        line-height: 1.5;
+      }
 
-    .header {
-      letter-spacing: 6px;
-      font-weight: 500;
-      font-size: 3rem;
-      line-height: 1.5;
-    }
+      .error-page-navbar {
+        display: flex;
+        width: 100%;
+        gap: 5px;
+        align-items: center;
+      }
 
-    .error-page-navbar {
-      display: flex;
-      width: 100%;
-      gap: 5px;
-      align-items: center;
-    }
+      .spotlight-title {
+        color: #fff;
+        font-weight: 700;
+        font-size: 1rem;
+        font-family: system-ui, sans-serif;
+      }
+    </style>
+  </head>
 
-    .spotlight-title {
-      color: #fff;
-      font-weight: 700;
-      font-size: 1rem;
-      font-family: system-ui, sans-serif;
-    }
-  </style>
-</head>
-
-<body>
-  <script type="module" src="./src/index.ts"></script>
-</body>
-
+  <body>
+    <script type="module" src="./src/index.ts"></script>
+  </body>
 </html>


packages/electron/src/electron/main/index.ts
```diff
@@ -2,10 +2,62 @@ import path from "node:path";
...
 import { BrowserWindow, Menu, app, dialog, ipcMain, shell } from "electron";
+import { autoUpdater as nativeUpdater } from "electron";
 import Store from "electron-store";
+import { autoUpdater } from "electron-updater";
 
 const store = new Store();
 
+autoUpdater.forceDevUpdateConfig = true;
+
+const ONE_HOUR = 1 * 60 * 1000;
```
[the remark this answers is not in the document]
BYK: @MathurAditya724 I think this one is also legit? :)

packages/electron/src/electron/main/index.ts
```diff
@@ -2,10 +2,62 @@ import path from "node:path";
 import * as Sentry from "@sentry/electron/main";
 import { clearBuffer, setupSidecar } from "@spotlightjs/sidecar";
 import { BrowserWindow, Menu, app, dialog, ipcMain, shell } from "electron";
+import { autoUpdater as nativeUpdater } from "electron";
 import Store from "electron-store";
+import { autoUpdater } from "electron-updater";
 
 const store = new Store();
 
+autoUpdater.forceDevUpdateConfig = true;
```
BYK: Is this staying in or out before merge?

packages/electron/src/electron/main/index.ts
```diff
@@ -2,10 +2,62 @@ import path from "node:path";
...
+    app.removeAllListeners("window-all-closed");
+
+    for (const win of BrowserWindow.getAllWindows()) {
+      if (win.isDestroyed()) continue;
+      win.removeAllListeners("close");
+      win.close();
+    }
+
+    nativeUpdater.once("before-quit-for-update", () => {
+      console.log("This was called!");
```
BYK:
```suggestion
```

packages/electron/src/electron/main/index.ts
```diff
@@ -2,10 +2,62 @@ import path from "node:path";
...
+    });
+  }
+
+  autoUpdater.quitAndInstall();
+}
+
+app.on("ready", () => {
+  const updateInterval = setInterval(() => {
+    autoUpdater.checkForUpdates();
+  }, ONE_HOUR);
```
BYK: It's always better to use `setTimeout` with self-reschedule as if `checkForUpdates()` hangs or keeps erroring out for some reason, you have no way of knowing that and adjusting the interval based on this data.

For more info, this looks like a good read: https://reallifejs.com/brainchunks/repeated-events-timeout-or-interval/

packages/electron/src/electron/main/index.ts
```diff
@@ -2,10 +2,62 @@ import path from "node:path";
...
+    dialog
+      .showMessageBox({
+        type: "question",
+        message: "A new update has been downloaded. It will be installed on restart.",
+        buttons: ["Restart", "Later"],
+      })
+      .then(result => {
+        if (result.response === 0) {
+          clearInterval(updateInterval);
+          installAndRestart();
```
BYK: Extreme case example:

1. we downloaded an update
2. the user clicked "later"
3. we stopped checking
4. some non-trivial amount of time passed
5. we released a new version
6. the user never got this until a restart as we stopped checking

In a worst-case scenario, if you release a broken version that cannot auto-update for some reason (crash on launch, broken updates etc), even if you release a new fixed version very quickly, you cannot save these people.

Long story short, unless we have a great reason, we should never stop checking for updates.

packages/electron/dev-app-update.yml
```diff
@@ -0,0 +1,4 @@
+owner: getsentry
+repo: spotlight
+provider: github
+updaterCacheDirName: spotlight-dev-updater
```
BYK, line 4: Why `dev`?

packages/electron/electron-builder.js
```diff
@@ -16,6 +16,21 @@ let mac = {
...
+        localhost: {
+          NSTemporaryExceptionAllowsInsecureHTTPSLoads: false,
+          NSIncludesSubdomains: false,
+          NSTemporaryExceptionAllowsInsecureHTTPLoads: true,
+          NSTemporaryExceptionMinimumTLSVersion: "1.0",
+          NSTemporaryExceptionRequiresForwardSecrecy: false,
+        },
+      },
+    },
+  },
```
BYK: Are these relevant? If yes, can you add some comments about what they are and why we need them?

BYK:
And we need a changelog entry for this please :)

packages/electron/src/electron/main/index.ts
MathurAditya724: Yes, this is need for checking things in dev mode

packages/electron/dev-app-update.yml
MathurAditya724: This file is used for checking things in dev mode, electron-builder creates another one for production out of the box

packages/electron/electron-builder.js
MathurAditya724: These were getting auto generated, so removed them

packages/electron/src/electron/main/index.ts
```diff
@@ -2,10 +2,68 @@ import path from "node:path";
...
+app.on("ready", () => {
+  let updateInfo: UpdateCheckResult | null = null;
+
+  async function checkForUpdates() {
+    updateInfo = await autoUpdater.checkForUpdates();
+
+    if (!updateInfo?.isUpdateAvailable) {
+      setTimeout(checkForUpdates, ONE_HOUR);
+    }
+  }
```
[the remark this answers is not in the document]
BYK: What Cursor says above?

packages/electron/src/electron/main/index.ts
BYK: I really am not comfortable shipping this to prod as the docs mention this under "debugging" and it feels like it is not supposed to be shipped to prod: https://github.com/electron-userland/electron-builder/pull/7117/files

Can we tie that to an env variable like `SPOTLIGHT_DEBUG` or `SPOTLIGHT_FORCE_DEV_UPDATES` or something please?

packages/electron/src/electron/main/index.ts
```diff
@@ -34,23 +37,22 @@ function installAndRestart() {
...
-      if (!updateInfo?.isUpdateAvailable) {
-        setTimeout(checkForUpdates, ONE_HOUR);
-      }
-    } catch (error) {
-      console.error(error);
+    if (!updateInfo?.isUpdateAvailable) {
+      timeout = 3 * ONE_HOUR;
     }
+  } catch (error) {
+    console.error(error);
```
BYK, line 50: We should report updater errors to Sentry, right? Sending them to console will essentially make them disappear?

packages/electron/src/electron/main/index.ts
```diff
@@ -34,23 +37,22 @@ function installAndRestart() {
...
+    updateInfo = await autoUpdater.checkForUpdates();
 
-      if (!updateInfo?.isUpdateAvailable) {
-        setTimeout(checkForUpdates, ONE_HOUR);
-      }
-    } catch (error) {
-      console.error(error);
+    if (!updateInfo?.isUpdateAvailable) {
+      timeout = 3 * ONE_HOUR;
     }
```
BYK: Can you explain this logic please? I really cannot wrap my head around. We check every 24 hours if an update is available but normally we do every 3 hours? Why?

packages/electron/src/electron/main/index.ts
```diff
@@ -8,7 +8,10 @@ import { type UpdateCheckResult, autoUpdater } from "electron-updater";
 
 const store = new Store();
 
-autoUpdater.forceDevUpdateConfig = process.env.NODE_ENV === "development";
+autoUpdater.forceDevUpdateConfig = true;
+
+const ONE_HOUR = 60 * 60 * 1000;
+let updateInfo: UpdateCheckResult | null;
```
BYK:
```suggestion
let updateInfo: UpdateCheckResult | null = null;
```

So you don't have to suppress the TypeScript error that you incorrectly suppress below.

packages/electron/src/electron/main/index.ts
```diff
@@ -139,6 +141,9 @@ app.on("window-all-closed", () => {
 
 const isMac = process.platform === "darwin";
 
+// @ts-expect-error updateInfo is initialized in the app.on("ready") event
```
BYK: We should not be suppressing this error and fix it instead. See my comment above.

packages/electron/src/electron/main/index.ts
```diff
@@ -2,10 +2,74 @@ import path from "node:path";
...
+        type: "question",
+        message: "A new update has been downloaded. It will be installed on restart.",
+        buttons: ["Restart", "Later"],
+      })
+      .then(result => {
+        if (result.response === 0) {
+          installAndRestart();
+        }
+      });
+  });
```
BYK: Can we not make the handler async and writh this as
```suggestion
  autoUpdater.on("update-downloaded", async () => {
    const {response} = await dialog
      .showMessageBox({
        type: "question",
        message: "A new update has been downloaded. It will be installed on restart.",
        buttons: ["Restart", "Later"],
      });
      if (response === 0) {
        installAndRestart();
      }
  });
```

BYK (changes requested):

packages/electron/src/electron/main/index.ts
MathurAditya724: Done, added `process.env.NODE_ENV === "development"`

packages/electron/src/electron/main/index.ts
MathurAditya724: Oops! This wasn't supposed to be pushed, I was testing out something. Sorry I will remove this.

We will be just checking every hour for updates, even if we have a update available.

packages/electron/src/electron/main/index.ts
MathurAditya724: Don't need this variable anymore, was trying out dynamic menu but changed the approach.

Was getting another error when adding `= null`, that's why and wanted to test first before working on another error. Tho this was not working anyways

packages/electron/src/electron/main/index.ts
MathurAditya724: Yes we can, mate!

.changeset/pretty-wasps-fly.md
BYK, line 1: Why do we see this as a new file? It is from #932

.github/workflows/publish.yml
```diff
@@ -173,6 +173,7 @@ jobs:
           MAIN_VITE_SENTRY_AUTH_TOKEN: ${{ secrets.MAIN_VITE_SENTRY_AUTH_TOKEN }}
           CSC_LINK: ${{ secrets.CSC_LINK }}
           CSC_KEY_PASSWORD: ${{ secrets.CSC_KEY_PASSWORD }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
BYK, line 178: Why do we need this, maybe a comment?

packages/electron/src/electron/main/index.ts
```diff
@@ -1,12 +1,82 @@
...
+    };
+
+    const _template = [...template];
+    if (_template[0].submenu?.[1].id === "check-for-updates") _template[0].submenu[1] = menuItem;
+    Menu.setApplicationMenu(Menu.buildFromTemplate(_template));
+  });
+
+  autoUpdater.on("error", error => {
+    Sentry.captureException(error);
+  });
```
BYK: Let's drop the braces, we are no longer teenagers :P

```suggestion
  autoUpdater.on("error", error => Sentry.captureException(error));
```

packages/electron/src/electron/main/index.ts
```diff
@@ -1,12 +1,82 @@
...
+
+async function checkForUpdates() {
+  try {
+    await autoUpdater.checkForUpdates();
+  } catch (error) {
+    console.error(error);
+    Sentry.captureException(error);
+  }
+
+  setTimeout(checkForUpdates, ONE_HOUR);
```
BYK: We should have a global `let CHECK_UPDATES_TIMER: Timeout | null = null` and then do
```
if (CHECK_UPDATES_TIMER) {
  clearTimeout(CHECK_UPDATES_TIMER);
  CHECK_UPDATES_TIMER = null;
}
....
CHECK_UPDATES_TIMER = setTimeout(checkForUpdates, ONE_HOUR);
```

to prevent multiple concurrent check update calls getting scheduled if something else in the code tries to call `checkForUpdates()` separately.

packages/electron/src/electron/main/index.ts
```diff
@@ -1,12 +1,82 @@
...
+
+    menuItem.label = "Restart to Update";
+    menuItem.click = () => {
+      installAndRestart();
+    };
+
+    const _template = [...template];
+    if (_template[0].submenu?.[1].id === "check-for-updates") _template[0].submenu[1] = menuItem;
+    Menu.setApplicationMenu(Menu.buildFromTemplate(_template));
+  });
```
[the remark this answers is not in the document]
BYK: Right, why is this duplicated??? (like the handler code)

packages/electron/src/electron/main/index.ts
```diff
@@ -308,7 +404,7 @@ store.onDidChange("sentry-send-envelopes", newValue => {
 const showErrorMessage = () => {
   if (win) {
     win.webContents.executeJavaScript(`{
-      const sentryRoot = document.getElementById('sentry-spotlight-root');
+      const sentryRoot = document.getElementById('spotlight-root');
```
BYK: Intentional?

packages/electron/electron-builder.js
```diff
@@ -49,9 +56,22 @@ const config = {
     "!{tsconfig.json,tsconfig.node.json,tsconfig.web.json}",
   ],
   mac,
+  dmg: {
```
BYK: What is this for?

packages/electron/electron-builder.js
```diff
@@ -5,15 +5,22 @@ const builder = require("electron-builder");
...
       arch: ["x64", "arm64"],
     },
   ],
   icon: "resources/icons/mac/icon.icns",
+  type: "distribution",
   hardenedRuntime: true,
   gatekeeperAssess: false,
   entitlements: "build/entitlements.mac.plist",
   entitlementsInherit: "build/entitlements.mac.plist",
+  // Additional entitlements for auto-updater compatibility
```
BYK: Yes but why do we need these? What happens when we don't have these?

BYK (changes requested):
I think you had some issue during merging with the latest main so seems like you need to do some clean up.

packages/electron/src/electron/main/index.ts
MathurAditya724: Yes, there is nothing like "sentry-spotlight-root"

packages/electron/src/electron/main/index.ts
```diff
@@ -95,7 +169,21 @@ const createWindow = () => {
...
+       errorScreen.style.display = 'none';
+       errorScreen.innerHTML = \`
+        <div class="error-page-navbar">
+        <img alt="spotlight-icon" src="./resources/sized.png" width="50" height="50" />
+        <p class="spotlight-title">Spotlight</p>
+      </div>
+      <h1 class="error header">Oops! An error occurred.</h1>
+      <p class="error description">Press Cmd + R to reload the app.</p>
+       \`;
+       document.body.appendChild(errorScreen);`,
```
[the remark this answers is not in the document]
MathurAditya724: The error page in the electron app gets replaced once the spotlight init function is called. Most like this is happening because of `ReactDOM.createRoot(appRoot)` in the overlay package. So even tho we have this in the index.html, at runtime it's not there. So like the draggable component we need to create this too, and now point in having the static version as it's `"display: none"`, so even at early error this won't be showing.

I believe by merging the `app.whenReady()`  and `app.on("ready")` we can better handle this situation in a try/catch. That will be a separate PR

.github/workflows/publish.yml
```diff
@@ -173,6 +173,8 @@ jobs:
           MAIN_VITE_SENTRY_AUTH_TOKEN: ${{ secrets.MAIN_VITE_SENTRY_AUTH_TOKEN }}
           CSC_LINK: ${{ secrets.CSC_LINK }}
           CSC_KEY_PASSWORD: ${{ secrets.CSC_KEY_PASSWORD }}
+          # Need this for auto-updater in electron
```
BYK: Yes, but WHY? 😀 Why does the auto-updater in electron need the GH token? Do something with the releases? Add an asset? What? (so we know how to scope this in a future security audit)

packages/electron/src/electron/main/index.ts
```diff
@@ -1,12 +1,82 @@
...
+  autoUpdater.quitAndInstall();
+}
+
+let checkingForUpdatesTimeout: NodeJS.Timeout | null = null;
+async function checkForUpdates() {
+  try {
+    if (checkingForUpdatesTimeout) {
+      clearTimeout(checkingForUpdatesTimeout);
+      checkingForUpdatesTimeout = null;
+    }
```
BYK: This part should not be in the `try` block.

packages/electron/src/electron/main/index.ts
```diff
@@ -1,12 +1,82 @@
...
+  checkForUpdates();
+
+  autoUpdater.on("update-downloaded", async () => {
+    const menuItem = isMac ? { ...template[0].submenu?.[1] } : undefined;
+
+    if (menuItem) {
+      menuItem.label = "Restart to Update";
+      menuItem.click = () => installAndRestart();
+
+      const _template = [...template];
```
BYK, line 68: It's better to move this `app.on("ready"` handler after `const template` definition below. This threw me off as I looked for the definition of `template` above here and could not find it (like Cursor) 😅

packages/electron/src/electron/main/index.ts
```diff
@@ -1,12 +1,81 @@
...
+    }
+
+    await autoUpdater.checkForUpdates();
+  } catch (error) {
+    console.error(error);
+    Sentry.captureException(error);
+  }
+
+  checkingForUpdatesTimeout = setTimeout(checkForUpdates, ONE_HOUR);
+}
```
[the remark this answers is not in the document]
BYK: @MathurAditya724 -- Cursor is right here but it is not as severe so we can think about this later. Scenario:

1. Call `checkForUpdates()`
2. `checkForUpdates()` schedules itself for the next hour
3. The next hour comes and `checkForUpdates` starts running
4. While `checkForUpdates()` is `await`ing for `autoUpdater.checkForUpdates()` another call is made for `checkForUpdates()`
5. The first call finishes, and schedules a new call in an hour -- timeout is stored in `checkingForUpdatesTimeout`
6. Right after this, the second call finishes, overwriting the old `checkingForUpdatesTimeout` without clearing it so now we have 2 checks scheduled for the next hour.

A solution to this would be to cache the promise and return that so multiple calls to `checkForUpdates()` while its awaiting returns the same pending promise.

packages/electron/src/electron/main/index.ts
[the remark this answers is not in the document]
BYK: We should put your comment in the code if we are not going to fix this properly from overlay side (or some other way)

packages/electron/src/electron/main/index.ts
MathurAditya724: Will do this in another PR

packages/electron/src/electron/main/index.ts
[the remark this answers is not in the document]
MathurAditya724: Added `isCheckingForUpdates` which takes care of this

.github/workflows/publish.yml
```diff
@@ -173,7 +173,8 @@ jobs:
           MAIN_VITE_SENTRY_AUTH_TOKEN: ${{ secrets.MAIN_VITE_SENTRY_AUTH_TOKEN }}
           CSC_LINK: ${{ secrets.CSC_LINK }}
           CSC_KEY_PASSWORD: ${{ secrets.CSC_KEY_PASSWORD }}
-          # Need this for auto-updater in electron
+          # Needed by electron-builder's Github Provider for publishing,
+          # This helps with auto-updater as we are using Github Releases
```
BYK, line 177: 🔥

PR #923, opened 2025-09-11
fix: electron app startup issue

MathurAditya724:
The issue was occurring because of 2 bugs -

1. The main issue, `afterAllSetup` handler function, this sends a serialized envelope to the Spotlight frontend. Even tho we are not using this envelope it is getting passed through the `processEnvelope` function. Now because we are handling the parsing/processing of envelopes on the backend, this was failing.
2. Because of the above issue, the electron app was trying to run `showErrorMessage` before showing the "askForPermission" dialog which was also failing.

Due to these reason the app was unable to startup.

Also some minor improvements I did while figuring out the issues
- [x] Added types for electron's contextBridge
- [x] Deleted props for the old Spotlight Integration
- [x] Changed Event to Envelope (Performance improvement)

### DIFF
--- a/packages/electron/tests/launch.test.ts
+++ b/packages/electron/tests/launch.test.ts
@@ -1,5 +1,9 @@
 import { _electron as electron, expect, test } from "@playwright/test";
 
+function sleep(ms: number) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
 test("launch app", async () => {
   const app = await electron.launch({ args: ["./out/main/index.js", "--no-sandbox"] });
 
@@ -14,7 +18,15 @@ test("launch app", async () => {
   // Wait for the first BrowserWindow to open
   // and return its Page object
   const window = await app.firstWindow();
-  await window.getByRole("navigation", { name: "Navigation" }).getByText("Spotlight").waitFor({ state: "visible" });
+
+  const sidebar = window.getByRole("navigation", { name: "Navigation" });
+
+  await sidebar.getByText("Spotlight").waitFor({ state: "visible" });
+
+  await sleep(2000);
+
+  expect(sidebar.getByText("Traces").first()).toHaveText("Traces1");
+  expect(sidebar.getByText("Errors").first()).toHaveText("Errors");
 
   // close app
   await app.close();


packages/electron/src/electron/main/index.ts
```diff
@@ -275,13 +275,11 @@ store.onDidChange("sentry-send-envelopes", newValue => {
...
-      const sentryRoot = document.getElementById('sentry-spotlight-root');
-      const errorScreen = document.getElementById('error-screen');
-      if(sentryRoot){
-        sentryRoot.style.display = 'none';
+      if (document.getElementById('sentry-spotlight-root')) {
+        document.getElementById('sentry-spotlight-root').style.display = 'none';
       }
-      if(errorScreen){
-        errorScreen.style.display = 'block';
+      if (document.getElementById('error-screen')) {
```
BYK: Why are we removing the variables here?

MathurAditya724: When running multiple times, it complains that the variable is already declared

BYK: I'd rather fix that rather than doing this weird stuff?

MathurAditya724: Added a scope

packages/overlay/src/telemetry/components/insights/envelopes/EnvelopeDetails.tsx
```diff
@@ -20,7 +20,7 @@ export default function EnvelopeDetails({ envelope }: { envelope: Envelope }) {
         title="Envelope Details"
         subtitle={
           <>
-            Event Id <span className="text-primary-500">&mdash;</span>{" "}
+            Envelope Id <span className="text-primary-500">&mdash;</span>{" "}
```
BYK, line 23: How is this relevant to the patch?

packages/overlay/src/telemetry/sentry-integration.ts
```diff
@@ -39,7 +38,10 @@ export const spotlightIntegration = () => {
     },
     afterAllSetup: (client: Client) =>
       client.on("beforeEnvelope", (envelope: Envelope) =>
-        trigger("event", { contentType: "application/x-sentry-envelope", data: serializeEnvelope(envelope) }),
+        trigger("event", {
```
BYK, line 41: Why this change? Also how is it relevant to the patch? Split this into its own PR please.

MathurAditya724: Added details for this in the PR description

packages/overlay/src/telemetry/components/insights/envelopes/EnvelopeDetails.tsx
MathurAditya724: Added details for this in the PR description

PR #911, opened 2025-08-26
Added copy prompt button

Shubhdeep12:
Before opening this PR:

- [x] I added a [Changeset Entry](https://spotlightjs.com/contribute/changesets/) with `pnpm changeset:add`
- [ ] I referenced issues that this PR addresses

- Added Copy Button on Error Details page to get the prompt for AI agents.
- Used the same formatting as in the MCP server for errors.
image

PR #926, opened 2025-09-15
feat: Added AI Copy button

MathurAditya724:
This PR is inspired by the work of @Shubhdeep12 in PR #911

- [x] Added AI Button
- [x] Used the formatter from the sidecar to generate the markdown
- [x] Added `'format'` and `'parser'` in the tsdown config, as it was not getting bundled

### DIFF
--- a/packages/overlay/src/telemetry/components/events/AICopyButton.tsx
+++ b/packages/overlay/src/telemetry/components/events/AICopyButton.tsx
@@ -1,0 +1,65 @@
+import { formatEventOutput, processErrorEvent } from "@spotlightjs/sidecar/format";
+import { type MouseEvent, useCallback, useState } from "react";
+import { ReactComponent as CheckIcon } from "~/assets/check.svg";
+import { ReactComponent as CopyIcon } from "~/assets/copy.svg";
+import { Button } from "~/ui/button";
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "~/ui/tooltip";
+import type { SentryErrorEvent } from "../../types";
+
+interface AICopyButtonProps {
+  event: SentryErrorEvent;
+}
+
+async function copyErrorToClipboard(prompt: string, setIsCopied: (copied: boolean) => void): Promise<void> {
+  await navigator.clipboard.writeText(prompt);
+
+  setIsCopied(true);
+  setTimeout(() => {
+    setIsCopied(false);
+  }, 1000);
+}
+
+export default function AICopyButton({ event }: AICopyButtonProps) {
+  const [isCopied, setIsCopied] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleClick = useCallback(
+    async (evt: MouseEvent) => {
+      evt.stopPropagation();
+      if (isLoading) return;
+
+      setIsLoading(true);
+      await copyErrorToClipboard(formatEventOutput(processErrorEvent(event as any)), setIsCopied);
+      setIsLoading(false);
+    },
+    [event],
+  );
+
+  return (
+    <TooltipProvider>
+      <Tooltip>
+        <TooltipTrigger asChild>
+          <Button
+            size="sm"
+            variant="outline"
+            onClick={handleClick}
+            disabled={isLoading}
+            className="border-primary-700 hover:bg-primary-900 bg-primary-950 h-10 mix-blend-screen hover:text-white cursor-pointer flex items-center gap-1 disabled:opacity-50 disabled:cursor-not-allowed"
+          >
+            {isLoading ? (
+              <div className="w-[18px] h-[18px] border-2 border-primary-50 border-t-transparent rounded-full animate-spin" />
+            ) : isCopied ? (
+              <CheckIcon width={18} height={18} title="Copied!" className="stroke-primary-50 transition-all" />
+            ) : (
+              <CopyIcon width={18} height={18} title="Copy" className="stroke-primary-50 transition-all" />
+            )}
+            <span className="text-xs">✨ AI</span>
+          </Button>
+        </TooltipTrigger>
+        <TooltipContent>
+          <p>Copy error details to use with any AI service</p>
+        </TooltipContent>
+      </Tooltip>
+    </TooltipProvider>
+  );
+}
--- a/packages/overlay/src/telemetry/components/events/EventDetails.tsx
+++ b/packages/overlay/src/telemetry/components/events/EventDetails.tsx
@@ -1,9 +1,10 @@
 import { Link, Outlet, Route, Routes, useParams } from "react-router-dom";
 import useSentryStore from "../../store";
-import type { SentryEvent } from "../../types";
+import type { SentryErrorEvent, SentryEvent } from "../../types";
 import { createTab } from "../../utils/tabs";
 import TelemetryTabs from "../TelemetryTabs";
 import PlatformIcon from "../shared/PlatformIcon";
+import AICopyButton from "./AICopyButton";
 import Event, { EventTitle } from "./Event";
 import EventBreadcrumbs from "./EventBreadcrumbs";
 import EventContexts from "./EventContexts";
@@ -33,11 +34,14 @@ export default function EventDetails() {
   ];
 
   const traceCtx = event.contexts?.trace;
+  const isErrorEvent = "exception" in event;
+
   return (
     <div className="w-full flex flex-1 flex-col overflow-y-auto overflow-x-hidden">
       <div className="bg-primary-950 flex items-center gap-x-2 px-6 py-4">
         <PlatformIcon event={event} className="rounded-md" />
         <h1 className="max-w-full flex-1 truncate text-2xl">{renderEventTitle(event)}</h1>
+        {isErrorEvent && <AICopyButton event={event as SentryErrorEvent} />}
         {traceCtx && (
           <div className="text-primary-300 font-mono">
             <div>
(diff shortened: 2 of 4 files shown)


Shubhdeep12 (approved):
lgtm

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
```diff
@@ -0,0 +1,65 @@
+import { formatEventOutput, processErrorEvent } from "@spotlightjs/sidecar/format";
```
BYK, line 1: I don't think the UI should have a dependency on the sidecar. We should move this into the core package.

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { type MouseEvent, useCallback, useState } from "react";
+import { ReactComponent as CheckIcon } from "~/assets/check.svg";
+import { ReactComponent as CopyIcon } from "~/assets/copy.svg";
+import { Button } from "~/ui/button";
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "~/ui/tooltip";
+import type { SentryErrorEvent } from "../../types";
```
BYK, line 7: These types should also be moved into the core package.

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+interface AICopyButtonProps {
+  event: SentryErrorEvent;
+}
+
+async function copyErrorToClipboard(prompt: string, setIsCopied: (copied: boolean) => void): Promise<void> {
+  await navigator.clipboard.writeText(prompt);
```
BYK: There's a note about this in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText):

>Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText#browser_compatibility).

Now `localhost` is still considered "secure context" but for people running Spotlight differently may encounter issues. I propose adding a check like this with a warning:

```ts
try {
    await navigator.clipboard.writeText(prompt);
} catch (err) {
    if (!window.secureContext) {
        // Alert the user about secure contexts)
    } else {
        // Show a generic error message and report to Sentry
    }
}
```

Also see https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#feature_detection

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleClick = useCallback(
+    async (evt: MouseEvent) => {
+      evt.stopPropagation();
+      if (isLoading) return;
+
+      setIsLoading(true);
+      await copyErrorToClipboard(formatEventOutput(processErrorEvent(event as any)), setIsCopied);
+      setIsLoading(false);
```
BYK: Why do we wrap this action in a loading indicator? Isn't it almost instant and w/o any server interaction?

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+            className="border-primary-700 hover:bg-primary-900 bg-primary-950 h-10 mix-blend-screen hover:text-white cursor-pointer flex items-center gap-1 disabled:opacity-50 disabled:cursor-not-allowed"
+          >
+            {isLoading ? (
+              <div className="w-[18px] h-[18px] border-2 border-primary-50 border-t-transparent rounded-full animate-spin" />
+            ) : isCopied ? (
+              <CheckIcon width={18} height={18} title="Copied!" className="stroke-primary-50 transition-all" />
+            ) : (
+              <CopyIcon width={18} height={18} title="Copy" className="stroke-primary-50 transition-all" />
+            )}
+            <span className="text-xs">✨ AI</span>
```
BYK: Can we have something more descriptive like `Copy for AI`? (I see the tooltip below but a user shouldn't have to hover onto something to understand what it is for)

packages/overlay/src/telemetry/components/events/EventDetails.tsx
```diff
@@ -33,11 +34,14 @@ export default function EventDetails() {
   ];
 
   const traceCtx = event.contexts?.trace;
+  const isErrorEvent = "exception" in event;
```
BYK: Can we use the `isErrorEvent` helper that we should have _somewhere_?

packages/sidecar/tsdown.config.ts
```diff
@@ -1,6 +1,13 @@
...
 
 export default defineConfig({
-  entry: ["./src/main.ts", "./src/vite-plugin.ts", "./src/constants.ts", "./server.ts"],
+  entry: [
+    "./src/main.ts",
+    "./src/vite-plugin.ts",
+    "./src/constants.ts",
+    "./server.ts",
+    "./src/parser/index.ts",
+    "./src/format/index.ts",
```
BYK, line 10: Are these related to this patch or they should have been part of an earlier PR but was missed?

MathurAditya724: Yes, these were missed

BYK: Would you mind splitting this into its own PR, referencing in which PR they were missed?

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
MathurAditya724: Should I add `react-hot-toast` package to show these messages? https://www.npmjs.com/package/react-hot-toast

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
MathurAditya724: That is true

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
MathurAditya724: Removing the tooltip and putting "Copy for AI" in the button text

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
MathurAditya724: Also, wouldn't it be better to not show this button in an unsecured context? We can just add `window.isSecureContext` next to the condition for showing the button

BYK: I like the conditional visibility (or maybe keep the button, disable it of it's not secure context and the tooltip explains the reason).

We'd still need to handle any potential errors and acknowledge to the user that we failed. Not sure if it's worth pulling in am entire notification library for just one info message.

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+      setIsCopied(true);
+      try {
+        await navigator.clipboard.writeText(formatEventOutput(processErrorEvent(event as any)));
+      } catch (err) {
+        console.error(err);
+      }
+
+      setTimeout(() => {
+        setIsCopied(false);
+      }, 1000);
```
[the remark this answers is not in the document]
BYK: This looks legit

packages/sidecar/tsdown.config.ts
MathurAditya724: Done #927

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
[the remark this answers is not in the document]
MathurAditya724: Easy fix

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
MathurAditya724: Done, for the time being I am doing `console.error` then or do you want to show the error below the button? I think that wouldn't look nice

BYK: Would they even see the console? Oh, we can probably send the error to Spotlight itself! 😅😅 Future PR tho, fine for now.

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
```diff
@@ -10,27 +10,30 @@ interface AICopyButtonProps {
...
-      if (isLoading) return;
 
-      setIsLoading(true);
-      await copyErrorToClipboard(formatEventOutput(processErrorEvent(event as any)), setIsCopied);
-      setIsLoading(false);
+      setIsCopied(true);
+      try {
+        await navigator.clipboard.writeText(formatEventOutput(processErrorEvent(event as any)));
+      } catch (err) {
+        console.error(err);
```
BYK, line 35: You gotta send this error to Sentry manually now that you catch it.

BYK (approved):

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
MathurAditya724: Sure, I can use the "sendEvent" helper function. Will create another PR for this

PR #929, opened 2025-09-16
electron(feat): App running in background + tray icon

betegon:
Now, doing `cmd/ctrl + w` or clicking the "X" on spotlight electron app won't close it, it will keep running in the background and it will have a tray icon in macos/windows.

Please @BYK, test it in windows. Attaching MacOS demo below to see behaviour.

To try it:

`pnpm run build`
`pnpm run dev:electron`

https://github.com/user-attachments/assets/5d8b2a5d-f6b7-43a6-9d28-f00d4adc45bb

There could be behaviours that may not be great, we'll address them in a follow up PR if we don't like them. A couple that come to my mind:
- if you have start at login enabled, when you restart/boot your laptop, it will open up the window, instead of just run in the background.
- if you click the tray icon, it will open the window. Other apps may show you a contextual menu (right click for us). you can see this behaviour in spotlight in the video above


packages/electron/src/electron/main/index.ts
```diff
@@ -360,11 +393,74 @@ function storeIncomingPayload(body: string) {
...
+    {
+      label: "Quit Spotlight",
+      click: () => {
+        isQuitting = true;
+        app.quit();
+      },
+    },
+  ]);
+
+  tray.setToolTip("Spotlight - Sidecar is running");
```
BYK, line 444: We should add the port here too and make that dynamic from the sidecar URL

packages/electron/src/electron/main/index.ts
```diff
@@ -360,11 +393,74 @@ function storeIncomingPayload(body: string) {
...
+  tray.setContextMenu(contextMenu);
+
+  tray.on("click", () => {
+    showOrCreateWindow();
```
BYK, line 448: I'd say clicking on the tray icon should toggle minimize/show behavior.

PR #924, opened 2025-09-12
ref(overlay): remove local resources handling

betegon:
Small refactor (aka deletion) of legacy code used for the overlay feature.
- Removes local resources (traces, spans,...) UI and parsing logic. Now we always show all resources.
- Removes a vite plugin that injected the overlay.
- Improves description of a past changeset 😬 .

PR #930, opened 2025-09-17
ref(overlay): small cleanup

betegon:
#924 follow up.

- Replaced destructured usage of `allSpans`, `allTraces`, `allLogs` and `allEnvelopes` from their respective hooks with a direct assignment.
- simplified `errorCount` calculation.


BYK (approved):
🚀

PR #933, opened 2025-09-17
feat(sidecar): Add stdio transport for MCP

BYK:


packages/sidecar/src/mcp/mcp.ts
```diff
@@ -12,7 +12,7 @@ import {
 import { getBuffer } from "~/utils/index.js";
 import { NO_ERRORS_CONTENT, NO_LOGS_CONTENT } from "./constants.js";
 
-export function createMcpInstance() {
+export function createMCPInstance() {
   const mcp = new McpServer({
```
betegon, line 16: shouldn't we track this with our instrumentation?
`WrapMcpServerWithSentry?`

packages/sidecar/src/utils/getBuffer.ts
```diff
@@ -1,21 +1,8 @@
...
-  let buffer = contextBasedBuffer.get(ctxId);
-  if (!buffer) {
-    buffer = new MessageBuffer<EventContainer>();
-    contextBasedBuffer.set(ctxId, buffer);
-  }
-
-  return buffer;
+export function getBuffer() {
+  return GLOBAL_BUFFER;
 }
```
betegon, line 8: i don't really have context on these buffer changes but looks like we weren't using the context for anything?

packages/sidecar/src/main.ts
```diff
@@ -97,16 +104,15 @@ function startServer(options: StartServerOptions): ServerType {
     }
   }
 
-  function handleServerListen(port: number, basePath?: string): void {
-    logger.info(`Sidecar listening on ${port}`);
-    if (basePath) {
-      logSpotlightUrl(port);
-    }
+  if (options.stdioMCP) {
```
betegon, line 157: if you're destructuring `port` and `basePath` up there, shouldn't we do the same for `stdioMCP`?

packages/sidecar/src/main.ts
```diff
@@ -115,18 +121,30 @@ export function setupSidecar({
...
-  if (debug || process.env.SPOTLIGHT_DEBUG) {
-    enableDebugLogging(true);
+  if (customLogger) {
+    activateLogger(customLogger);
+  } else if (stdioMCP) {
+    activateLogger({
+      info: noop,
+      warn: noop,
+      error: noop,
+      debug: noop,
```
betegon: not sure if this is super clever or super weird in JS.  I guess I would have done a factory function like you did with `createMCPInstance` and return a `silentLogger`

betegon (approved):
looks good! thought this was going to take more time, BYK delivering 

Some notes on trying the mcp locally:
- i was trying to use the `spotlight` package at first instead of `sidecar`. shouldn't we expose the `--stdio-mcp` argument by adding it to `packages/spotlight/bin/run.js` and passing it to `setupSidecar`.
- We should do dynamically change the sidecar port. If it's already running and you run another server, it says that sidecar is already running. this would cause trouble if you want to connect to the mcp and alread have the electron app running. same happened to me when I was trying it on the mcp inspector and then went to cursor (without closing the inspector connection).

packages/sidecar/src/mcp/mcp.ts
BYK: Yes @MathurAditya724 shouldn't we track our stuff? 😀

packages/sidecar/src/utils/getBuffer.ts
BYK: Yeah, checked with @MathurAditya724. This stuff was not used anyway.

packages/sidecar/src/main.ts
BYK: Not really as it's only used once below. I'd rather keep it close to where it's used. LMK if this is too confusing.

packages/sidecar/src/main.ts
```diff
@@ -115,18 +121,30 @@ export function setupSidecar({
...
   }
   let sidecarPort = DEFAULT_PORT;
 
-  if (customLogger) {
-    activateLogger(customLogger);
+  const debugLogging = debug || Boolean(process.env.SPOTLIGHT_DEBUG);
+  if (stdioMCP && !customLogger && debugLogging) {
+    logger.error("STDIO transport for MCP cannot be used with debug as they would interfere.");
+    process.exit(1);
+  } else if (debugLogging) {
```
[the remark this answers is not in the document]
BYK: This is correct but intentional.

packages/sidecar/src/main.ts
BYK: I really dislike the factory pattern but I think I can sidestep these issues by just moving all logging to stderr.

BYK:
>i was trying to use the spotlight package at first instead of sidecar. shouldn't we expose the --stdio-mcp argument by adding it to packages/spotlight/bin/run.js and passing it to setupSidecar.

Yes, that was actually my next step once we validated that this thing works. Will add it to this patch.

>We should do dynamically change the sidecar port. If it's already running and you run another server, it says that sidecar is already running. this would cause trouble if you want to connect to the mcp and alread have the electron app running. same happened to me when I was trying it on the mcp inspector and then went to cursor (without closing the inspector connection).

This is a great point, thanks for sharing! We cannot create multiple sidecar instances listening on random ports tho as we need to tell the SDKs which sidecar to talk to. Instead, when we detect an existing instance, I'll make the new one a proxy for the other one. Since this is gonna be a bit tricky, I'll do that in a follow up PR tho.

betegon (approved):
_thought it looked good from afar, far better up close_. [reference](https://www.facebook.com/reel/1528895850995729)

PR #935, opened 2025-09-18
feat(sidecar): Allow multiple stdio MCP servers to run

BYK:
When we detect Spotlight running on the port we try to connect, instead of quitting we just open a STDIO proxy to that MCP server when `--stdio-mcp` flag is passed.


packages/sidecar/package.json
```diff
@@ -44,13 +44,14 @@
   },
   "dependencies": {
     "@hono/mcp": "^0.1.1",
-    "@hono/node-server": "^1.18.0",
+    "@hono/node-server": "^1.19.3",
```
BYK, line 47: This upgrade is crucial as we need a bugfix introduced in 1.18.2: https://github.com/honojs/node-server/pull/271

packages/sidecar/src/main.ts
```diff
@@ -199,6 +200,13 @@ export function setupSidecar({
       if (hasSpotlightUI) {
         logSpotlightUrl(sidecarPort);
       }
+      if (stdioMCP) {
+        logger.info("Connecting to existing MCP instance with stdio proxy...");
+        await startStdioServer({
+          serverType: ProxyServerType.HTTPStream,
+          url: `http://localhost:${sidecarPort}/mcp`,
+        });
+      }
```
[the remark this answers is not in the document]
BYK: Will add this in a follow up.

packages/sidecar/src/main.ts
```diff
@@ -199,6 +200,13 @@ export function setupSidecar({
...
     } else if (!serverInstance) {
       serverInstance = startServer({ port: sidecarPort, basePath, filesToServe, incomingPayload, stdioMCP });
```
betegon, line 211: somewhat related: how's that we didn't need async for this but it's needed for `startStdioServer`?

betegon (approved):
I thought this was going to be kind of a big PR, what a surprise. This is super easy to maintain :)

let's merge this thing!

packages/sidecar/src/main.ts
BYK: It kind of did and we cheated to not make a breaking change. This actually is a great point, I'll change the public API to make it async.

PR #936, opened 2025-09-19
breaking(sidecar): Change setupSidecar API to be async

BYK:


packages/sidecar/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "@spotlightjs/tsconfig/tsconfig.json",
   "compilerOptions": {
-    "lib": ["ES2023"],
+    "lib": ["ES2024"],
```
betegon, line 4: is this change because of the `Promise.withResolvers`?

betegon (approved):
great follow up :)

packages/sidecar/tsconfig.json
BYK: Yup. And removind Node 20 is for the same reason.

PR #937, opened 2025-09-19
breaking: Remove all vite plugins

BYK:


betegon (approved):
thank you for taking care of this

PR #951, opened 2025-09-25
fix: corrected the clear api

MathurAditya724:


packages/overlay/src/telemetry/store/slices/sharedSlice.ts
```diff
@@ -61,6 +61,8 @@ export const createSharedSlice: StateCreator<SentryStore, [], [], SharedSliceAct
       tracesById: new Map(),
       sdks: new Map(),
       profilesByTraceId: new Map(),
+      logsById: new Map(),
```
BYK, line 64: These don't seem to be used anywhere?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -60,16 +61,25 @@ export class MessageBuffer<T> {
       atReadPos += 1;
     }
 
-    setTimeout(() => this.stream(readerId, atReadPos), 500);
+    const timeoutKey = setTimeout(() => this.stream(readerId, atReadPos), 500);
```
BYK: We need a comment here explaining why we are using a timeout before calling stream

packages/overlay/src/telemetry/store/slices/sharedSlice.ts
MathurAditya724: These are for clearing the logs when the clear button is pressed. These store the logs which I am resetting here when the "clearEvent" is fired.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -60,16 +63,22 @@ export class MessageBuffer<T> {
       atReadPos += 1;
     }
 
-    setTimeout(() => this.stream(readerId, atReadPos), 500);
+    this.subscribersReadPositions.set(readerId, atReadPos);
+
+    // Processing the next batch of messages after a delay to avoid blocking the main thread
+    setTimeout(() => this.stream(readerId), 500);
```
BYK: Why not use setImmediate instead then?

packages/overlay/src/telemetry/store/slices/sharedSlice.ts
BYK: Oh, lol yes got it now.

BYK (commented):
Gonna see if I can use the same map for reader positions, otherwise looks pretty good. Love the new approach

PR #955, opened 2025-09-26
fix: changed the stream logic

MathurAditya724:


packages/sidecar/src/messageBuffer.ts
```diff
@@ -30,16 +30,36 @@ export class MessageBuffer<T> {
       if (atItem[0] > minTime) break;
       this.head += 1;
     }
+
+    for (const readerId of this.readers.keys()) {
+      const readerInfo = this.readers.get(readerId);
+      if (!readerInfo) continue;
```
BYK: This should never happen? Also, we should be using [`Map.prototype.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries):

```ts
for (const [readerId, readerInfo] of this.readers.entries()) {
```

packages/sidecar/src/messageBuffer.ts
```diff
@@ -30,16 +30,36 @@ export class MessageBuffer<T> {
...
+    }
   }
 
   subscribe(callback: (item: T) => void): string {
     const readerId = uuidv7();
     this.readers.set(readerId, { pos: this.head, callback });
-    setImmediate(() => this.stream(readerId));
+
+    const readerInfo = this.readers.get(readerId)!;
+    readerInfo.tid = setImmediate(() => this.stream(readerId));
```
BYK: Vibe coded or just a blunder? Why not:

```ts
const readerInfo = {
  callback,
  pos: this.head,
  tid: setImmediate(() => this.stream(readerId)),
};
this.readers.set(readerId, readerInfo);
```

BYK (changes requested):
Awesome, really like this new approach!

Just a few touch ups and I think we're ready to go 🚀

BYK (approved):

Issue #967, opened 2025-09-30
get_local_errors not showing errors


betegon:
Pasting our convo here:

@sergical said:

>also get_local_errors says no errors detected in spotlight but you can see from the screenshot above (and the log, with the deliberate placement) that it worked fine?

<img width="2057" height="707" alt="Image" src="https://github.com/user-attachments/assets/473a0c0e-facb-49f8-bfef-304125b91d59" />

@BYK 
>It might be due to those ANSI escape characters and we'd ideally strip them at least. (we can look into supporting them in the future)

betegon:
@MathurAditya724 just to be sure, are you going to add this to #968 ? just to know if I should take it

PR #968, opened 2025-09-30
feat: added more filters in the MCP tools

MathurAditya724:
Possible fix for #967 

> [!NOTE]
> Unifies tool input with duration/filename/pagination, applies pagination to outputs, extends MessageBuffer with filename filtering and default duration, and adjusts trace/event reading behavior.
> 
> - **MCP Tools**:
>   - **Unified input**: Replace `timeWindow` with shared `inputSchema` (`duration`, `filename`, `pagination`).
>   - **Filtering/Pagination**: Use buffer `read(filters)` and paginate `content` in `get_local_errors`, `get_local_logs`, and `get_local_traces`.
>   - **Traces**: Paginate trace summaries before appending guidance.
>   - **Trace Details**: `get_events_for_trace` now reads all envelopes (no fixed 600s window).
> - **MessageBuffer**:
>   - **Read defaults**: `read()` defaults to `duration: 60`.
>   - **New filter**: Add `filename` filter matching stacktrace frame filenames; export `ReadFilter` from `InputSchema` (omit `pagination`).
>   - **Filtering logic**: Refactor to apply registered filter handlers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c717d19ef08c286cb83d1b7efd4a28d193857d84. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/sidecar/src/mcp/mcp.ts
+++ b/packages/sidecar/src/mcp/mcp.ts
@@ -20,7 +20,7 @@ const inputSchema = {
       "Look back this many seconds for errors. Use 300+ for broader investigation. Use 30 for recent errors only. For debugging, use 10. For most cases, use 60.",
     ),
   search: z.string().optional().describe("Search envelopes by content. (e.g. filename, error message, etc.)"),
-  filename: z.string().optional().describe("Search envelopes by filename."),
+  filename: z.string().optional().describe("Search events by filename."),
   pagination: z
     .object({
       limit: z.number().describe("Get the last n events."),
--- a/packages/sidecar/src/mcp/mcp.ts
+++ b/packages/sidecar/src/mcp/mcp.ts
@@ -22,29 +22,36 @@ const inputSchema = {
         ),
     }),
     z.object({
-      filename: z.string().describe("Search events by filename."),
+      filename: z
+        .string()
+        .describe(
+          "Filename to search for in the stack trace. Useful when you want to find errors related to a specific file.",
+        ),
     }),
+    /**
+     * TODO: Need to check, if this approach is better then a cursor based approach,
+     * where the user can pass `events since [event_id]` as "last N events" is a
+     * moving target as events keep coming in.
+     *
+     * https://github.com/getsentry/spotlight/pull/968#discussion_r2391587907
+     */
+    z
+      .object({
+        limit: z.number().describe("The number of events to get"),
+        offset: z
+          .number()
+          .default(0)
+          .describe("The position to start from. This is useful when you want to leave out the first n events"),
+      })
+      .optional()
+      .describe("Get the latest n events, useful when you are unable to get data from the timeWindow filter"),
   ]),
-  /**
-   * TODO: Need to check, if this approach is better then a cursor based approach,
-   * where the user can pass `events since [event_id]` as "last N events" is a
-   * moving target as events keep coming in.
-   *
-   * https://github.com/getsentry/spotlight/pull/968#discussion_r2391587907
-   */
-  pagination: z
-    .object({
-      limit: z.number().describe("Get the last n events."),
-      offset: z.number().default(0).describe("Get events starting from the nth position."),
-    })
-    .optional()
-    .describe("Paginate the results. Don't pass this if you want all results."),
 };
 
-export type InputSchema = { [K in keyof typeof inputSchema]?: z.infer<(typeof inputSchema)[K]> };
+export type InputSchema = { [K in keyof typeof inputSchema]: z.infer<(typeof inputSchema)[K]> };
 
-function applyPagination<T>(envelopes: T[], pagination: InputSchema["pagination"]) {
-  if (pagination == null) {
+function applyPagination<T>(envelopes: T[], pagination: InputSchema["filters"]) {
+  if (pagination == null || !("limit" in pagination)) {
     return envelopes;
   }
 
@@ -120,8 +127,7 @@ User: "App crashes after deployment"
       inputSchema,
     },
     async args => {
-      const { pagination, filters } = args;
-      const envelopes = getBuffer().read(filters);
+      const envelopes = getBuffer().read(args.filters);
 
       if (envelopes.length === 0) {
         return NO_ERRORS_CONTENT;
@@ -150,7 +156,7 @@ User: "App crashes after deployment"
       }
 
       return {
-        content: applyPagination(content, pagination),
+        content: applyPagination(content, args.filters),
       };
     },
   );
@@ -227,8 +233,7 @@ User: "I added logging to track user actions"
       inputSchema,
     },
     async args => {
-      const { pagination, filters } = args;
-      const envelopes = getBuffer().read(filters);
+      const envelopes = getBuffer().read(args.filters);
 
       if (envelopes.length === 0) {
         return NO_LOGS_CONTENT;
@@ -257,7 +262,7 @@ User: "I added logging to track user actions"
       }
 
       return {
-        content: applyPagination(content, pagination),
+        content: applyPagination(content, args.filters),
       };
     },
   );
@@ -291,8 +296,7 @@ After identifying a trace of interest, use \`get_events_for_trace\` with the tra
       inputSchema,
     },
     async args => {
-      const { pagination, filters } = args;
-      const envelopes = getBuffer().read(filters);
+      const envelopes = getBuffer().read(args.filters);
 
       if (envelopes.length === 0) {
         return {
@@ -338,7 +342,7 @@ After identifying a trace of interest, use \`get_events_for_trace\` with the tra
         });
       }
 
-      content = applyPagination(content, pagination);
+      content = applyPagination(content, args.filters);
 
       content.push({
         type: "text",
@@ -385,7 +389,7 @@ get_events_for_trace(traceId: "71a8c5e41ae1044dee67f50a07538fe7")  // Using full
     },
     async args => {
       // Getting all the envelopes
-      const envelopes = getBuffer().read({});
+      const envelopes = getBuffer().read({ all: true });
       const traces = extractTracesFromEnvelopes(envelopes);
 
       // Find trace by full ID or partial ID match
--- a/packages/sidecar/src/mcp/mcp.ts
+++ b/packages/sidecar/src/mcp/mcp.ts
@@ -13,43 +13,50 @@ import type { ReadFilter } from "~/messageBuffer.js";
 import { getBuffer } from "~/utils/index.js";
 import { NO_ERRORS_CONTENT, NO_LOGS_CONTENT } from "./constants.js";
 
-export const filterSchema = z.union([
-  z.object({
-    duration: z
-      .number()
-      .describe(
-        "Look back this many seconds for errors. Use 300+ for broader investigation. Use 30 for recent errors only. For debugging, use 10. For most cases, use 60.",
-      ),
-  }),
-  z
-    .object({
-      limit: z.number().describe("Get the last n events."),
-      offset: z.number().default(0).describe("Get events starting from the nth position."),
-    })
-    .describe(
-      "Get events based on a limit and offset. Eg. get the last 20 events or get events starting from the 10th position.",
-    ),
-  z.object({
-    filename: z.string().describe("Search by filename."),
-  }),
-]);
+const filterSchema = z
+  .union([
+    z.object({
+      duration: z
+        .number()
+        .describe(
+          "Look back this many seconds for errors. Use 300+ for broader investigation. Use 30 for recent errors only. For debugging, use 10. For most cases, use 60.",
+        ),
+    }),
+    z.object({
+      filename: z.string().describe("Search by filename."),
+    }),
+  ])
+  .optional();
 
-function getFilterParams(args: z.infer<typeof filterSchema>) {
+const paginationSchema = z
+  .object({
+    limit: z.number().describe("Get the last n events."),
+    offset: z.number().default(0).describe("Get events starting from the nth position."),
+  })
+  .optional();
+
+function getFilterParams(args?: z.infer<typeof filterSchema>) {
   const filter: ReadFilter = {};
 
-  if ("duration" in args) {
-    filter.duration = args.duration;
-  }
+  if (args != null) {
+    if ("duration" in args) {
+      filter.duration = args.duration;
+    }
 
-  if ("limit" in args) {
-    filter.pagination = { limit: args.limit, offset: args.offset };
-  }
-
-  if ("filename" in args) {
-    filter.filename = args.filename;
+    if ("filename" in args) {
+      filter.filename = args.filename;
+    }
   }
 
   return filter;
+}
+
+function applyPagination<T>(envelopes: T[], pagination?: z.infer<typeof paginationSchema>) {
+  if (pagination == null) {
+    return envelopes;
+  }
+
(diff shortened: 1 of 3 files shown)


packages/sidecar/src/mcp/mcp.ts
```diff
@@ -12,6 +12,34 @@ import {
 import { getBuffer } from "~/utils/index.js";
 import { NO_ERRORS_CONTENT, NO_LOGS_CONTENT } from "./constants.js";
 
+const inputSchema = {
+  duration: z
+    .number()
+    .optional()
+    .describe(
+      "Look back this many seconds for errors. Use 300+ for broader investigation. Use 30 for recent errors only. For debugging, use 10. For most cases, use 60.",
```
BYK: Just too many values here. Can we not settle on 2: 300+ for broader investigation and something around 60 for most other stuff?

packages/sidecar/src/mcp/mcp.ts
```diff
@@ -12,6 +12,34 @@ import {
...
+    ),
+  search: z.string().optional().describe("Search envelopes by content. (e.g. filename, error message, etc.)"),
```
BYK: I'd remove the filename from here to make things more efficient and less confusing

packages/sidecar/src/mcp/mcp.ts
```diff
@@ -12,6 +12,34 @@ import {
...
+  filename: z.string().optional().describe("Search envelopes by filename."),
```
BYK:
```suggestion
  filename: z.string().optional().describe("Search events by filename."),
```

(or just errors as not sure if we have filenames for traces/spans/logs)

packages/sidecar/src/mcp/mcp.ts
```diff
@@ -12,6 +12,34 @@ import {
...
+  pagination: z
+    .object({
+      limit: z.number().describe("Get the last n events."),
+      offset: z.number().default(0).describe("Get events starting from the nth position."),
+    })
+    .optional()
+    .describe("Paginate the results. Don't pass this if you want all results."),
+};
```
BYK: This logic feels a bit broken, at least with the name `pagination` as I'd expect an opaque cursor value or something like `events since event id X` as "last N events" is a moving target as events keep coming in.

packages/sidecar/src/mcp/mcp.ts
```diff
@@ -78,17 +106,10 @@ User: "App crashes after deployment"
...
-          .default(60)
-          .describe(
-            "Look back this many seconds for errors. Use 300+ for broader investigation. Use 30 for recent errors only. For debugging, use 10. For most cases, use 60.",
-          ),
-      },
+      inputSchema,
     },
     async args => {
-      const envelopes = getBuffer().read({ duration: args.timeWindow });
+      const envelopes = getBuffer().read(args);
```
BYK: See Cursor's comment above about passing irrelevant stuff here?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -112,25 +113,29 @@ export class MessageBuffer<T> {
...
 
-  read(filters: ReadFilter): T[] {
+  read(filters?: ReadFilter): T[] {
     const result: T[] = [];
     const start = this.head;
     const end = this.writePos;
 
+    const filterHandlers = Object.keys(filters ?? {})
+      .map(key => this.filterHandlers[key])
+      .filter(Boolean);
```
BYK: You may wanna replace this with a single `for..of` loop as that would be quite a bit more efficient? Important because this is probably a hot path, otherwise your approach is fine and declarative

packages/sidecar/src/messageBuffer.ts
```diff
@@ -112,25 +113,29 @@ export class MessageBuffer<T> {
...
+
     for (let i = end - 1; i >= start; i--) {
       const item = this.items[i % this.size];
 
-      if (item !== undefined) {
+      if (item === undefined) continue;
+
+      if (filters != null) {
```
BYK: Very confused here. Why are we handling `null` and `undefined` differently? Also `filters != null` actually covers `item !== undefined` too?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -112,25 +113,29 @@ export class MessageBuffer<T> {
...
-        const isValid = Object.keys(filters).every(key => this.filterHandlers[key](item, filters));
+        const isValid = filterHandlers.every(handler => handler(item, filters));
 
         // Check if the item passes all filters
-        if (!isValid) {
-          continue;
-        }
-
-        result.push(item[1]);
+        if (!isValid) continue;
```
BYK: Why not merge these?
```suggestion
// Check if the item passes all filters
        if (!filterHandlers.every(handler => handler(item, filters))) continue;
```

packages/sidecar/src/messageBuffer.ts
```diff
@@ -153,11 +158,40 @@ export class MessageBuffer<T> {
...
       return data.event[0].__spotlight_envelope_id === value.envelopeId;
     },
+    search: (item, value) => {
+      if (value.search == null) {
+        return true;
+      }
+
+      const data = (item[1] as EventContainer).getParsedEnvelope();
+
+      return JSON.stringify(data.event).includes(value.search);
```
BYK: Very very very inefficient and also probably matches too much stuff that wouldn't produce useful output. I vote against this.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -153,11 +158,40 @@ export class MessageBuffer<T> {
...
+    },
+    filename: (item, value) => {
+      if (value.filename == null) {
+        return true;
+      }
+
+      const filename = value.filename;
```
BYK:
```suggestion
      const { filename } = value;
      if (filename == null) {
        return true;
      }


```

packages/sidecar/src/messageBuffer.ts
```diff
@@ -153,11 +158,40 @@ export class MessageBuffer<T> {
...
+      const contents = (item[1] as EventContainer).getParsedEnvelope();
+
+      for (const [, payload] of contents.event[1]) {
+        if (typeof payload === "object" && "exception" in payload && payload.exception) {
+          const found = payload.exception.values?.some(val =>
+            val.stacktrace?.frames?.some(frame => frame.filename?.match(new RegExp(filename, "i"))),
```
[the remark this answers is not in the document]
BYK: 100% this.

Also compile the regexp only once. Also why are we using a regex in the first place? Can we not just do `endsWith()` or something?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -153,11 +158,40 @@ export class MessageBuffer<T> {
...
+          );
+
+          if (found) {
+            return true;
+          }
```
BYK: Instead of this very annoying logic, you can extend your `.some()` usage which automatically short-circuts. You probably need another outer loop such as:

```ts
return contents.event[1].some(...);
```

Also please see how expensive this is with multiple nested loops with a bunch of conditionals. We should definitely optimize this like checking if `val.stacktrace` exists and if it doesn't just bail early. We may also consider building a filemap index as we get errors to very quickly get the events associated with a file.

BYK (changes requested):
Good start, a bunch of concerns to discuss and/or work over.

packages/sidecar/src/messageBuffer.ts
MathurAditya724: Removing this

packages/sidecar/src/messageBuffer.ts
MathurAditya724: I have changed the implementation. For a filemap, I will create another PR as I need to think about a good implementation for that. I have some ideas but nothing good right now.

packages/sidecar/src/messageBuffer.ts
MathurAditya724: This implementation is changed because of the duration filter.

packages/sidecar/src/mcp/mcp.ts
MathurAditya724: My workflow was, if there is a single error I can just ask, check the last error or last 5 errors. If there is a stream of errors, I close the app and then ask the agent.

As the data is already categoriesed then errors won't be afftected if logs are coming. IMHO, if someone goes in the UI for the error, they can just copy the prompt for that using the AI button.

packages/sidecar/src/mcp/mcp.ts
```diff
@@ -12,6 +12,33 @@ import {
 import { getBuffer } from "~/utils/index.js";
 import { NO_ERRORS_CONTENT, NO_LOGS_CONTENT } from "./constants.js";
 
+const inputSchema = {
+  duration: z
+    .number()
+    .optional()
+    .describe(
+      "Look back this many seconds for errors. Use 300+ for broader investigation and something around 60 for most other stuff",
```
BYK: for everything else

packages/sidecar/src/mcp/mcp.ts
BYK: Not convinced but I'm fine to let this pass until it becomes a real issue. Would you mind noting my concern as a comment for future reference?

packages/sidecar/src/mcp/mcp.ts
```diff
@@ -12,6 +12,33 @@ import {
 import { getBuffer } from "~/utils/index.js";
 import { NO_ERRORS_CONTENT, NO_LOGS_CONTENT } from "./constants.js";
 
+const inputSchema = {
+  duration: z
```
BYK: Duration is not a good name as in the future we may want to search traces by their total duration. What was wrong with timeWindow? If you don't like that, maybe we can do "recency"?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -112,25 +113,27 @@ export class MessageBuffer<T> {
...
   }
 
-  read(filters: ReadFilter): T[] {
+  read(filters: ReadFilter = { duration: 60 }): T[] {
     const result: T[] = [];
     const start = this.head;
     const end = this.writePos;
 
+    const filterHandlers = [];
+    for (const key in Object.keys(filters)) {
```
BYK: Don't use for..in

Pretty sure that's not what you want

packages/sidecar/src/messageBuffer.ts
```diff
@@ -153,11 +156,26 @@ export class MessageBuffer<T> {
 
       return data.event[0].__spotlight_envelope_id === value.envelopeId;
     },
+    filename: (item, value) => {
```
BYK, line 159: Have you read my comments around this? If yes what do you think?

BYK (changes requested):

packages/sidecar/src/mcp/mcp.ts
MathurAditya724: I will create a issue from this

packages/sidecar/src/messageBuffer.ts
MathurAditya724: https://github.com/getsentry/spotlight/pull/968#discussion_r2392059899

packages/sidecar/src/mcp/mcp.ts
MathurAditya724: timeWindow is good

packages/sidecar/src/messageBuffer.ts
MathurAditya724: Ahh my bad

packages/sidecar/src/mcp/mcp.ts
```diff
@@ -361,7 +379,8 @@ get_events_for_trace(traceId: "71a8c5e41ae1044dee67f50a07538fe7")  // Using full
       },
     },
     async args => {
-      const envelopes = getBuffer().read({ duration: 600 }); // Look back further for trace details
+      // Getting all the envelopes
+      const envelopes = getBuffer().read({});
```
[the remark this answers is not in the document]
MathurAditya724: Based upon the comments, it seems that we wanna look at all of the buffer to find the trance

packages/sidecar/src/messageBuffer.ts
```diff
@@ -112,37 +113,39 @@ export class MessageBuffer<T> {
     this.head = this.writePos;
   }
 
-  read(filters: ReadFilter): T[] {
+  read(filters: ReadFilter = { timeWindow: 60 }): T[] {
```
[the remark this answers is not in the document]
MathurAditya724: Changed the filtering approach

packages/sidecar/src/mcp/mcp.ts
```diff
@@ -12,6 +12,45 @@ import {
...
+      timeWindow: z
+        .number()
+        .describe(
+          "Look back this many seconds for errors. Use 300+ for broader investigation and something around 60 for everything else",
+        ),
+    }),
+    z.object({
+      filename: z.string().describe("Search events by filename."),
+    }),
+  ]),
```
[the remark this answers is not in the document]
MathurAditya724: Yes, this is intended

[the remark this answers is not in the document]
MathurAditya724: Yes, this is intended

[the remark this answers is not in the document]
BYK: I think we should revisit this decision?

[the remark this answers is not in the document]
BYK: Same as above so resolving this.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -157,8 +157,7 @@ export class MessageBuffer<T> {
       return data.event[0].__spotlight_envelope_id === value.envelopeId;
     },
     filename: (item, value) => {
-      const { filename } = value;
-      if (filename == null) {
+      if (!("filename" in value)) {
```
BYK, line 160: I feel like we need a lint rule banning the `in` keyword usage. This is not what you want.
```suggestion
      if (!value.filename) {
```

Alternatively `if (!value.hasOwnProperty('filename'))`.

Not sure what exactly you are trying to do here tho. Can you share so we can converge on a solution together?

MathurAditya724: So, now we have 3 ways to filter data. 
1. Time Window
2. Filename
3. Limit and Offset

And thanks to the validation we get the types on this and by this just help with the types. The alternate way doesn't work with the types.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -112,52 +113,71 @@ export class MessageBuffer<T> {
...
-        // Apply all filters
-        const isValid = Object.keys(filters).every(key => this.filterHandlers[key](item, filters));
+      if (item == null) continue;
 
-        // Check if the item passes all filters
-        if (!isValid) {
-          continue;
-        }
+      // Check if the item passes all filters
+      if (!filterHandlers.every(handler => handler(item, filters))) continue;
```
BYK:
```suggestion
      if (filterHandlers.every(handler => handler(item, filters))) {
        result.push(item[1]);
      }
```

BYK (approved):
As long as this works, I'm fine. Not super happy with the `in` usage and only one filter can be active at a time but we can address these in a follow up.

Let's sheep this 🐑

PR #977, opened 2025-10-01
fix: removed the shadow dom

### LINKED ISSUE #974: Spotlight root not found errors
We got reports of "Spotlight root not found" when looking at the web UI.

We need to fix this. Probably related to the move away from the embedded model _and_ something with the shadow DOM usage.

MathurAditya724:
Closes #974


packages/overlay/index.html
```diff
@@ -1,27 +1,28 @@
...
-  <script type="module">
-    import * as Spotlight from './src/index.tsx';
-    Spotlight.init({ fullPage: true, debug: true });
-  </script>
-</body>
-
+  <body>
+    <script type="module">
+      import * as Spotlight from "./src/index.tsx";
+      Spotlight.init({ debug: true });
```
MathurAditya724, line 25: The only thing I changed in this file is this line, as that prop is no longer needed. Rest is just formatting changes happened automatically.

BYK: We should probably not default `debug` to `true`? Follow up fix.

BYK (approved):
Okay, did not expect this to be so easy 😀

packages/overlay/index.html
MathurAditya724: I believe this only works when we are doing `pnpm run dev`, in our development env. This is not effecting the production env

BYK: That's right.

Issue #975, opened 2025-09-30
Random Sentry error reports from Spotlight


BYK:
We should deprecate the use of `Spotlight.init()` as some people still have this in their apps, triggering our own Sentry error reporting for other people's apps.

Quick fix: replace `Spotlight.init()` with `Spotlight._init()` for now, update it across the app and make `Spotlight.init()` do nothing but print a warning to the console saying this is now deprecated and should not be used.

MathurAditya724:
Created a quick fix PR for this - https://github.com/getsentry/spotlight/pull/978

BYK:
Yeah let's close for now and revisit after the next launch.

PR #978, opened 2025-10-01
fix: simple override for init

MathurAditya724:
Quick Fix for #975

### DIFF
--- a/packages/overlay/src/index.tsx
+++ b/packages/overlay/src/index.tsx
@@ -8,10 +8,10 @@ import { on, trigger } from "./lib/eventTarget";
 import initSentry from "./lib/instrumentation";
 import { activateLogger, log } from "./lib/logger";
 import { removeURLSuffix } from "./lib/removeURLSuffix";
+import { getDataFromServerTiming } from "./lib/serverTimingMeta";
 import { SpotlightContextProvider } from "./lib/useSpotlightContext";
 import { React, ReactDOM } from "./react-instance";
 import type { WindowWithSpotlight } from "./types";
-import { getDataFromServerTiming } from "./lib/serverTimingMeta";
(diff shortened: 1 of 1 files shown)


packages/overlay/src/index.tsx
```diff
@@ -36,7 +36,14 @@ export type SpotlightInitOptions = {
   debug?: boolean;
 };
 
-export async function init(initOptions: SpotlightInitOptions = {}) {
+/**
+ * @deprecated This function is deprecated and should not be used.
+ */
+export async function init(_initOptions: SpotlightInitOptions = {}) {
+  console.warn("init is deprecated, and should not be used");
```
BYK:
```suggestion
  console.warn("Spotlight has moved away from the embedded UI overlay model so you should not be using this module at all.");
  console.info("See the new roadmap for Spotlight here: https://github.com/getsentry/spotlight/issues/891");
  console.info("Our standalone UI along with the Electron app are going stronger than ever so please try those out and share any feedback you have with us.");
  console.info("Thanks for using Spotlight!");
```

BYK (approved):
Note the recommended change for the console messaging before merging please.

packages/spotlight/src/index.html
```diff
@@ -19,7 +19,7 @@
   </head>
   <body>
     <script type="module">
-      import * as Spotlight from '@spotlightjs/overlay';
+      import * as Spotlight from "@spotlightjs/overlay";
       if (!window.__spotlight) {
```
BYK, line 23: We should probably remove all this stuff now. In a follow up?

MathurAditya724: This wasn't me but biome for some reason 😆

PR #979, opened 2025-10-01
fix: removed the tooltip from the filter dropdown

MathurAditya724:


BYK (approved):
Nice! If we are not using this module, we should probably remove it from our dependencies too?

PR #990, opened 2025-10-08
fix: Added reconnect logic for sidecar in the electron app

### LINKED ISSUE #956: Electron app should be able to run when sidecar is already running
If there's already a sidecar instance running on the default port, Electron app should just connect to it and spawn its own server in the case the existing one dies for some reason.

This is in line with how we handle proxying with the `--stdio-mcp` option.

MathurAditya724:
Closes #956

### DIFF
--- a/.changeset/busy-shoes-unite.md
+++ b/.changeset/busy-shoes-unite.md
@@ -1,5 +1,5 @@
 ---
-"@spotlightjs/electron": patch
+"@spotlightjs/electron": minor
 ---
 
 Added restart logic for the sidecar
--- a/packages/electron/src/electron/main/index.ts
+++ b/packages/electron/src/electron/main/index.ts
@@ -604,15 +604,12 @@ async function makeSureSidecarIsRunning() {
 
   async function handler() {
     try {
-      /**
-       * Checking if the sidecar is running
-       * And if not, starting it up
-       */
       await setupSidecar({
         port: DEFAULT_PORT,
         incomingPayload: storeIncomingPayload,
         isStandalone: true,
       });
+
       retries = 0;
     } catch (error) {
       console.error(error);


packages/electron/src/electron/main/index.ts
```diff
@@ -599,3 +593,55 @@ Promise.all([
...
+    if (subscriber) {
+      clearTimeout(subscriber);
+      subscriber = null;
+    }
+
+    if (retries >= MAX_RETRIES) {
+      return;
+    }
+
+    subscriber = setTimeout(handler, RECHECK_DELAY + retries * RETRY_DELAY_INCREMENT);
```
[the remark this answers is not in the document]
MathurAditya724: I have updated the port to DEFAULT_PORT

.changeset/busy-shoes-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spotlightjs/electron": patch
```
betegon: i'd say this should be a minor. From semver:

> MINOR version when you add functionality in a backward compatible manner

packages/electron/src/electron/main/index.ts
```diff
@@ -599,3 +593,55 @@ Promise.all([
...
+      });
+      retries = 0;
+    } catch (error) {
+      console.error(error);
+      retries++;
+
+      if (retries >= MAX_RETRIES) {
+        Sentry.captureException(error);
+
+        // Notifying the user that the sidecar is not running
```
betegon:
```suggestion

```

These kind of comments are a bit redundant, but LLMs love to put them. We should try to use comments to document behaviour. 

an example for this (although also redundant) that could document behaviour could be: 

> // If the sidecar isn't running, we alert the user because some features depend on it

packages/electron/src/electron/main/index.ts
```diff
@@ -599,3 +593,55 @@ Promise.all([
...
+
+async function makeSureSidecarIsRunning() {
+  let retries = 0;
+  let subscriber: NodeJS.Timeout | null = null;
+
+  async function handler() {
+    try {
+      /**
+       * Checking if the sidecar is running
+       * And if not, starting it up
```
betegon: same as my comment below. if we want to add comments, comments should answer:

- Why did Aid add this? 
- Because the user could have a sidecar from another spotlight service already running

PR #997, opened 2025-10-09
Add a CLI that agents can also use instead of the MCP

### LINKED ISSUE #958: Spotlight CLI
`spotlight logs`
`spotlight errors`
`spotlight traces`
`spotlight logs -f`

These should all work.

BYK:
This CLI introduces commands like `logs` or `errors` which can also be
combined with a `,` or `+` to listen to multiple streams at once.

The first iteration will be a streaming log like `tail -f` but in the next
iteration we will add a `no-follow` mode that accepts some filters like a
`timeWindow` or `limit` as the MCP does.

Closes #958.


packages/sidecar/src/routes/stream/index.ts
```diff
@@ -96,6 +96,11 @@ const router = new Hono<HonoEnv>()
       logIncomingEvent(container);
 
       getBuffer().put(container);
+
+      const onEnvelope = ctx.get("onEnvelope");
+      if (onEnvelope) {
+        onEnvelope(container);
+      }
```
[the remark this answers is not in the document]
betegon: I KNOW. I'M FIXING THIS ALREADY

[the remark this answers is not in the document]
betegon: I had to push a commit so you figured this out. You missed this in your first review ;)

packages/sidecar/server.ts
```diff
@@ -21,9 +44,112 @@ Examples:
...
+      } else {
+        const types = new Set([...eventTypes.flatMap(type => NAME_TO_TYPE_MAPPING[type] || [])]);
+        client.addEventListener(SENTRY_CONTENT_TYPE, event => {
+          const envelope = JSON.parse(event.data);
+          for (const [header] of envelope[1]) {
+            if (header.type && types.has(header.type)) {
+              displayEnvelope(envelope);
+            }
+          }
+        });
```
betegon: when connecting to an existing server we receive a an `Envelope`, while if we're using our own serrve the `onEnvelope` receives `EventContainer` and extracts the envelope from it.

i.e. we can't use the same callback for both :)

packages/sidecar/server.ts
```diff
@@ -21,9 +44,112 @@ Examples:
...
+        if (!parsed) return;
+        for (const [header] of parsed.event[1]) {
+          if (header.type && types.has(header.type)) {
+            displayEnvelope(parsed.event);
+          }
+        }
+      };
+    }
+
+    // try to connect to an already existing server first
```
betegon, line 119: Any suggestion on the best behaviour when connecting to a new server?

Right now, it will throw all the events into the terminal right when you connect to it, instead of starting "blank".

packages/sidecar/server.ts
```diff
@@ -21,9 +44,112 @@ Examples:
...
+    break;
+  case undefined:
+  case "":
+    break;
+  default: {
+    if (args._positionals.length > 1) {
+      console.error("Error: Too many positional arguments.");
+      printHelp();
+    }
+    const eventTypes = cmd.toLowerCase().split(/\s*[,+]\s*/gi);
```
betegon: shouldn't we allow: `errors, logs` with an space in between? felt more natural to me

packages/sidecar/src/routes/stream/index.ts
```diff
@@ -96,6 +96,11 @@ const router = new Hono<HonoEnv>()
       logIncomingEvent(container);
 
       getBuffer().put(container);
+
+      const onEnvelope = ctx.get("onEnvelope");
+      if (onEnvelope) {
+        onEnvelope(container);
```
Shubhdeep12: just a thought @BYK @betegon 

Can we enhance the `onEnvelope` to something like `beforeSend` option of Sentry SDK?

this way 
- Users could control which envelope they wanna drop. and maybe it could resolve issues like - https://github.com/getsentry/sentry-python/issues/4712

Shubhdeep12: My bad, this is not possible until we set up the sidecar in code... But it'll run most of the time directly by electron app or via terminal.. please ignore the above message.😅

packages/sidecar/server.ts
BYK: No we can as I made them use the same parsed data afterwards. I'll have to revert your commits and work from there.

packages/sidecar/server.ts
BYK: that is the intention and will become important when we add "print and exit" behavior instead of the current tailing behavior. Do you think it is not good practice?

packages/sidecar/server.ts
BYK: The CLI convention is usually `<CLITOOL> <command> <subcommand>` hence when you do `spotlight logs errors` or `spotlight errors logs` for instance, the expected interpretation is there to be a command called `logs` under `errors` or vice versa. That's why I opted for `+` or `,`. If it doesn't make sense, we can think more on this tho.

packages/sidecar/server.ts
betegon: > will become when we add "print and exit" behavior
alright, this makes sense.

It was just that I had the sidecar going on for another app when i tried it and flushed all the info from the other app obviously.

packages/sidecar/server.ts
betegon: hmm, ok. Let's go with these and see if we get some feedback

packages/sidecar/server.ts
betegon: Oh so I could have used `getParsedEnvelope().event` on the `onEnvelope`. Ok sorry for the mess

packages/sidecar/src/format/traces.ts
```diff
@@ -123,7 +123,7 @@ function extractTraceEventsFromContainer(container: EventContainer): TraceEvent[
       data: container.getData(),
     });
 
-    const [, items] = parsed.event;
+    const [, items] = parsed!.envelope;
```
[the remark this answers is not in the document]
BYK: This is in a try-catch block specifically added to catch these kinds of issues so I'll pass.

packages/sidecar/src/main.ts
```diff
@@ -61,11 +71,21 @@ export function parseCLIArgs(): { port: number; debug: boolean; stdioMCP: boolea
     process.exit(1);
   }
 
-  return {
-    ...values,
-    stdioMCP: values["stdio-mcp"],
+  if (values["stdio-mcp"]) {
+    positionals.unshift("mcp");
+    console.warn("Warning: --stdio-mcp is deprecated. Please use the positional argument 'mcp' instead.");
```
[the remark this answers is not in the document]
BYK: Intentional

.changeset/petite-humans-drop.md
```diff
@@ -0,0 +1,6 @@
+---
+"@spotlightjs/sidecar": minor
+"@spotlightjs/spotlight": minor
+---
+
+Add CLI -- `spotlight logs+errors` etc are now available
```
BYK:
```suggestion
Add CLI -- `spotlight tail logs errors` etc are now available
```

PR #999, opened 2025-10-09
fix: improved attachments display

MathurAditya724:
- [x] Improved support for images
- [x] Added shiki for code attachments like CSS, HTML and JS
image
image

### DIFF
--- a/packages/overlay/src/telemetry/components/insights/envelopes/CodeViewer.tsx
+++ b/packages/overlay/src/telemetry/components/insights/envelopes/CodeViewer.tsx
@@ -1,21 +1,6 @@
 import { transformerNotationHighlight } from "@shikijs/transformers";
-import { useEffect, useState } from "react";
-import { type Highlighter, bundledLanguages, createHighlighter } from "shiki/bundle-web.mjs";
-
-const THEME = "github-dark";
-
-function useShiki() {
-  const [highlighter, setHighlighter] = useState<Highlighter | null>(null);
-
-  useEffect(() => {
-    createHighlighter({
-      themes: [THEME],
-      langs: Object.keys(bundledLanguages),
-    }).then(setHighlighter);
-  }, []);
-
-  return highlighter;
-}
+import { bundledLanguages } from "shiki/bundle-web.mjs";
+import { SHIKI_DEFAULT_THEME, useShiki } from "~/ShikiProvider";
 
 export function CodeViewer({ code, lang }: { code: string; lang: string }) {
   const highlighter = useShiki();
@@ -31,7 +16,7 @@ export function CodeViewer({ code, lang }: { code: string; lang: string }) {
         dangerouslySetInnerHTML={{
           __html: highlighter.codeToHtml(code, {
             lang,
-            theme: THEME,
+            theme: SHIKI_DEFAULT_THEME,
             transformers: [transformerNotationHighlight()],
           }),
         }}
--- a/packages/overlay/src/App.tsx
+++ b/packages/overlay/src/App.tsx
@@ -1,4 +1,5 @@
 import { Navigate, Route, Routes } from "react-router-dom";
+import { ShikiProvider } from "./ShikiProvider";
 // TODO: we'll lazy load this in case of multiple routes
 import { Telemetry } from "./telemetry";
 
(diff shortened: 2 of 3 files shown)


.changeset/polite-parts-fall.md
```diff
@@ -0,0 +1,6 @@
+---
+"@spotlightjs/overlay": patch
```
BYK:
```suggestion
"@spotlightjs/overlay": minor
```

packages/overlay/src/telemetry/components/insights/envelopes/Attachment.tsx
```diff
@@ -1,22 +1,57 @@
...
-  if (header.content_type === "text/plain") {
-    extension = "txt";
+
+  const decodedAttachmentUrl = useMemo(
+    () =>
+      IMAGE_CONTENT_TYPES.has(header.content_type as string)
+        ? URL.createObjectURL(
+            new Blob(
+              [
+                // @ts-expect-error
```
BYK: Would be great to note the expected error here, and why do we expect it

packages/overlay/src/telemetry/components/insights/envelopes/Attachment.tsx
```diff
@@ -1,22 +1,57 @@
...
 
 export default function Attachment({ header, attachment }: { header: EnvelopeItem[0]; attachment: string }) {
   let content: ReactNode | null = null;
   let extension = "bin";
-  if (header.content_type === "text/plain") {
-    extension = "txt";
+
+  const decodedAttachmentUrl = useMemo(
+    () =>
+      IMAGE_CONTENT_TYPES.has(header.content_type as string)
```
BYK: Why are we limiting downloads to only images?

packages/overlay/src/telemetry/components/insights/envelopes/Attachment.tsx
```diff
@@ -1,22 +1,57 @@
...
+  } else if (JSON_CONTENT_TYPES.has(header.content_type as string)) {
     extension = "json";
     content = <JsonViewer data={JSON.parse(attachment)} />;
+  } else if (header.content_type === "text/css") {
+    extension = "css";
+    content = <CodeViewer code={attachment} lang={extension} />;
+  } else if (header.content_type === "text/html") {
+    extension = "html";
+    content = <CodeViewer code={attachment} lang={extension} />;
+  } else if (header.content_type === "text/javascript") {
```
BYK: This section begs for a map of content type to extensions rather than a bunch of if/else statements

packages/overlay/src/telemetry/components/insights/envelopes/Attachment.tsx
```diff
@@ -1,22 +1,57 @@
...
+    extension = "js";
+    content = <CodeViewer code={attachment} lang={extension} />;
+  } else if (IMAGE_CONTENT_TYPES.has(header.content_type as string)) {
+    extension = "bin";
```
BYK: Why is the extension for images is bin??

packages/overlay/src/telemetry/components/insights/envelopes/CodeViewer.tsx
```diff
@@ -0,0 +1,30 @@
+import { useEffect, useState } from "react";
+import { type Highlighter, createHighlighter } from "shiki";
```
BYK: We should use the same highlighter from error stacktraces. I think we don't have highlighting there so we should add your new component there too

packages/overlay/src/telemetry/store/slices/eventsSlice.ts
```diff
@@ -66,6 +66,7 @@ export const createEventsSlice: StateCreator<SentryStore, [], [], EventsSliceSta
       }
     }
 
+    const { eventsById } = get();
```
BYK, line 69: What's going on here exactly?

MathurAditya724: This was the reference bug I was talking about in slack. If the envelope is an error we are doing a request which can take some time to complete.

In my testing I was sending multiple errors all together and each took it's time to complete and as the Map pointer was not getting updated when others envelopes where added, it's was always empty for all of them and it always showed a single envelope, which was the last one updating the Map.

packages/overlay/src/telemetry/components/insights/envelopes/Attachment.tsx
MathurAditya724: yeah, idk what I was thinking

PR #1005, opened 2025-10-14
Fix(sidecar): memory leak on filenameCache

betegon:
Adding:
- **a fix for a memory leak we had on `filenameCache`:** When a element from the buffer is removed because of being old, we didn't clear the cache.

### DIFF
--- a/packages/sidecar/src/messageBuffer.ts
+++ b/packages/sidecar/src/messageBuffer.ts
@@ -232,7 +232,9 @@ export class MessageBuffer<T> {
 
       for (const [filename, envelopeIds] of ctx.filenameCache.entries()) {
         if (filename.endsWith(value.filename)) {
-          return envelopeIds.has(String(spotlightEnvelopeId));
+          if (envelopeIds.has(String(spotlightEnvelopeId))) {
+            return true;
+          }
         }
       }
 


packages/sidecar/src/messageBuffer.ts
```diff
@@ -220,9 +242,7 @@ export class MessageBuffer<T> {
 
       for (const [filename, envelopeIds] of ctx.filenameCache.entries()) {
         if (filename.endsWith(value.filename)) {
-          if (envelopeIds.has(String(spotlightEnvelopeId))) {
-            return true;
-          }
+          return envelopeIds.has(String(spotlightEnvelopeId));
         }
```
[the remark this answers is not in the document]
betegon: fixed.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -46,13 +46,25 @@ export class MessageBuffer<T> {
       atItem = this.items[this.head % this.size];
       if (atItem === undefined) break;
       if (atItem[0] > minTime) break;
+
+      const expiredEnvelope = (atItem[1] as EventContainer)?.getParsedEnvelope();
+      if (expiredEnvelope?.event) {
+        const expiredEnvelopeId = String(expiredEnvelope.event[0].__spotlight_envelope_id);
```
BYK, line 52: Why do we need the string cast here?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -46,13 +46,25 @@ export class MessageBuffer<T> {
...
+
+        for (const envelopeIds of this.filenameCache.values()) {
+          if (envelopeIds.has(expiredEnvelopeId)) {
+            envelopeIds.delete(expiredEnvelopeId);
```
BYK, line 56: This is gonna be tricky but ideally we'd also check if the set becomes empty and remove the filename entry from the map completely at the end of the loop.

packages/sidecar/src/messageBuffer.ts
betegon: ok maybe it's a TS skill issue but that `__spotlight_envelope_id` comes from: 
```typescript
export type BaseEnvelopeHeaders = {
    [key: string]: unknown;
```

so if later (line 55) we do `envelopeIds.has(expiredEnvelopeId`) without the cast, we get a type problem because `.has()`  expects a string

BYK: We should probably add this as a known key to the type then. I'll see if this is possible.

PR #1017, opened 2025-10-21
fix(sidecar): Fix type checking and existing errors

betegon:
We lost type checking when we added tsdown. Now it's back baby.


BYK (approved):

.changeset/afraid-geckos-stand.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spotlightjs/sidecar": patch
+---
+
+Add type checking on build
```
BYK:
```suggestion
Fix type checking and existing errors
```

PR #1018, opened 2025-10-21
fix: corrected the read function

### LINKED ISSUE #1011: MCP: Traces expire too quickly from buffer (~30-60s)
## Problem

  Traces expire from Spotlight's in-memory buffer within 30-60 seconds, making MCP-based analysis difficult.

  ## Use Case

  Using Claude Code with MCP (`@spotlightjs/spotlight@latest --stdio-mcp`) for async debugging:

  1. User triggers operation (e.g., multi-service process taking 2+ seconds)
  2. User asks Claude to analyze the trace
  3. Claude calls `get_local_traces` / `get_events_for_trace`
  4. **Trace already expired** - "trace not found"

  This requires tight timing coordination and multiple attempts.

  ## Question

 Is there a way to:
  - Configure trace retention duration?
  - Enable persistent storage?
  - Any recommended workarounds for async MCP analysis?

The MCP integration is great for AI-assisted debugging, but the short retention makes it hard to use effectively for anything beyond real-time analysis.

MathurAditya724:
Closes #1011

### DIFF
--- a/packages/sidecar/src/messageBuffer.ts
+++ b/packages/sidecar/src/messageBuffer.ts
@@ -183,7 +183,7 @@ export class MessageBuffer<T> {
 
   read(filters: ReadFilter = { timeWindow: 60 }): T[] {
     const result: T[] = [];
-    const start = this.writePos - this.size;
+    const start = Math.max(0, this.writePos - this.size);
     const end = this.writePos;
 
     const filterHandlers = [];


packages/sidecar/src/messageBuffer.ts
```diff
@@ -183,7 +183,7 @@ export class MessageBuffer<T> {
 
   read(filters: ReadFilter = { timeWindow: 60 }): T[] {
     const result: T[] = [];
-    const start = this.head;
+    const start = this.writePos - this.size;
```
betegon: great catch!

I believe we should do  `Math.max(this.head, this.WritePos - this.size)` ?

.changeset/tired-candles-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spotlightjs/sidecar": minor
```
BYK, line 2:
```suggestion
"@spotlightjs/sidecar": patch
```

We're fixing a bug, not adding a feature 😀

PR #1020, opened 2025-10-21
feat: Add spotlight run command

### LINKED ISSUE #987: `spotlight run`
Add a `run` command to Spotlight CLI that does the following:

- Select a random port that is not in use for a new sidecar instance
- Inject `$SENTRY_SPOTLIGHT=http://localhost:{port}` into env variables
  - Alternatively or in addition also inject `$SENTRY_DSN=http://spotlight@localhost:{port}/0` for SDKs that don't support `$SENTRY_SPOTLIGHT` env variable but `$SENTRY_DSN`
- Run the given command(s)
- Capture the stdio of the given commands and convert them into special `log` events
- Start streaming all the logs, errors, and traces coming from this new invocation. Combined with the `stdio` capture above, this may require some deduplication if we already capture stdio as part of Sentry SDK

Primary goal:

- Make this work great for JS-based projects like Next.js, Astro, SvelteKit

Secondary goal:

- Make this work great with Docker Compose and integrate into existing Sentry repo

Stretch goals:

- Auto detect the `dev` command for a JS project and invoke that automatically when the user just runs `spotlight run` without a command. Initial idea: look under `package.json` `scripts` section and try `dev`, `watch`, `serve`, and `start` in that order
- Auto-restart when a process dies with some back off?
- Auto detect Docker Compose and inject additional config to make sure `SENTRY_SPOTLIGHT` and `SENTRY_SDK` env vars are propagated to all services
- Auto modify Sentry SDK settings such as trace_sample_rate etc.
- Auto instrument a project if Sentry SDK does not exist there -- can use existing work from the wizard and agent-wizard projects

BYK:
Fixes #987

### DIFF
--- a/packages/sidecar/src/cli.ts
+++ b/packages/sidecar/src/cli.ts
@@ -193,10 +193,8 @@ export async function main(filesToServe?: Record<string, Buffer>) {
         ? SUPPORTED_ENVELOPE_TYPES
         : new Set([...eventTypes.flatMap(type => NAME_TO_TYPE_MAPPING[type] || [])]);
       onEnvelope = envelope => {
-        for (const [header] of envelope[1]) {
-          if (header.type && types.has(header.type)) {
-            displayEnvelope(envelope);
-          }
+        if (envelope[1].some(([header]) => header.type && types.has(header.type))) {
+          displayEnvelope(envelope);
         }
       };
 


packages/sidecar/src/cli.ts
```diff
@@ -89,12 +90,71 @@ switch (cmd) {
...
+    const runCmd = spawn(cmdArgs[0], cmdArgs.slice(1), {
+      cwd: process.cwd(),
+      env: {
+        ...process.env,
+        SENTRY_SPOTLIGHT: `http://localhost:${port}/stream`,
+        // This is not supported in all SDKs but worth adding
+        // for the ones that support it
+        SENTRY_TRACES_SAMPLE_RATE: "1",
+        //  Consider setting SENTRY_DSN to our hack server too?
+      },
```
betegon: we may want to add `shell: true` to allow shell syntax like `spotlight run npm start | grep error`.  Llms love to pipe commands

packages/sidecar/src/cli.ts
```diff
@@ -89,12 +90,71 @@ switch (cmd) {
...
+        // TODO: Check docker-compose.yaml existence
+        // TODO: Check `docker compose --version` existence
+        // TODO: Check `docker-compose --version` existence
+      } catch {
+        // pass
+      }
+    }
+    console.log(cmdArgs);
+    const cmdStr = cmdArgs.join(" ");
+    const runCmd = spawn(cmdArgs[0], cmdArgs.slice(1), {
```
betegon: shouldn't we first start the sidecar and then spawn the user's command so we don't miss any data?

BYK: Yeah, I thought about this too. I was being wishful and assuming we'd already start our server before the process comes up as they are async but your point is valid.

I'll still defer that to a follow up though as it requires more refactor of this CLI file (moving away from `switch` and `case` statements).

packages/sidecar/src/cli.ts
BYK: I'm always vary of `shell` as it is brittle and dangerous (and not portable). That said I'll check how npm/pnpm does these and adjust accordingly.

BYK: I switched to `shell` for npm scripts as I think that's the expectation anyway.

packages/sidecar/src/main.ts
```diff
@@ -10,57 +10,74 @@ import { cors } from "hono/cors";
...
+  const preParse = parseArgs({
+    ...PARSE_ARGS_CONFIG,
     strict: false,
+    tokens: true,
+  });
+  let cmdArgs: string[] | undefined = undefined;
+  const runToken = preParse.tokens.find(token => token.kind === "positional" && token.value === "run");
+  if (runToken) {
+    const cutOff = preParse.tokens.find(token => token.index > runToken.index && token.kind === "positional");
+    cmdArgs = cutOff ? args.splice(cutOff.index) : [];
```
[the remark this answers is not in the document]
BYK: Nope, this is intentional

packages/sidecar/src/cli.ts
```diff
@@ -0,0 +1,233 @@
...
+        logger.error(err);
+        process.exit(1);
+      });
+      runCmd.on("close", code => {
+        if (!code) {
+          logger.error(`${cmdStr} exited, terminating.`);
+        } else {
+          logger.error(`${cmdStr} exited with code ${code}.`);
+        }
+        process.exit(code);
```
betegon, line 164: shouldn't we account for code equalt to 0, that wont be an error.

packages/sidecar/src/cli.ts
```diff
@@ -0,0 +1,233 @@
...
+      });
+      const killRunCmd = () => {
+        runCmd.removeAllListeners();
+        runCmd.kill();
+      };
```
betegon, line 169: This could cause problems on windows as it doesn't kill the child processes.

I may be terribly wrong but it will fail on windows if we run:
- `spotlight run npm run dev`  the npm process gets killed, but not the dev server

packages/sidecar/src/cli.ts
```diff
@@ -0,0 +1,233 @@
...
+        logger.error("Error: No command specified to run and could not infer the command automatically.");
+        process.exit(1);
+      }
+      const cmdStr = cmdArgs.join(" ");
+      logger.info(`Starting command: ${cmdStr}`);
+      // TODO: Run this _after_ the server is started
+      const runCmd = spawn(cmdArgs[0], cmdArgs.slice(1), {
+        cwd: process.cwd(),
+        shell,
+        windowsVerbatimArguments: true,
```
betegon, line 148: didn't know what this was and from what I've read, this can produce weird quoting on windows if we have the shell to true

packages/sidecar/src/cli.ts
```diff
@@ -0,0 +1,233 @@
...
+        }
+      }
+
+      const types = eventTypes.some(type => EVERYTHING_MAGIC_WORDS.has(type))
+        ? SUPPORTED_ENVELOPE_TYPES
+        : new Set([...eventTypes.flatMap(type => NAME_TO_TYPE_MAPPING[type] || [])]);
+      onEnvelope = envelope => {
+        for (const [header] of envelope[1]) {
+          if (header.type && types.has(header.type)) {
+            displayEnvelope(envelope);
```
betegon: shouldn't we add a `break` here?

if we have an envelope with multiple items of the same type, that will display the same envelope multiple times

betegon (commented):
I think there's the "break" thing that we may wanna add

packages/sidecar/src/cli.ts
betegon: i Mean the `windowsVerbatimArguments` and the `shell` combined

packages/sidecar/src/cli.ts
BYK: Not a break but goot point. I'll check this out.

packages/sidecar/src/cli.ts
BYK: `!code` covers that?

packages/sidecar/src/cli.ts
BYK: Hmm, good point. I'll see what we can do about this.

packages/sidecar/src/cli.ts
BYK: I turned it on explicitly to make sure we are passing the script without any changes. I'll revisit if this causes any issues for anyone.

packages/sidecar/src/cli.ts
betegon: lol yes 😆

packages/sidecar/src/cli.ts
BYK: Okay I checked this and this actually only happens on Linux and unfortunately doesn't have a great solution. The only consolation is running `.kill()` actually sends `SIGTERM` which should trigger a graceful shutdown. So I cannot do anything more than a 🤷🏻 here for now.

packages/sidecar/src/cli.ts
BYK: Fixed, great catch!

PR #1033, opened 2025-10-23
tests: added more tests for message buffer

MathurAditya724:
All of theses are generated by AI 🤖 and it cover almost all the uses of the ring buffer and subscribers. I believe we can refine them further after the rewrite of message buffer

### DIFF
--- a/packages/sidecar/src/messageBuffer.spec.ts
+++ b/packages/sidecar/src/messageBuffer.spec.ts
@@ -279,6 +279,14 @@ describe("messageBuffer", () => {
   });
 
   describe("concurrent operations", () => {
+    beforeEach(() => {
+      vi.useFakeTimers();
+    });
+
+    afterEach(() => {
+      vi.restoreAllMocks();
+    });
+
     it("should handle rapid puts", () => {
       const messageBuffer = new MessageBuffer<number>(10);
 


packages/sidecar/src/messageBuffer.spec.ts
```diff
@@ -1,24 +1,371 @@
...
+    });
+
+    it("should handle exact buffer size fill", () => {
+      const messageBuffer = new MessageBuffer<number>(5);
+
+      for (let i = 0; i < 5; i++) {
+        messageBuffer.put(i);
+      }
+
+      expect(messageBuffer.read()).toEqual([4, 3, 2, 1, 0]);
```
betegon, line 25: what's really he point of this one?

packages/sidecar/src/messageBuffer.spec.ts
```diff
@@ -1,24 +1,371 @@
...
+
+    it("should maintain correct order after wraparound", () => {
+      const messageBuffer = new MessageBuffer<number>(4);
+
+      for (let i = 0; i < 10; i++) {
+        messageBuffer.put(i);
+      }
+
+      expect(messageBuffer.read()).toEqual([9, 8, 7, 6]);
+    });
```
betegon: this is the same test as the above one, i'd remove this one.

packages/sidecar/src/messageBuffer.spec.ts
```diff
@@ -1,24 +1,371 @@
...
+    it("should handle buffer size of 1", () => {
+      const messageBuffer = new MessageBuffer<number>(1);
+
+      messageBuffer.put(0);
+      messageBuffer.put(1);
+      messageBuffer.put(2);
+
+      expect(messageBuffer.read()).toEqual([2]);
+    });
+  });
```
betegon, line 60: i'm not sure about this. i just get the impression we can get the fails with the test I haven't said anything about (above).

packages/sidecar/src/messageBuffer.spec.ts
```diff
@@ -1,24 +1,371 @@
...
+      const messageBuffer = new MessageBuffer<number>(5);
+
+      messageBuffer.clear();
+
+      expect(messageBuffer.read()).toEqual([]);
+    });
+  });
+
+  describe("reset operation", () => {
+    it("should clear items but maintain write position", () => {
```
betegon, line 99: > but maintain write position

 Is this true? we don't test that here.

_(the test is great, just that part of the description that bothers me 😆 )_

packages/sidecar/src/messageBuffer.spec.ts
BYK: Edge case where your incoming message count exactly equals to the buffer size. Ensures you don't have any off-by-one errors anywhere and data is intact even at boundary conditions.

packages/sidecar/src/messageBuffer.spec.ts
BYK: Agree with @betegon here. The first test tests both conditions so the name may be amended.

packages/sidecar/src/messageBuffer.spec.ts
BYK: Again, checking boundary conditions. I opt to keeping this one although we should ideally never have a buffer size of 1 😅

packages/sidecar/src/messageBuffer.spec.ts
BYK: Agree with @betegon -- if the aim of the test is to ensure write position is not changed, we need to assert that somehow.

Aside, why is this important? Like why would we prefer `reset` over `clear` (or vice versa)?

packages/sidecar/src/messageBuffer.spec.ts
```diff
@@ -1,24 +1,371 @@
...
+      expect(result).toEqual([99, 98, 97, 96, 95, 94, 93, 92, 91, 90]);
+    });
+
+    it("should handle interleaved put and read operations", () => {
+      const messageBuffer = new MessageBuffer<number>(5);
+
+      messageBuffer.put(0);
+      messageBuffer.put(1);
+      expect(messageBuffer.read()).toEqual([1, 0]);
+
```
BYK, line 315: I'd add some subsccribers here and pauses (sleep etc) to make sure this part is robust.

packages/sidecar/src/messageBuffer.spec.ts
```diff
@@ -1,24 +1,371 @@
...
+      messageBuffer.put(null);
+      messageBuffer.put(undefined);
+      messageBuffer.put(0);
+
+      const result = messageBuffer.read();
+      expect(result).toEqual([0, undefined, null]);
+    });
+
+    it("should maintain FIFO order when reading in reverse", () => {
```
BYK: I think this is covered in many other tests above so not sure if we need to have it explicitly.

MathurAditya724: Makes sense, removing this then.

packages/sidecar/src/messageBuffer.spec.ts
MathurAditya724: Oh, it was there I was just checking something so must have removed that, added that back 😅 

`clear` function removes the subscriber's timeout so they will stop receiving updates, while `reset` just clears the `items` array

PR #1034, opened 2025-10-23
chore: updated the images for the tray icon

MathurAditya724:
Just minor sizing and spacing refinement in the tray icon
Screenshot 2025-10-23 at 4 09 41 PM


BYK:
Looks like this: 
image

So the transparency seems to be no there or proper?

packages/electron/resources/tray/logoTemplate@2x.png
BYK, line 1: Yeah these files have white backgrounds, they should have transparent instead

BYK (approved):

PR #1035, opened 2025-10-23
ref(sidecar): Refactor CLI for better organization

BYK:
Splits command handlers into individual modules for easier reasoning and composing.

This also helps fix the issue where we run the server _after_ we run a command with `spotlight run`. Now it is the other way around: the server starts first, then we run the command.

### DIFF
--- a/packages/sidecar/src/cli/mcp.ts
+++ b/packages/sidecar/src/cli/mcp.ts
@@ -1,5 +1,5 @@
-import type { CLIHandlerOptions } from "../types/cli.js";
 import { setupSidecar } from "../main.js";
+import type { CLIHandlerOptions } from "../types/cli.js";
 
 export default async function mcp({ port, basePath, filesToServe }: CLIHandlerOptions) {
   return await setupSidecar({


packages/sidecar/src/cli/run.ts
```diff
@@ -0,0 +1,82 @@
+import { spawn } from "node:child_process";
+import { readFileSync } from "node:fs";
+import * as path from "node:path";
+import { logger } from "src/logger.js";
+import type { CLIHandlerOptions } from "src/types/cli.js";
+import { getSpotlightURL } from "src/utils/extras.js";
```
betegon: these imports fail when running the compiled code, node thinks `src` is a module. we could use relative imports for this. 

Image

packages/sidecar/src/cli/mcp.ts
```diff
@@ -0,0 +1,12 @@
+import type { CLIHandlerOptions } from "src/types/cli.js";
```
betegon: _read relative imports comment_

packages/sidecar/src/cli/run.ts
betegon: ok i didn't see Seer comment but it's the same one.

PR #1036, opened 2025-10-23
feat(spotlight): Use dynamic port for run cmd

### LINKED ISSUE #1025: Automatic port selection
We should add automatic port selection through `--port 0` so Spotlight can find the first available port rather than sticking with its default or asking a specific port from the user.

This behavior should be the default for `spotlight run` (#987)

BYK:
Closes #1025.

Had to simplify and modularize the functions in `main.ts` -- especially splitting out the `stdioMCP` logic into the `cli/mcp.ts` where it should have always been. This made things a lot more simple for the regular `server` command and all MCP-related complex logic is now contained in `cli/mcp.ts` and in a more understandable way.

Rest is some more reorganization, and essentially making sure we return the created server instance from `server` and `tail` commands so we can get the automatically assigned port from there when a dynamic port is used.

### DIFF
--- a/packages/sidecar/src/cli.ts
+++ b/packages/sidecar/src/cli.ts
@@ -10,6 +10,10 @@ import run from "./cli/run.js";
 import server from "./cli/server.js";
 import tail from "./cli/tail.js";
 
+// When port is set to `0` for a server to listen on,
+// it tells the OS to find a random available port.
+// This can then be retrieved after the server starts.
+const MAGIC_PORT_FOR_DYNAMIC_ASSIGNMENT = 0;
 const PARSE_ARGS_CONFIG = {
   options: {
     port: {
@@ -66,7 +70,7 @@ export function parseCLIArgs(): CLIArgs {
   // Handle legacy positional argument for port (backwards compatibility)
   const portInput = positionals.length === 1 && /^\d{1,5}$/.test(positionals[0]) ? positionals.shift() : values.port;
 
-  const port = portInput != null ? Number(portInput) : runToken ? 0 : DEFAULT_PORT;
+  const port = portInput != null ? Number(portInput) : runToken ? MAGIC_PORT_FOR_DYNAMIC_ASSIGNMENT : DEFAULT_PORT;
   if (Number.isNaN(port)) {
     // Validate port number
     console.error(`Error: Invalid port number '${portInput}'`);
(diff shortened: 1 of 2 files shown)


packages/sidecar/src/cli/tail.ts
```diff
@@ -76,6 +82,6 @@ export default async function tail({ port, cmdArgs, basePath, filesToServe }: CL
       process.exit(1);
     }
 
-    await setupSidecar({ port, filesToServe, basePath, onEnvelope, isStandalone: true });
+    return await setupSidecar({ port, filesToServe, basePath, onEnvelope, isStandalone: true });
```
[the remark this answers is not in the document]
BYK: This is mostly correct but was intentional. I'll think about this case more before making a final decision.

packages/sidecar/src/cli.ts
```diff
@@ -1,27 +1,112 @@
...
+  }
+  const { values, positionals } = parseArgs({
+    args,
+    ...PARSE_ARGS_CONFIG,
+  });
+
+  // Handle legacy positional argument for port (backwards compatibility)
+  const portInput = positionals.length === 1 && /^\d{1,5}$/.test(positionals[0]) ? positionals.shift() : values.port;
+
+  const port = portInput != null ? Number(portInput) : runToken ? 0 : DEFAULT_PORT;
```
betegon: had look up what `0` meant 😆

packages/sidecar/src/cli/run.ts
```diff
@@ -5,14 +5,20 @@ import { logger } from "../logger.js";
...
 
 export default async function run({ port, cmdArgs, basePath, filesToServe }: CLIHandlerOptions) {
-  await tail({ port, cmdArgs: [], basePath, filesToServe });
+  const serverInstance = await tail({ port, cmdArgs: [], basePath, filesToServe });
+  if (!serverInstance) {
+    logger.error("Failed to start Spotlight sidecar server.");
+    process.exit(1);
+  }
 
+  const actualServerPort = (serverInstance.address() as AddressInfo).port;
```
betegon, line 20: `.port()` can return `null` so this can crash in that case. don't we need a null check?

BYK: In our case, if it returns null it indicates a problem with the server so we should never reach that case here. That's why I didn't add a check. Will add a comment about this tho.

packages/sidecar/src/cli.ts
BYK: Heh, will add a comment about it

packages/sidecar/src/cli/tail.ts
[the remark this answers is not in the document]
BYK: Decided to leave it as is until someone complains about a specific use case.

Issue #1014, opened 2025-10-20
Support more output formats in CLI


BYK:
We should support `logfmt` (use this as default probably) and `json` output formats from `spotlight tail` command with a `--format` flag. We may optionally support a `--json` flag for convenience.

PR #1039, opened 2025-10-24
Feat(sidecar): Logfmt formatter

betegon:
from #1014 

Adds `logfmt` format for sentry traces, errors and logs. Created a directory structure + a `Formatter` interface so it's easier to add new formatters.

the next follow up will be to add a human readable format, as the one we have right now is just markdown style output.

Screenshot 2025-10-24 at 17 32 08
Screenshot 2025-10-24 at 17 39 27

Screenshot 2025-10-24 at 18 03 31

### DIFF
--- a/packages/sidecar/src/cli/tail.ts
+++ b/packages/sidecar/src/cli/tail.ts
@@ -60,7 +60,7 @@ export default async function tail({
     : new Set([...eventTypes.flatMap(type => NAME_TO_TYPE_MAPPING[type] || [])]);
   const onEnvelope: (envelope: ParsedEnvelope["envelope"]) => void = envelope => {
     if (envelope[1].some(([header]) => header.type && types.has(header.type))) {
-      displayEnvelope(envelope, format as FormatterType);
+      displayEnvelope(envelope, format);
     }
   };
 


packages/sidecar/src/formatters/logfmt/traces.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (trace.op) data.op = trace.op;
+    if (trace.status) data.status = trace.status;
+    if (trace.description) data.description = trace.description;
+  }
+
+  if (event.transaction) data.transaction = event.transaction;
+
+  if (Number.isFinite(event.timestamp) && Number.isFinite(event.start_timestamp)) {
+    data.duration_ms = Math.round((event.timestamp - event.start_timestamp) * 1000);
+  }
```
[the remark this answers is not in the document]
betegon: it's not our problem we we have a negative duration, it may even indicate a problem. Best we can do is changing the trace status or adding a new field, but if we do that with duration, we should then check every other field.

i'd leave it as it is.

packages/sidecar/src/formatters/general.ts
```diff
@@ -0,0 +1,41 @@
...
+} from "../parser/index.js";
+import human from "./human/index.js";
+import logfmt from "./logfmt/index.js";
+import type { Formatter, FormatterType } from "./types.js";
+
+const FORMATTERS: Record<FormatterType, Formatter> = {
+  human,
+  logfmt,
+  // TODO: add json formatter
+  json: logfmt, // fallback until we have a json formatter
```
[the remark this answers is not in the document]
betegon: on purpose as we're adding json in a bit (prior to the release probably)

packages/overlay/src/telemetry/components/events/AICopyButton.tsx
```diff
@@ -1,5 +1,6 @@
 import { captureException } from "@sentry/core";
-import { formatEventOutput, processErrorEvent } from "@spotlightjs/sidecar/format";
+import { processErrorEvent } from "@spotlightjs/sidecar/formatters/human/errors";
```
BYK, line 2: I'd name this formatter `markdown` or `md` since we're gonna replace the human one soon?

packages/sidecar/src/cli/help.ts
```diff
@@ -1,4 +1,5 @@
 import { EVERYTHING_MAGIC_WORDS, NAME_TO_TYPE_MAPPING } from "../cli/tail.js";
+import { VALID_FORMATTERS } from "../formatters/types.js";
```
BYK, line 2:
```suggestion
import { AVAILABLE_FORMATTERS } from "../formatters/types.js";
```

Available would be better than "valid"

packages/sidecar/src/cli/run.ts
```diff
@@ -1,11 +1,11 @@
 import { spawn } from "node:child_process";
 import { readFileSync } from "node:fs";
+import type { AddressInfo } from "node:net";
```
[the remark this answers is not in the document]
betegon: oh god, that's true

packages/sidecar/src/cli/tail.ts
```diff
@@ -59,7 +53,7 @@ export default async function tail({ port, cmdArgs, basePath, filesToServe }: CL
     : new Set([...eventTypes.flatMap(type => NAME_TO_TYPE_MAPPING[type] || [])]);
   const onEnvelope: (envelope: ParsedEnvelope["envelope"]) => void = envelope => {
     if (envelope[1].some(([header]) => header.type && types.has(header.type))) {
-      displayEnvelope(envelope);
+      displayEnvelope(envelope, format as FormatterType);
```
BYK: Protip: we can have a "isFormatter" function with return type "is FormatterType", validate it before passing here and ensure the type?

packages/sidecar/src/cli/tail.ts
```diff
@@ -65,7 +60,7 @@ export default async function tail({
...
```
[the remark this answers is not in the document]
BYK: This sounds legit

packages/sidecar/src/cli/tail.ts
```diff
@@ -24,18 +25,11 @@ const SUPPORTED_ENVELOPE_TYPES = new Set(Object.values(NAME_TO_TYPE_MAPPING).fla
...
 export const SUPPORTED_TAIL_ARGS = new Set([...Object.keys(NAME_TO_TYPE_MAPPING), ...EVERYTHING_MAGIC_WORDS]);
 
-const SEPARATOR = Array(10).fill("─").join("");
-
-function displayEnvelope(envelope: ParsedEnvelope["envelope"]) {
-  console.log(`${envelope[0].event_id} | ${envelope[1][0][0].type} | ${envelope[0].sdk?.name}\n\n`);
-  const lines = formatEnvelope(envelope);
+function displayEnvelope(envelope: ParsedEnvelope["envelope"], format: FormatterType) {
+  const lines = formatEnvelope(envelope, format);
   if (lines.length > 0) {
```
BYK, line 30: I'd say this should be handled by the formatters at this point?

packages/sidecar/src/formatters/logfmt/errors.ts
```diff
@@ -0,0 +1,71 @@
+import type { ErrorEvent } from "@sentry/core";
+import logfmt from "logfmt";
+
+export function formatError(event: ErrorEvent): string {
+  const data: Record<string, any> = {
+    type: "error",
+    event_id: event.event_id,
+    timestamp: event.timestamp ? new Date(event.timestamp * 1000).toISOString() : new Date().toISOString(),
```
BYK, line 8: I think we should use a centralized helper function to avoid repeating this logic around

packages/sidecar/src/formatters/logfmt/errors.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (event.user?.email) data.user_email = event.user.email;
+  if (event.user?.username) data.user_username = event.user.username;
+
+  if (event.request?.url) data.request_url = event.request.url;
+  if (event.request?.method) data.request_method = event.request.method;
+
+  if (event.logger) data.logger = event.logger;
+  if (event.server_name) data.server_name = event.server_name;
+
+  if (event.breadcrumbs?.length) data.breadcrumb_length = event.breadcrumbs.length;
```
BYK, line 61: Can we have a mapping of fields and loop over this and concentrate this "if defined set to X" logic to avoid a billion if statements?

packages/sidecar/src/formatters/logfmt/errors.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (event.spans?.length) data.span_length = event.spans.length;
+
+  if (event.tags) {
+    for (const [key, value] of Object.entries(event.tags)) {
+      data[`tag_${key}`] = value;
```
BYK, line 66:
```suggestion
      data[`tag:${key}`] = value;
```

Or "tag.key" or just drop the tag prefix altogether

packages/sidecar/src/formatters/logfmt/logs.ts
```diff
@@ -0,0 +1,32 @@
...
+    data.trace_id = log.trace_id;
+  }
+
+  if (log.severity_number) {
+    data.severity_number = log.severity_number;
+  }
+
+  if (log.attributes) {
+    for (const [key, attr] of Object.entries(log.attributes)) {
+      if (!key.startsWith("sentry.")) {
```
BYK, line 25: This logic needs explanation. Why are we doing this?

packages/sidecar/src/formatters/logfmt/traces.ts
```diff
@@ -0,0 +1,58 @@
+import logfmt from "logfmt";
+import type { SentryTransactionEvent } from "../../parser/index.js";
+
+export function formatTrace(event: SentryTransactionEvent): string {
+  const data: Record<string, any> = {
+    type: "trace",
+    event_id: event.event_id,
+  };
+
+  if (Number.isFinite(event.timestamp)) {
```
BYK, line 10: What? Why are we using isFinite here instead of the other simpler ways we do in other events? We should centralize this logic as I mentioned earlier

packages/sidecar/src/formatters/general.ts
BYK, line 19: Nope. If people ask for JSON and you give them invalid JSON they'd be mad

packages/sidecar/src/formatters/general.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export function formatEnvelope(envelope: ParsedEnvelope["envelope"], format: FormatterType): string[] {
+  const [, items] = envelope;
+  const formatter = FORMATTERS[format];
+
+  const formatted: string[] = [];
+  for (const item of items) {
+    const [{ type }, payload] = item;
+
+    if (type === "event" && isErrorEvent(payload as SentryEvent)) {
```
BYK, line 30: This also calls for a mapping of event types to formatters as we'll have more in the future and repeated if statements are not very easy to follow or maintain.

packages/sidecar/src/formatters/types.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Interface that all formatters must implement.
+ * Each formatter provides methods to format different types of Sentry events.
+ */
+export interface Formatter {
+  /**
+   * Format an error event
+   */
+  formatError(payload: ErrorEvent): string;
```
BYK, line 12: I'd drop the "format" prefix and expose the mapping from here where the keys come from a fixed set of event types

packages/sidecar/package.json
```diff
@@ -38,8 +38,16 @@
       "types": "./dist/parser/index.d.ts"
     },
     "./format": {
-      "import": "./dist/format/index.js",
-      "types": "./dist/format/index.d.ts"
+      "import": "./dist/formatters/index.js",
+      "types": "./dist/formatters/index.d.ts"
+    },
+    "./formatters/human/errors": {
```
BYK, line 44: Not a fan of having to export the submodules explicitly like this for every formatter. Maybe we should keep them in a single file pee formatter?

PR #1041, opened 2025-10-24
Ref(sidecar): Formatters

betegon:
It's a follow up from #1039

### DIFF
--- a/packages/sidecar/src/formatters/logfmt/traces.ts
+++ b/packages/sidecar/src/formatters/logfmt/traces.ts
@@ -1,6 +1,6 @@
 import logfmt from "logfmt";
 import type { SentryTransactionEvent } from "../../parser/index.js";
-import { formatTimestamp, mapFields, mapSdkFields, mapTags } from "../utils.js";
+import { formatTimestamp, getDuration, mapFields, mapSdkFields, mapTags } from "../utils.js";
 
 export function formatTrace(event: SentryTransactionEvent): string {
   const data: Record<string, any> = {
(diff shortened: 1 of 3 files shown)


packages/sidecar/src/formatters/logfmt/traces.ts
```diff
@@ -1,58 +1,61 @@
...
+      status: "status",
+      description: "description",
+    });
   }
 
-  if (event.transaction) data.transaction = event.transaction;
+  const { timestamp, start_timestamp } = event;
+  if (typeof timestamp === "number" && typeof start_timestamp === "number") {
+    data.duration_ms = Math.round((timestamp - start_timestamp) * 1000);
+  }
```
BYK: You should be able to standardize this one too with `getSpanDuration` or something along those lines.

packages/sidecar/src/formatters/logfmt/traces.ts
```diff
@@ -1,58 +1,61 @@
...
 
-  if (Number.isFinite(event.timestamp) && Number.isFinite(event.start_timestamp)) {
+  if (
+    event.timestamp !== undefined &&
+    event.start_timestamp !== undefined &&
+    Number.isFinite(event.timestamp) &&
+    Number.isFinite(event.start_timestamp)
```
BYK: Again, what's the reason for `Number.isFinite`? Comments on this please :)

packages/sidecar/src/formatters/logfmt/traces.ts
betegon: ok done!

packages/sidecar/src/formatters/logfmt/traces.ts
betegon: it gets you a guard from a null/undefined or an string timestamp in one 😬. No need for that, `getSpanDuration` doesn't use it now.

packages/sidecar/src/formatters/logfmt/logs.ts
```diff
@@ -1,30 +1,23 @@
...
+    trace_id: "trace_id",
+    severity_number: "severity_number",
+  });
 
   if (log.attributes) {
     for (const [key, attr] of Object.entries(log.attributes)) {
-      if (!key.startsWith("sentry.")) {
-        data[key] = attr.value;
-      }
+      data[key] = attr.value;
```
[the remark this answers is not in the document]
betegon: on purpose. I was filtering them as I see that's what we did in the `md` format but these could be useful, like `sentry.trace.parent_span_id`

packages/sidecar/src/formatters/utils.ts
```diff
@@ -0,0 +1,69 @@
+/**
+ * Format a Sentry timestamp (in seconds) to ISO string.
+ * Sentry timestamps are Unix timestamps in seconds, so we multiply by 1000 to convert to milliseconds.
+ * Falls back to epoch (1970-01-01T00:00:00.000Z) if timestamp is missing/invalid.
+ */
+export function formatTimestamp(timestamp?: number): string {
+  const date = timestamp ? new Date(timestamp * 1000) : new Date(0);
+  return date.toISOString();
```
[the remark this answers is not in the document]
betegon: on purpose to have the same behaviour on both.

PR #1044, opened 2025-10-26
feat(sidecar): Auto capture stdio logs as Sentry log events

BYK:
When using `spotlight run` if the process is not using Sentry SDK
with logs integration, we lose all its logging output that comes
to stdout/stderr. This change automatically captures that output
checks against any log items we receive from the process on the
sidecar and sends them as Sentry log events only if we cannot
get any logs from our integration.

Note: mostly vibe-coded

### DIFF
--- a/packages/sidecar/src/cli/tail.ts
+++ b/packages/sidecar/src/cli/tail.ts
@@ -91,6 +91,7 @@ export default async function tail(
   try {
     const client = await connectUpstream(port);
     client.addEventListener(SENTRY_CONTENT_TYPE, event => onEnvelope!(JSON.parse(event.data)));
+    // Early return - don't start our own server if we can connect to an upstream one
     return undefined;
   } catch (err) {
     // if we fail, fine then we'll start our own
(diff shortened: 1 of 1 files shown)


packages/sidecar/src/cli/run.ts
```diff
@@ -2,13 +2,138 @@ import { spawn } from "node:child_process";
...
-import tail from "./tail.js";
+import tail, { type OnItemCallback } from "./tail.js";
+
+const STDIO_LOG_MARKER = "__spotlight_stdio_log";
+
+/**
+ * Creates a Sentry log envelope from a log message
+ * Marks the log with STDIO_LOG_MARKER to exclude from duplicate detection
+ */
+function createLogEnvelope(level: "info" | "error", body: string, timestamp: number): Buffer {
```
betegon, line 18: could we move this function to another file to avoid growing the run command file?

packages/sidecar/src/cli/run.ts
```diff
@@ -2,13 +2,138 @@ import { spawn } from "node:child_process";
...
+    body,
+    attributes: {
+      [STDIO_LOG_MARKER]: {
+        value: true,
+        type: "boolean",
+      },
+    },
+  };
+
+  // Create log event payload
```
betegon: you're going to hate me for this, but we could remove all redundant comment 😬

packages/sidecar/src/cli/run.ts
```diff
@@ -2,13 +2,138 @@ import { spawn } from "node:child_process";
...
+  // Build envelope (newline-delimited JSON)
+  const parts = [JSON.stringify(envelopeHeader), JSON.stringify(itemHeader), payloadJson];
+
+  return Buffer.from(`${parts.join("\n")}\n`, "utf-8");
+}
+
+/**
+ * Sends an envelope to the sidecar server via HTTP POST
+ */
+async function sendEnvelopeToSidecar(envelopeBuffer: Buffer, sidecarPort: number): Promise<void> {
```
betegon: what about using `onEnvelope` callback, similar to what we have on `tail` command.

betegon: i mean anything but doing a http request to our own. server

packages/sidecar/src/cli/run.ts
BYK: I prefer to keep it here as it's the only place this is used now. We can revisit this later.

packages/sidecar/src/cli/tail.ts
```diff
@@ -88,6 +91,8 @@ export default async function tail({
   try {
     const client = await connectUpstream(port);
     client.addEventListener(SENTRY_CONTENT_TYPE, event => onEnvelope!(JSON.parse(event.data)));
+    // Early return - don't start our own server if we can connect to an upstream one
+    return undefined;
```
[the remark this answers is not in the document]
BYK: This is intentional.

PR #1059, opened 2025-10-28
fix(mcp): Fix hang during shutdown during mcp mode

### LINKED ISSUE #1058: MCP proxy mode refuses to shutdown
When we connect to an upstream MCP server with `spotlight mcp`, the process refuses to exit (neither `SIGINT` nor `SIGTERM` works)

BYK:
Fixes #1058

### DIFF
--- a/.changeset/sweet-vans-tan.md
+++ b/.changeset/sweet-vans-tan.md
@@ -1,0 +1,6 @@
+---
+"@spotlightjs/sidecar": patch
+"@spotlightjs/spotlight": patch
+---
+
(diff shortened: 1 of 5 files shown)


BYK:
Waiting on https://github.com/honojs/middleware/pull/1535 to be merged and released and we'll bump that package's version. That fix is required to fix MCP server not shutting down once it gets a connection.

betegon (commented):
looks good, i'd like to try it out but i suppose we have to wait until it's fixed in hono

PR #1088, opened 2025-11-03
Fix(docs): Small fixes for the new packages

betegon:
- Removed `-sidecar` from a bunch of places
- Marked overlay options as deprecated
- Correct the default logging formatter in docs
- Update examples


packages/website/src/content/docs/docs/reference/configuration.md
```diff
@@ -89,62 +89,35 @@ Experiment names are:
 
 - `sentry:focus-local-events` - if set to true, errors and traces will hide events from other sessions when possible.
 
-#### `injectImmediately`
+:::note[Deprecated Options]
```
BYK: I'd just remove this page completely.

PR #1102, opened 2025-11-05
chore: added a more clear error message

MathurAditya724:


packages/sidecar/src/cli/run.ts
```diff
@@ -102,6 +102,7 @@ export default async function run({ port, cmdArgs, basePath, filesToServe, forma
   const serverInstance = await tail({ port, cmdArgs: [], basePath, filesToServe, format }, logChecker);
   if (!serverInstance) {
     logger.error("Failed to start Spotlight sidecar server.");
+    logger.error(`The port ${port} might already be in use by another process.`);
```
betegon:
```suggestion
logger.error(`The port ${port} might already be in use — most likely by another Spotlight instance.`);
```

I'd say something like this ;)

BYK (approved):

PR #1104, opened 2025-11-05
feat: Added SSE Last Event Id Header

### LINKED ISSUE #1095: Implement SSE Last-Event-ID Support for Event Stream Resumption
## Summary
Implement Server-Sent Events (SSE) `Last-Event-ID` functionality to enable the sidecar to resume event streaming from a specific point when the overlay reconnects. This will prevent event loss during disconnections and allow clients to catch up on missed events.

## Motivation
Currently, when the overlay UI disconnects and reconnects to the sidecar's `/stream` endpoint, it loses all events that occurred during the disconnection. With SSE `Last-Event-ID` support, the overlay can automatically resume from where it left off, ensuring no telemetry data is missed.

### Sidecar (Backend)
- **SSE Streaming**: Located in `packages/sidecar/src/routes/stream/index.ts` and `packages/sidecar/src/routes/stream/streaming.ts`
- **Event Storage**: Uses a circular `MessageBuffer` (500 events by default) in `packages/sidecar/src/messageBuffer.ts`
- **Event IDs**: Each envelope gets a unique `__spotlight_envelope_id` generated using `uuidv7()` in `packages/sidecar/src/parser/processEnvelope.ts` (line 47)
- **Current Stream Implementation**: Subscribers receive all new events as they arrive, but no mechanism exists to replay historical events

### Overlay (Frontend)
- **Connection**: Uses native `EventSource` API in `packages/overlay/src/sidecar.ts`
- **Storage**: Events stored in Zustand store (`packages/overlay/src/telemetry/store/`)
- **Current Behavior**: No tracking of last received event ID; complete state loss on reconnection

### 1. Sidecar Changes

**a) Modify SSE Stream Handler** (`packages/sidecar/src/routes/stream/index.ts`)
- Read `Last-Event-ID` header from incoming requests
- When `Last-Event-ID` is provided:
  - Query the `MessageBuffer` for events after the provided ID
  - Send historical events first (with their original IDs)
  - Then continue streaming new events

**b) Update SSE Message Format** (`packages/sidecar/src/routes/stream/streaming.ts`)
- Add `id` field to SSE messages using the `__spotlight_envelope_id`
- Current format:
  ```javascript
  stream.writeSSE({
    event: `${container.getContentType()}${base64Indicator}`,
    data: JSON.stringify(parsedEnvelope.envelope),
  });
  ```
- New format should include:
  ```javascript
  stream.writeSSE({
    id: envelope.envelope[0].__spotlight_envelope_id, // Add this
    event: `${container.getContentType()}${base64Indicator}`,
    data: JSON.stringify(parsedEnvelope.envelope),
  });
  ```

**c) Extend MessageBuffer** (`packages/sidecar/src/messageBuffer.ts`)
- Add method to read events starting from a specific envelope ID:
  ```typescript
  readFrom(lastEventId: string): T[]
  ```
- This should leverage the existing circular buffer (currently stores 500 events)
- Consider buffer overflow scenarios (when requested ID has been evicted)

### 2. Overlay Changes

**a) EventSource Connection** (`packages/overlay/src/sidecar.ts`)
- The browser's `EventSource` API automatically sends `Last-Event-ID` header on reconnection
- No explicit changes needed for reconnection handling
- However, may want to track last event ID for debugging/logging purposes

**b) State Management** (`packages/overlay/src/telemetry/store/`)
- Track the last received envelope ID in the store
- Handle potential duplicate events during reconnection (envelope IDs can be used for deduplication)
- Current stores already use Maps with IDs, making deduplication straightforward

**c) Connection State Handling**
- Consider showing a "catching up" state in the UI when replaying historical events
- Track metrics: number of events caught up, time to sync, etc.

### 3. Edge Cases to Handle

1. **Buffer Overflow**: Requested `Last-Event-ID` is older than the oldest event in the buffer
   - Send an error event or specific marker
   - Client should handle full state reset

2. **Invalid Event ID**: Client sends an ID that never existed
   - Treat as new connection, start from current position

3. **Duplicate Events**: During replay, client might receive events it already has
   - Use `__spotlight_envelope_id` for deduplication
   - Overlay's `Map`-based storage already prevents duplicates by ID

4. **Multiple Content Types**: Current implementation supports different event types
   - Ensure ID tracking works across all content types
   - Currently: `application/x-sentry-envelope`, `application/x-sentry-envelope;base64`

5. **Performance**: Replaying large numbers of events on reconnection
   - Consider batch size limits
   - May need throttling for replay

### Phase 1: Basic SSE ID Support
- Add `id` field to SSE messages in sidecar
- Verify EventSource receives and tracks IDs
- Test manual reconnection

### Phase 2: Historical Event Replay
- Implement `readFrom()` in MessageBuffer
- Add Last-Event-ID header handling in stream endpoint
- Test with simulated disconnections

### Phase 3: Edge Cases & Polish
- Handle buffer overflow scenarios
- Add deduplication in overlay
- Add UI indicators for catch-up state
- Performance testing with large event counts

### Phase 4: Testing
- Unit tests for MessageBuffer.readFrom()
- Integration tests for reconnection scenarios
- E2E tests simulating network interruptions

## Technical Details

**SSE ID Format**: Use existing `uuidv7()` generated `__spotlight_envelope_id`
- Format: UUIDv7 (time-ordered, suitable for sequential tracking)
- Example: `0190d9a7-0b52-7000-a000-000000000000`

**Buffer Retention**: Current circular buffer holds 500 events
- Consider making this configurable
- Document that events older than buffer size cannot be replayed

**Alternative ID Strategy Considerations**:
- Current: Envelope ID (one per envelope, which may contain multiple events)
- Alternative: Per-event IDs (more granular but more complex)
- Recommendation: Start with envelope IDs for simplicity

## Related Files

**Sidecar:**
- `packages/sidecar/src/routes/stream/index.ts` - Stream endpoint
- `packages/sidecar/src/routes/stream/streaming.ts` - SSE utilities
- `packages/sidecar/src/messageBuffer.ts` - Event buffer
- `packages/sidecar/src/parser/processEnvelope.ts` - Envelope ID generation
- `packages/sidecar/src/utils/eventContainer.ts` - Event container wrapper

**Overlay:**
- `packages/overlay/src/sidecar.ts` - EventSource connection
- `packages/overlay/src/telemetry/store/slices/envelopesSlice.ts` - Envelope storage
- `packages/overlay/src/telemetry/store/slices/eventsSlice.ts` - Event storage

## References

- [MDN: Server-Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events)
- [SSE Specification](https://html.spec.whatwg.org/multipage/server-sent-events.html)
- [EventSource API](https://developer.mozilla.org/en-US/docs/Web/API/EventSource)

---

This implementation would significantly improve the reliability of Spotlight's real-time telemetry streaming, especially in development environments where network conditions may be unstable or when developers need to restart services frequently.

MathurAditya724:
Closes #1095

### DIFF
--- a/packages/sidecar/src/messageBuffer.spec.ts
+++ b/packages/sidecar/src/messageBuffer.spec.ts
@@ -1,5 +1,7 @@
+import { UUID } from "uuidv7";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { MessageBuffer } from "./messageBuffer.js";
+import { EventContainer } from "./utils/eventContainer.js";
 
 describe("messageBuffer", () => {
   describe("ring buffer behavior", () => {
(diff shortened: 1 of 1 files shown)


packages/sidecar/src/messageBuffer.ts
```diff
@@ -87,11 +87,36 @@ export class MessageBuffer<T> {
...
+
+        if (item[1] instanceof EventContainer) {
+          const envelope = item[1].getParsedEnvelope();
+          if (envelope?.envelope && envelope.envelope[0].__spotlight_envelope_id === lastEventId) {
+            // Start from the position after the found event
+            startPos = i + 1;
+            break;
+          }
+        }
+      }
```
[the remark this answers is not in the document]
MathurAditya724: For this only I was telling that we can use the timestamp, that way we can do a binary search.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -87,11 +87,36 @@ export class MessageBuffer<T> {
...
+      const start = this.head;
+      const end = this.writePos;
+
+      for (let i = start; i < end; i++) {
+        const item = this.items[i % this.size];
+        if (item == null) continue;
+
+        if (item[1] instanceof EventContainer) {
+          const envelope = item[1].getParsedEnvelope();
+          if (envelope?.envelope && envelope.envelope[0].__spotlight_envelope_id === lastEventId) {
```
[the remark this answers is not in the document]
BYK: @MathurAditya724 can you make sure `__spotlight_envelope_id` is always a string or `undefined`?

packages/sidecar/src/messageBuffer.ts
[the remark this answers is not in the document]
BYK: We are using uuidv7 which should be alpha/time-sortable: https://uuid7.com/ (see the first item in benefits)

So we should be able to do binary search here which would probably the most efficient way. Starting from the end also makes sense. Not so sure about a revers-hash-map as it's more book-keeping. I'd say binary search should be the optimal solution.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -87,11 +87,36 @@ export class MessageBuffer<T> {
...
     const readerId = uuidv7();
+
+    // Determine starting position based on lastEventId
+    let startPos = this.head;
+    if (lastEventId) {
+      // Find the position of the last event ID
+      const start = this.head;
+      const end = this.writePos;
+
+      for (let i = start; i < end; i++) {
```
betegon: we could change this linear search to using a map like  EnvelopeIdPosition= [envelopeId, position].

We could update it in the `put` and checking it in here in the `subscribe`. Then we will need to clean it up like we do with the filenameCache.

packages/sidecar/src/messageBuffer.ts
```diff
@@ -87,11 +87,36 @@ export class MessageBuffer<T> {
     }
   }
 
-  subscribe(callback: (item: T) => void): string {
+  subscribe(callback: (item: T) => void, lastEventId?: string): string {
     const readerId = uuidv7();
+
+    // Determine starting position based on lastEventId
+    let startPos = this.head;
+    if (lastEventId) {
```
betegon, line 206: in the issue (did @BYK wrote it?) says

> Buffer Overflow: Requested Last-Event-ID is older than the oldest event in the buffer
> 
> Send an error event or specific marker
> 
> Client should handle full state reset


We no just start from this.head and we don't send anything to the client, so the client doesn't know if it missed some events

BYK: I'm fine skipping events. I'm unsure if we should go to the end and not send anything or start from the beginning. I'd say we should treat this case as empty last event I'd, meaning we should start from head. 

Great catch @betegon

packages/sidecar/src/messageBuffer.ts
BYK: See https://github.com/getsentry/spotlight/pull/1104/files#r2510970166

packages/sidecar/src/parser/processEnvelope.ts
```diff
@@ -44,7 +44,7 @@ export function processEnvelope(rawEvent: RawEventContext): ParsedEnvelope | nul
     return null;
   }
 
-  envelopeHeader.__spotlight_envelope_id = uuidv7();
+  envelopeHeader.__spotlight_envelope_id = uuidv7obj();
```
BYK, line 47: >uuidv7obj() and uuidv4obj() return an object that represents a UUID as a 16-byte byte array:

Okay, why do we even need this?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -15,6 +15,117 @@ export class MessageBuffer<T> {
...
+    let targetUuid: UUID;
+    try {
+      targetUuid = UUID.parse(targetId);
+    } catch {
+      // Invalid UUID format, cannot find
+      return -1;
+    }
+
+    while (left <= right) {
+      const mid = Math.floor((left + right) / 2);
```
BYK, line 37: Shouldn't this be `Math.floor((right - left) / 2)`?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -15,6 +15,117 @@ export class MessageBuffer<T> {
...
+                  left = mid + 1;
+                } else {
+                  right = mid - 1;
+                }
+                boundsAdjusted = true;
+                break;
+              }
+            }
+          }
+        }
```
BYK, line 91: Okay, this entire block is very hard to follow. Vibe code?

packages/sidecar/src/messageBuffer.ts
```diff
@@ -15,6 +15,117 @@ export class MessageBuffer<T> {
     this.items = new Array(size);
   }
 
+  /**
+   * Binary search for an event by its envelope ID (UUIDv7).
+   * Returns the position of the event, or -1 if not found.
+   * Takes advantage of UUIDv7's time-ordered property.
+   */
+  private binarySearchEventId(targetId: string): number {
```
BYK, line 23: Let's just name this helper `findEventIdIdx`?

BYK (changes requested):

BYK (approved):
You know what, all the tests pass and they make sense. Let's address my comments in a follow-up

PR #1108, opened 2025-11-06
Feat(sidecar): Support docker compose on Spolitght run

### LINKED ISSUE #1089: Support `docker-compose` for `spotlight run`
We want `spotlight run` to support `docker-compose` out of the box. This means the following:

1. When no specific command is passed, we should check if:
   1. There's at least one of the following files in the directory we are in: `docker-compose.yaml `, `docker-compose.yml`, `compose.yml`, `compose.yaml` -- note which file exists, we'll use it later
   2. If we have a `docker-compose` or `docker` command executable in the `PATH`
   3. If we can successfully run `docker-compose version` or `docker compose version`respectively
   4. The version should be above X
2. If we pass step 1 above, we then should:
   1. Set the run command to `docker compose up` or `docker-compose up` accordingly. Preference should be for `docker-compose`. Follow the logic in https://github.com/getsentry/self-hosted/blob/6ae40d837a78c2968a28acf952b1beb9911237f8/install/dc-detect-version.sh#L33
   2. Extend the command in step 2.1 above with `-f <filename_from_step_1.1>`
   3. Based on the file name in step 1.1 above, check if a file with the suffix `.override.yml` or `.override.yaml` based on the original filename. If it does, note the file name.
   4. If a matching file is found on step 2.3 above, extend the command from 2.2 with `-f <filename_from_step_2.3>`
   5. Parse the file from 1.1 (you'll need a YAML parser), get all the key names under `services` and note this down
   6. Generate a yaml file in memory, using all the keys from step 2.5 above with the following format:
      ```yaml
      services:
        <key1>:
          environment:
            - SENTRY_SPOTLIGHT
            - SENTRY_TRACES_SAMPLE_RATE
        <key2>:
          environment:
            - SENTRY_SPOTLIGHT
            - SENTRY_TRACES_SAMPLE_RATE
        ...
        <keyN>:
          environment:
            - SENTRY_SPOTLIGHT
            - SENTRY_TRACES_SAMPLE_RATE
       ```
    7. Add the following to the final version of the command from step 2.4 above: `-f -` -- This is following the merge logic from https://docs.docker.com/compose/how-tos/multiple-compose-files/merge, feel free to refer to that for any corrections or debugging
    8. Push the yaml file generated in step 2.6 above as `stdin` for the process you are running ast part of `spotlight run`

betegon:
Closes #1089

Example: 

Screenshot 2025-11-06 at 22 49 40

### DIFF
--- a/packages/sidecar/src/utils/docker-compose.ts
+++ b/packages/sidecar/src/utils/docker-compose.ts
@@ -7,7 +7,7 @@ export const DOCKER_MIN_VERSION = "20.10.0";
 export const DOCKER_HOST_INTERNAL = "host.docker.internal";
 export const DOCKER_HOST_GATEWAY = "host-gateway";
 
-export interface DockerComposeConfig {
+interface DockerComposeConfig {
   composeFile: string;
   overrideFile?: string;
   command: string[];
@@ -96,7 +96,7 @@ function detectComposeCommand(): { command: string[]; version: string } | null {
 /**
  * Generate the override YAML for injecting Spotlight environment variables
  */
-export function generateSpotlightOverrideYaml(serviceNames: string[], port: number): string {
+function generateSpotlightOverrideYaml(serviceNames: string[], port: number): string {
   const services: Record<string, unknown> = {};
 
   // We use host.docker.internal to ensure the container can access the host machine


packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+  // We use host.docker.internal to ensure the container can access the host machine
+  // for backend services that need to access the host machine to send events to the Sidecar.
+  const spotlightUrl = `http://${DOCKER_HOST_INTERNAL}:${port}/stream`;
+  const publicSpotlightUrl = `http://localhost:${port}/stream`;
+
+  for (const serviceName of serviceNames) {
+    services[serviceName] = {
+      environment: [
+        `SENTRY_SPOTLIGHT=${spotlightUrl}`,
+        `NEXT_PUBLIC_SENTRY_SPOTLIGHT=${publicSpotlightUrl}`, // This is needed for Next.js
```
betegon, line 111: this will actually fail for frontend sdks that are not nextjs, as we will be passing the host internal url.

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  // We use host.docker.internal to ensure the container can access the host machine
+  // for backend services that need to access the host machine to send events to the Sidecar.
+  const spotlightUrl = `http://${DOCKER_HOST_INTERNAL}:${port}/stream`;
+  const publicSpotlightUrl = `http://localhost:${port}/stream`;
+
+  for (const serviceName of serviceNames) {
+    services[serviceName] = {
+      environment: [
+        `SENTRY_SPOTLIGHT=${spotlightUrl}`,
```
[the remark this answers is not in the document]
betegon: I know

packages/sidecar/src/cli/run.ts
```diff
@@ -102,41 +103,43 @@ export default async function run({ port, cmdArgs, basePath, filesToServe, forma
...
+    if (dockerCompose) {
+      logger.info(
+        `Detected Docker Compose project with ${dockerCompose.serviceNames.length} service(s): ${dockerCompose.serviceNames.join(", ")}`,
+      );
+      const command = buildDockerComposeCommand(dockerCompose, actualServerPort);
+      cmdArgs = command.cmdArgs;
+      dockerComposeOverride = command.dockerComposeOverride;
+    } else {
+      try {
+        const scripts = JSON.parse(readFileSync("./package.json", "utf-8")).scripts;
```
BYK, line 136: I'm not sure which one should take precedence: `package.json` or `docker-compose.yml`? Or may be neither and we should bail/throw when both exist and no command given?

packages/sidecar/src/cli/run.ts
```diff
@@ -145,19 +148,30 @@ logger.error(`The port ${port} might already be in use — most likely by anothe
...
+  if (!stdout || !stderr) {
+    logger.error("Failed to create process streams");
+    process.exit(1);
+  }
+
+  // If we have Docker Compose override YAML, write it and close stdin
+  if (dockerComposeOverride && runCmd.stdin) {
+    runCmd.stdin.write(dockerComposeOverride);
+    runCmd.stdin.end();
+  }
```
BYK, line 173: If we have `dockerComposeOverride` but not `runCmd.stdin` we should throw. (we can either simply remove the `runCmd.stdi` check here and let it throw if it's missing or handle this case explicitly)

packages/sidecar/src/utils/docker-compose-parser.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+  return null;
+}
+
+/**
+ * Find the override file for a given compose file
+ */
+export function findOverrideFile(composeFile: string): string | null {
+  const ext = composeFile.endsWith(".yaml") ? ".yaml" : ".yml";
+  const baseName = composeFile.replace(/\.(yaml|yml)$/, "");
```
BYK, line 24: Not sure if this is vibe coded but we can make this part easier by changing `COMPOSE_FILE_NAMES` above a bit like this:

```ts
const COMPOSE_FILE_BASE_NAMES = ["compose", "docker-compose"];
const COMPOSE_FILE_EXTENSIONS = [".yml", ".yaml"];
```

Then we can just loop over both and build both file names (compose file and override file)

packages/sidecar/src/utils/docker-compose-parser.ts
```diff
@@ -0,0 +1,55 @@
...
+    return overrideFile;
+  }
+
+  return null;
+}
+
+/**
+ * Parse compose file and extract service names
+ */
+export function parseComposeFile(filePath: string): string[] {
```
BYK, line 38: `getDockerComposeServices` is a better name for this helper

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+  const parts2 = v2.replace(/^v/, "").split(".").map(Number);
+
+  for (let i = 0; i < Math.max(parts1.length, parts2.length); i++) {
+    const p1 = parts1[i] || 0;
+    const p2 = parts2[i] || 0;
+    if (p1 > p2) return true;
+    if (p1 < p2) return false;
+  }
+  return true;
+}
```
BYK, line 32: https://www.npmjs.com/package/semver-compare or https://www.npmjs.com/package/semver which is updated and used way more frequently.

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
+import { execSync } from "node:child_process";
+import { stringify as stringifyYaml } from "yaml";
+import { logger } from "../logger.js";
+import { findComposeFile, findOverrideFile, parseComposeFile } from "./docker-compose-parser.js";
+
+export const DOCKER_MIN_VERSION = "20.10.0";
```
BYK, line 6: Love these constants. Can we add a comment here why we need this minimum version? (I got confused at first but then remember it's due t' `host.docker.internal` availability.

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  // Use standalone if it's newer, otherwise prefer plugin
+  const pluginVer = composePluginVersion!.replace(/^v/, "");
+  const standaloneVer = composeStandaloneVersion!.replace(/^v/, "");
+
+  if (isVersionGreaterOrEqual(standaloneVer, pluginVer)) {
+    return { command: ["docker-compose"], version: composeStandaloneVersion! };
+  }
+
+  return { command: ["docker", "compose"], version: composePluginVersion! };
```
BYK, line 93: Quite sure you (or Cursor) can simplify the branching here a bit as some conditions are mutually exclusive or covers the other :)

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * Generate the override YAML for injecting Spotlight environment variables
+ */
+function generateSpotlightOverrideYaml(serviceNames: string[], port: number): string {
+  const services: Record<string, unknown> = {};
+
+  // We use host.docker.internal to ensure the container can access the host machine
+  // for backend services that need to access the host machine to send events to the Sidecar.
+  const spotlightUrl = `http://${DOCKER_HOST_INTERNAL}:${port}/stream`;
+  const publicSpotlightUrl = `http://localhost:${port}/stream`;
```
BYK, line 105:
```suggestion
  const spotlightUrlForBrowser = `http://localhost:${port}/stream`;
```

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  return { command: ["docker", "compose"], version: composePluginVersion! };
+}
+
+/**
+ * Generate the override YAML for injecting Spotlight environment variables
+ */
+function generateSpotlightOverrideYaml(serviceNames: string[], port: number): string {
+  const services: Record<string, unknown> = {};
```
BYK, line 100:
```suggestion
  const services: Record<string, {"environment": string[], "extra_hosts": string[]}> = {};
```

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+  // for backend services that need to access the host machine to send events to the Sidecar.
+  const spotlightUrl = `http://${DOCKER_HOST_INTERNAL}:${port}/stream`;
+  const publicSpotlightUrl = `http://localhost:${port}/stream`;
+
+  for (const serviceName of serviceNames) {
+    services[serviceName] = {
+      environment: [
+        `SENTRY_SPOTLIGHT=${spotlightUrl}`,
+        `NEXT_PUBLIC_SENTRY_SPOTLIGHT=${publicSpotlightUrl}`, // This is needed for Next.js
+        "SENTRY_TRACES_SAMPLE_RATE=1",
```
BYK, line 112: Don't set these here please, just add them to the array without values, which should pass the env variables set by the higher up `run` command. Otherwise this part and the "main" part in the `run` command will diverge eventually without anyone noticing.

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+    cmdArgs.push("-f", config.overrideFile);
+  }
+
+  cmdArgs.push("-f", "-");
+
+  cmdArgs.push("up");
+
+  const dockerComposeOverride = generateSpotlightOverrideYaml(config.serviceNames, port);
+
+  return { cmdArgs, dockerComposeOverride };
```
BYK, line 142: To make this more generic (in case we support other commands similar to this in the future), I'd make the return values `cmdArgs` and `stdin`

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+ */
+export function detectDockerCompose(): DockerComposeConfig | null {
+  const docker = detectDocker();
+  if (!docker) {
+    logger.debug("Docker is not installed or not in PATH");
+    return null;
+  }
+
+  if (!docker.valid) {
+    return null;
```
BYK, line 159: Where's my "need minimum docker version X" error message?

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  logger.debug(`Detected Docker Compose version ${compose.version} (${compose.command.join(" ")})`);
+
+  const composeFile = findComposeFile();
+  if (!composeFile) {
+    logger.debug("No compose file found in current directory");
+    return null;
+  }
```
BYK, line 176: Check this before doing the others above as they are way more expensive (at least 2 execs vs one existence check)

packages/sidecar/src/utils/docker-compose.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  logger.debug(`Found compose file: ${composeFile}`);
+
+  const serviceNames = parseComposeFile(composeFile);
```
BYK, line 180: If we are reading the file right away, we can save one I/O operation (which really counts) by simply trying to read each possible file and if all raises `E_NOENT` we know we don't have the file.

packages/sidecar/src/cli/run.ts
betegon: followed up on #1120

PR #1111, opened 2025-11-07
Fix(sidecar): Unknown type event error

betegon:
This error what was exiting the program;
Screenshot 2025-11-07 at 15 28 04

This fixes it and also shows a better human output for ISO and UNIX timestamps

Before: 
Screenshot 2025-11-07 at 15 29 23

Now:
Screenshot 2025-11-07 at 15 32 00

### DIFF
--- a/.changeset/calm-pears-fix.md
+++ b/.changeset/calm-pears-fix.md
@@ -1,0 +1,5 @@
+---
+"@spotlightjs/sidecar": patch
+---
+
+Fix breaking error when event type is not supported


packages/sidecar/src/formatters/human/utils.ts
```diff
@@ -123,8 +61,24 @@ export function inferEnvelopeSource(envelopeHeader: Envelope[0], event?: any): "
...
+    date = new Date(timestamp);
+  } else {
+    // Handle Unix timestamp
+    date = new Date(timestamp * 1000);
+  }
+
+  if (Number.isNaN(date.getTime())) {
+    // placeholder with same width as valid timestamp for alignment in the logs
+    return "??:??:??";
+  }
```
BYK, line 80: <3

packages/sidecar/src/formatters/index.ts
```diff
@@ -1,6 +1,30 @@
...
+export function applyFormatter<K extends keyof FormatterRegistry>(
+  registry: FormatterRegistry,
+  eventType: K,
+  payload: EnvelopeItem[1],
+  envelopeHeader: Envelope[0],
+): string[] {
+  const entry = registry[eventType];
+  const event = payload as SentryEvent;
+  if (!entry.typeGuard(event)) {
+    logger.warn(`Skipping event: type guard failed for ${eventType} (event.type=${(event as any).type})`);
```
BYK, line 25: It'd be good to dump the event using `logger.debug(utils.inspect(event))` and/or sending this to Sentry for inspection

packages/sidecar/src/parser/helpers.ts
```diff
@@ -18,7 +18,10 @@ export const SUPPORTED_EVENT_TYPES = new Set([
 ]);
 
 export function isErrorEvent(event: SentryEvent): event is SentryErrorEvent {
-  return (!event.type || ERROR_EVENT_TYPES.has(event.type)) && Boolean((event as SentryErrorEvent).exception);
+  const hasValidType = !event.type || ERROR_EVENT_TYPES.has(event.type);
+  const hasException = Boolean((event as SentryErrorEvent).exception);
+  const hasMessage = Boolean((event as SentryErrorEvent).message);
+  return hasValidType && (hasException || hasMessage);
```
BYK, line 24: Keep in mind that this version is less performant than the one you replaced as it doesn't "short circut". Might become an issue under heavy load, otherwise impact should be insignificant.

PR #1112, opened 2025-11-07
Improvement(sidecar): Beautify CLI human formatter

betegon:
Improved human formatter.
- Caps for event types
- Brackets for event types
- Bracket for log attributes

Before:

Screenshot 2025-11-07 at 19 53 35

After:

Screenshot 2025-11-07 at 19 48 07

### DIFF
--- a/packages/sidecar/src/formatters/human/utils.ts
+++ b/packages/sidecar/src/formatters/human/utils.ts
@@ -144,8 +144,24 @@ export function inferEnvelopeSource(envelopeHeader: Envelope[0], event?: any): S
 /**
  * Format timestamp as local time HH:MM:SS
  */
-export function formatLocalTime(timestamp?: number): string {
-  const date = timestamp ? new Date(timestamp * 1000) : new Date();
+export function formatLocalTime(timestamp?: number | string): string {
+  let date: Date;
+
+  if (!timestamp) {
+    date = new Date();
+  } else if (typeof timestamp === "string") {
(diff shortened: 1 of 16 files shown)


packages/sidecar/src/formatters/human/utils.ts
```diff
@@ -139,8 +139,8 @@ export function padLabel(label: string, width: number): string {
 }
 
 // Constants for consistent padding
-export const SOURCE_WIDTH = 7; // "browser" is the longest
-export const TYPE_WIDTH = 7; // "warning" is the longest
+export const SOURCE_WIDTH = 9; // "[BROWSER]" is the longest
```
BYK: If you did
```

Math.max(values.map(v=>v.length));

```
or even `"[BROWSER]".length` it would be more self-explanatory and maintainable

packages/sidecar/src/formatters/human/utils.ts
```diff
@@ -153,23 +153,28 @@ export function colorizeTime(time: string): string {
  * Colorize source based on envelope source
  */
 export function colorizeSource(source: string): string {
+  const uppercaseSource = source.toUpperCase();
```
BYK: Not sure if I actually like the upper case source. It looks  especially next to fully upper case level/type.

Maybe pull the type before the source and keep the lower case but bracketed?

BYK (approved):

PR #1121, opened 2025-11-11
ref(sidecar): Docker compose refactor

betegon:
amongst some things, this includes:
- setting up the env vars on the run command instead of in the docker compose specific code.
- renaming variables
- simplify some logic
-

### DIFF
--- a/packages/spotlight/src/server/utils/docker-compose-parser.ts
+++ b/packages/spotlight/src/server/utils/docker-compose-parser.ts
@@ -7,24 +7,6 @@ const COMPOSE_FILE_BASE_NAMES = ["docker-compose", "compose"] as const;
 const COMPOSE_FILE_EXTENSIONS = [".yml", ".yaml"] as const;
 
 /**
- * Find a file by trying all combinations of base names and extensions.
- */
-function findFile(buildFileName: (baseName: string, ext: string) => string): string | null {
-  for (const baseName of COMPOSE_FILE_BASE_NAMES) {
-    for (const ext of COMPOSE_FILE_EXTENSIONS) {
-      const fileName = buildFileName(baseName, ext);
-      try {
-        accessSync(fileName, constants.R_OK);
-        return fileName;
-      } catch {
-        // File doesn't exist or isn't readable, continue searching
-      }
-    }
-  }
-  return null;
-}
-
-/**
  * Get compose files from COMPOSE_FILE environment variable
  * Returns an array of file paths if the variable is set, null otherwise
  */
@@ -44,16 +26,31 @@ export function getComposeFilesFromEnv(): string[] | null {
 }
 
 /**
- * Find the compose file in the current directory
+ * Find the compose file in the current directory by trying all combinations
+ * of base names and extensions.
  */
 export function findComposeFile(): string | null {
-  return findFile((baseName, ext) => `${baseName}${ext}`);
+  for (const baseName of COMPOSE_FILE_BASE_NAMES) {
+    for (const ext of COMPOSE_FILE_EXTENSIONS) {
+      const fileName = `${baseName}${ext}`;
+      try {
+        accessSync(fileName, constants.R_OK);
+        return fileName;
+      } catch {
+        // File doesn't exist or isn't readable, continue searching
+      }
+    }
+  }
+  return null;
 }
 
 /**
- * Find the override file for a given compose file
+ * Find the override file for a given compose file.
+ * Docker Compose looks for an override file with the same base name and extension.
+ * For example: docker-compose.yml -> docker-compose.override.yml
  */
 export function findOverrideFile(composeFile: string): string | null {
+  // Extract extension from the original file to maintain consistency
   const ext = composeFile.endsWith(".yaml") ? ".yaml" : ".yml";
   const baseName = composeFile.replace(/\.(yaml|yml)$/, "");
   const overrideFile = `${baseName}.override${ext}`;


packages/spotlight/src/server/cli/run.ts
```diff
@@ -139,9 +139,14 @@ export default async function run({ port, cmdArgs, basePath, filesToServe, forma
       logger.info(
         `Detected Docker Compose project with ${dockerCompose.serviceNames.length} service(s): ${dockerCompose.serviceNames.join(", ")}`,
       );
-      const command = buildDockerComposeCommand(dockerCompose, actualServerPort);
+      // Use host.docker.internal for backend services to access the host machine
+      env.SENTRY_SPOTLIGHT = `http://host.docker.internal:${actualServerPort}/stream`;
+      // Keep localhost for browser-side (Next.js public env vars)
+      env.NEXT_PUBLIC_SENTRY_SPOTLIGHT = `http://localhost:${actualServerPort}/stream`;
```
BYK: We do not need to set this again?
```suggestion
```

packages/spotlight/src/server/cli/run.ts
```diff
@@ -139,9 +139,14 @@ export default async function run({ port, cmdArgs, basePath, filesToServe, forma
...
+
+      const command = buildDockerComposeCommand(dockerCompose);
       cmdArgs = command.cmdArgs;
-      dockerComposeOverride = command.dockerComposeOverride;
+      dockerComposeOverride = command.stdin;
```
BYK: We should name this `stdin` too?

packages/spotlight/src/server/cli/run.ts
```diff
@@ -178,7 +183,11 @@ export default async function run({ port, cmdArgs, basePath, filesToServe, forma
   }
 
   // If we have Docker Compose override YAML, write it and close stdin
```
BYK:
```suggestion
  // If our command has a stdin input _and_ we can send to stdin, write and close
```

packages/spotlight/src/server/utils/docker-compose-parser.ts
```diff
@@ -2,40 +2,49 @@ import { existsSync, readFileSync } from "node:fs";
...
  */
-export function findComposeFile(): string | null {
-  for (const fileName of COMPOSE_FILE_NAMES) {
-    if (existsSync(fileName)) {
-      return fileName;
+function findFile(buildFileName: (baseName: string, ext: string) => string): string | null {
+  for (const baseName of COMPOSE_FILE_BASE_NAMES) {
+    for (const ext of COMPOSE_FILE_EXTENSIONS) {
+      const fileName = buildFileName(baseName, ext);
+      if (existsSync(fileName)) {
```
BYK: nit: might be better to use `fs.accessSync()` with mode `fs.constants.R_OK` to ensure readability as we'll definitely need to read it.

packages/spotlight/src/server/utils/docker-compose-parser.ts
```diff
@@ -1,41 +1,63 @@
...
+      const fileName = buildFileName(baseName, ext);
+      try {
+        readFileSync(fileName);
+        return fileName;
+      } catch (error) {
+        if ((error as NodeJS.ErrnoException).code === "ENOENT") {
+          continue;
+        }
+        logger.error(`Error checking file ${fileName}: ${error}`);
+      }
```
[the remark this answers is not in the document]
BYK: Not sure if you changed your code after this comment @betegon but I think this was inaccurate as we should return as soon as we can read the contents and we should return both the file name _and_ the contents (as we'll open the file to get the contents and parse out the services anyway)

packages/spotlight/src/server/utils/docker-compose-parser.ts
```diff
@@ -2,40 +2,49 @@ import { existsSync, readFileSync } from "node:fs";
...
   const baseName = composeFile.replace(/\.(yaml|yml)$/, "");
-
   const overrideFile = `${baseName}.override${ext}`;
 
   if (existsSync(overrideFile)) {
     return overrideFile;
   }
-
   return null;
 }
```
BYK, line 65: I think this needs to use `findFile` too? Otherwise `findFile` taking a builder function as a parameter makes no sense to me.

packages/spotlight/src/server/cli/run.ts
```diff
@@ -190,9 +194,11 @@ export default async function run({ port, cmdArgs, basePath, filesToServe, forma
         logger.info(
           `Detected Docker Compose project with ${dockerCompose.serviceNames.length} service(s): ${dockerCompose.serviceNames.join(", ")}`,
         );
-        const command = buildDockerComposeCommand(dockerCompose, actualServerPort);
+        // Use host.docker.internal for backend services to access the host machine
+        env.SENTRY_SPOTLIGHT = getSpotlightURL(actualServerPort, DOCKER_HOST);
+        const command = buildDockerComposeCommand(dockerCompose);
         cmdArgs = command.cmdArgs;
-        dockerComposeOverride = command.dockerComposeOverride;
+        stdin = command.stdin;
```
BYK, line 201: Follow up: let's consolidate this logic with the `else if (dockerCompose)` part as we just repeat the same logic there.

packages/spotlight/src/server/cli/run.ts
```diff
@@ -241,9 +250,13 @@ export default async function run({ port, cmdArgs, basePath, filesToServe, forma
     process.exit(1);
   }
 
-  // If we have Docker Compose override YAML, write it and close stdin
-  if (dockerComposeOverride && runCmd.stdin) {
-    runCmd.stdin.write(dockerComposeOverride);
+  // If our command has a stdin input _and_ we can send to stdin, write and close
+  if (stdin) {
+    if (!runCmd.stdin) {
+      logger.error("Failed to pipe Docker Compose override: stdin is not available");
```
BYK, line 256: We should not use Docker-specific language here as it might be a different command in the future.

packages/spotlight/src/server/utils/docker-compose-parser.ts
```diff
@@ -25,37 +26,48 @@ export function getComposeFilesFromEnv(): string[] | null {
...
+      const fileName = `${baseName}${ext}`;
+      try {
+        accessSync(fileName, constants.R_OK);
+        return fileName;
+      } catch {
+        // File doesn't exist or isn't readable, continue searching
+      }
     }
   }
   return null;
```
BYK, line 44: Just musing: if you are okay with `async` we can actually do `Promise.race(files.map(f => access(f, constants.R_OK))` or something like that and check all in parallel 😅

packages/spotlight/src/server/utils/docker-compose-parser.ts
```diff
@@ -25,37 +26,48 @@ export function getComposeFilesFromEnv(): string[] | null {
...
 
 /**
- * Find the override file for a given compose file
+ * Find the override file for a given compose file.
+ * Docker Compose looks for an override file with the same base name and extension.
+ * For example: docker-compose.yml -> docker-compose.override.yml
  */
 export function findOverrideFile(composeFile: string): string | null {
+  // Extract extension from the original file to maintain consistency
   const ext = composeFile.endsWith(".yaml") ? ".yaml" : ".yml";
```
BYK, line 54: We should just use `COMPOSE_FILE_EXTENSIONS` here? Also I think it is fair game to have `docker-compose.yaml` along with `docker-compose.override.yml` so the extensions shouldn't necessarily match.

PR #1131, opened 2025-11-13
Implement compose file environment variable support

BYK:
Addresses https://github.com/getsentry/spotlight/issues/1089#issuecomment-3524334445

This PR implements support for the `COMPOSE_FILE` environment variable, allowing users to specify one or more Docker Compose files via an environment variable, aligning with native Docker Compose behavior.

Key changes:
*   **Prioritize `COMPOSE_FILE`**: Spotlight now checks the `COMPOSE_FILE` environment variable first to determine which compose files to use.
*   **Docker Compose behavior alignment**: When `COMPOSE_FILE` is set, automatic detection of `docker-compose.override.yml` is skipped, matching native Docker Compose behavior.
*   **Multiple file support**: Supports multiple compose files specified in `COMPOSE_FILE` (separated by `:` on Unix, `;` on Windows).
*   **Environment cleanup**: The `COMPOSE_FILE` variable is unset in the spawned Docker Compose process to prevent conflicts with explicitly passed `-f` flags.

### DIFF
--- a/.changeset/compose-file-env-support.md
+++ b/.changeset/compose-file-env-support.md
@@ -1,0 +1,5 @@
+---
+"@spotlightjs/spotlight": minor
+---
+
+Support COMPOSE_FILE environment variable for Docker Compose projects
--- a/.changeset/soft-hairs-sip.md
+++ b/.changeset/soft-hairs-sip.md
@@ -1,5 +1,5 @@
 ---
-"@spotlightjs/sidecar": minor
+"@spotlightjs/spotlight": minor
 ---
 
 Prompt user to choose between docker compose and package.json when both are present


packages/spotlight/src/server/utils/docker-compose-parser.ts
```diff
@@ -1,9 +1,29 @@
...
 import { logger } from "../logger.ts";
 
 const COMPOSE_FILE_NAMES = ["docker-compose.yaml", "docker-compose.yml", "compose.yml", "compose.yaml"] as const;
 
+/**
+ * Get compose files from COMPOSE_FILE environment variable
+ * Returns an array of file paths if the variable is set, null otherwise
+ */
+export function getComposeFilesFromEnv(): string[] | null {
+  const composeFileEnv = process.env.COMPOSE_FILE;
```
betegon, line 13: the `COMPOSE_FILE` env is loaded when the docker-compose starts as it'd be part of our user's project, so it will be always empty.

Why could check if there's a `.env` file in the directory and search for `COMPOSE_FILE`

betegon (changes requested):

packages/spotlight/src/server/utils/docker-compose-parser.ts
betegon: talked offline: if this is necessary we'll do it in another PR, for now we'll assume the variable is already set

betegon (approved):
merging thiis!

PR #1133, opened 2025-11-18
ref(electron): Overhaul electron build process

BYK:
- Replace electron-vite with vite-plugin-electron for better flexibility
- Merge electron.html and index.html into a single index.html
- Enable @sentry/react on electron too
- Update electron main to use VITE_DEV_SERVER_URL instead of ELECTRON_RENDERER_URL
- Update dev:electron and build:electron scripts to use standard Vite

Benefits:
- Entry point configured in vite config, not package.json
- Standard Vite workflow and configuration
- Better compatibility with npm exports field for dual-purpose package
- More control over build process

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -160,9 +160,34 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: electron-build
-          path: packages/spotlight/dist-electron/
+          path: ${{ runner.temp }}/electron-artifact/
           github-token: ${{ secrets.GITHUB_TOKEN }}
           run-id: ${{ github.event.workflow_run.id }}
+
+      - name: Validate and copy Electron build files
+        run: |
+          # Create/empty the target directory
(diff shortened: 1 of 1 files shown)


packages/spotlight/src/ui/lib/instrumentation.ts
```diff
@@ -30,6 +31,29 @@ export default function initSentry() {
     integrations.push(Sentry.spotlightBrowserIntegration({ sidecarUrl }));
   }
 
+  // For Electron, use the combined SDK pattern for better observability
+  // This gives us Electron-specific error handling PLUS React features
+  if (isElectron) {
+    // Dynamically import the Electron SDK
+    import("@sentry/electron/renderer").then(ElectronSentry => {
```
MathurAditya724: Question - Don't we have to load sentry before the app starts?

BYK: Yes, fix on the way. Great catch.

PR #1140, opened 2025-11-20
feat: added spotlight SDK

### LINKED ISSUE #1094: Proposal: Extract core functionality into @spotlightjs/core package
**Problem**

Currently, building custom Spotlight-compatible tools (like the MCP/Agents Inspector) requires copying code from the sidecar and overlay packages. The core utilities for buffer management, envelope parsing, and streaming are tightly coupled to the existing implementation and specific frameworks (Hono).

**Motivation**

Developers should be able to:

- Build custom sidecar implementations with their preferred framework (Express, Fastify, etc.)
- Create custom overlay/client applications
- Integrate Spotlight's protocol into their own tools
- Reuse encoding/decoding logic without duplication

**Proposal**

Extract the core, framework-agnostic functionality into a new `@spotlightjs/core` package that provides:

1. **Buffer Management** - MessageBuffer with pub/sub pattern for event storage and streaming
2. **Envelope Parsing** - Pure functions for encoding/decoding Sentry envelopes
3. **Event Container** - Lazy-parsing wrapper for efficient event handling
4. **Streaming Utilities** - Framework-agnostic helpers for SSE and client connections
5. **Shared Types & Constants** - Common types, content types, and event definitions

**Benefits**

- **Framework flexibility**: Use with any Node.js framework
- **Code reuse**: No more copy-pasting for custom implementations
- **Better architecture**: Clear separation between protocol and framework
- **Ecosystem growth**: Enable third-party tools and integrations
- **Easier testing**: Test core logic independently

**High-Level Approach**

The package would provide pure functions and classes with zero framework dependencies. The existing sidecar and overlay would migrate to use this package, maintaining backward compatibility. Framework integration examples (Hono, Express, Fastify) would demonstrate usage patterns.

MathurAditya724:
Added new Spotlight SDK for helping others to build on top of Spotlight

Closes #1094.

### DIFF
--- a/packages/spotlight/src/server/routes/__tests__/server.test.ts
+++ b/packages/spotlight/src/server/routes/__tests__/server.test.ts
@@ -103,7 +103,7 @@ describe("encoded envelopes", () => {
 
   it("gzip", testEncodedEnvelope("gzip"));
   it("deflate", testEncodedEnvelope("deflate"));
-  it("brotli", testEncodedEnvelope("brotli"));
+  it("br", testEncodedEnvelope("br"));
 });
 
 describe("mcp", () => {


.changeset/breezy-words-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spotlightjs/spotlight": patch
```
BYK:
```suggestion
"@spotlightjs/spotlight": minor
```

This is a new feature :)

packages/spotlight/vite.node.config.ts
```diff
@@ -15,6 +15,7 @@ export default defineConfig({
       entry: {
         run: resolve(__dirname, "src/run.ts"),
         "server/main": resolve(__dirname, "src/server/main.ts"),
+        "server/sdk": resolve(__dirname, "src/server/sdk.ts"),
```
BYK, line 18: Why not just `SDK`?

BYK (commented):
The AI reviewers' concerns seem legit

packages/spotlight/vite.node.config.ts
MathurAditya724: Isn't this just about compilation, actual export definition is in package.json

BYK: Oh, right sorry.

packages/spotlight/src/server/sdk.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export function pushToSpotlightBuffer(options: PushToSpotlightBufferOptions) {
+  const body = options.overrideBufferDecoding
+    ? options.body
+    : decompressBody(options.body, options.headers.contentEncoding);
+
+  let contentType = options.headers.contentType?.split(";")[0].toLocaleLowerCase();
+  if (options.query.sentry_client?.startsWith("sentry.javascript.browser") && options.headers.origin) {
+    // This is a correction we make as Sentry Browser SDK may send messages with text/plain to avoid CORS issues
+    contentType = "application/x-sentry-envelope";
```
BYK: We should move this check outside as it is specific to the server implementation. It has no place in the SDK API.

packages/spotlight/src/server/sdk.ts
```diff
@@ -0,0 +1,62 @@
...
+  };
+  query: {
+    sentry_client?: string;
+  };
+};
+
+export function pushToSpotlightBuffer(options: PushToSpotlightBufferOptions) {
+  const body = options.overrideBufferDecoding
+    ? options.body
+    : decompressBody(options.body, options.headers.contentEncoding);
```
BYK: Let's just not pass `contentEncoding` here.

packages/spotlight/src/server/sdk.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export function createSpotlightBuffer() {
+  return new MessageBuffer<EventContainer>();
+}
+
+type PushToSpotlightBufferOptions = {
+  spotlightBuffer: MessageBuffer<EventContainer>;
+  body: Buffer;
+  overrideBufferDecoding?: boolean;
+  headers: {
```
BYK: This API interface should not get server-specific stuff like `headers` or `query`. We should not pass them in the first place if possible or pass normalized/derived values to help with anything else. That said I see the value of a simpler interface so I guess it can stay this way. My ideal version would have had `source` field calculated before being passed to this method, but to calculate that we need both the user agent and the decoded envelope. Still, let's just use `userAgent` and `contentType` outside of `headers`, make `userAgent` optional, and keep server-specific stuff to a minimum with all of them being optional with sensible defaults.

packages/spotlight/src/server/sdk.ts
```diff
@@ -15,48 +15,36 @@ export function createSpotlightBuffer() {
...
-  headers: {
-    contentEncoding?: string;
-    contentType?: string;
-    userAgent?: string;
-    origin?: string;
-  };
-  query: {
-    sentry_client?: string;
-  };
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
BYK: What's going on here? 😅

packages/spotlight/src/server/sdk.ts
```diff
@@ -15,48 +15,36 @@ export function createSpotlightBuffer() {
...
+  userAgent?: string;
+  origin?: string;
+  sentryClient?: string;
 };
 
 export function pushToSpotlightBuffer(options: PushToSpotlightBufferOptions) {
-  const body = options.overrideBufferDecoding
-    ? options.body
-    : decompressBody(options.body, options.headers.contentEncoding);
+  const body = decompressBody(options.body, options.encoding);
```
BYK, line 25: Let's add a little comment here about why we're doing decompression manually. (To use body below without another decompression)

packages/spotlight/src/server/sdk.ts
```diff
@@ -15,48 +15,36 @@ export function createSpotlightBuffer() {
...
 
-  let contentType = options.headers.contentType?.split(";")[0].toLocaleLowerCase();
-  if (options.query.sentry_client?.startsWith("sentry.javascript.browser") && options.headers.origin) {
+  let contentType = options.contentType?.split(";")[0].toLocaleLowerCase();
+  if (options.sentryClient?.startsWith("sentry.javascript.browser") && options.origin) {
```
BYK: This logic should be moved out of the SDK as it's specifically a server hack we do. That will allow you to remove origin etc from the props too.

packages/spotlight/src/server/sdk.ts
MathurAditya724: eslint is giving type error for not using `{}`, but we need this for supporting string variables

packages/spotlight/src/server/sdk.ts
MathurAditya724: Done

packages/spotlight/src/server/sdk.ts
```diff
@@ -19,18 +19,12 @@ type PushToSpotlightBufferOptions = {
...
 };
 
 export function pushToSpotlightBuffer(options: PushToSpotlightBufferOptions) {
   const body = decompressBody(options.body, options.encoding);
 
-  let contentType = options.contentType?.split(";")[0].toLocaleLowerCase();
-  if (options.sentryClient?.startsWith("sentry.javascript.browser") && options.origin) {
-    // This is a correction we make as Sentry Browser SDK may send messages with text/plain to avoid CORS issues
-    contentType = "application/x-sentry-envelope";
-  }
```
BYK: I meant put this code to the server side, not delete it 😆  -- I'll update your patch and merge it afterwards 😄

packages/spotlight/src/server/sdk.ts
```diff
@@ -0,0 +1,46 @@
...
+    return container;
+  }
+}
+
+export function decompressBody(body: Buffer, contentEncoding?: ContentEncoding) {
+  if (!contentEncoding) {
+    return body;
+  }
+  return decompressors[contentEncoding](body);
+}
```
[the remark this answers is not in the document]
BYK: This is intentional

PR #1143, opened 2025-11-27
fix(spotlight): Disable Sentry for development environment

betegon:
Having it enabled for the development environment was causing a lot of noise.

### DIFF
--- a/packages/spotlight/src/electron/main/index.ts
+++ b/packages/spotlight/src/electron/main/index.ts
@@ -102,6 +102,11 @@ app.on("ready", () => {
 Sentry.init({
   ...sentryBaseConfig,
   dsn: "https://192df1a78878de014eb416a99ff70269@o1.ingest.sentry.io/4506400311934976",
+  integrations: [
+    Sentry.consoleLoggingIntegration({
+      levels: ["log", "info", "warn", "error", "debug"],
+    }),
+  ],
   beforeSend: askForPermissionToSendToSentry,
 });
 
--- a/packages/spotlight/src/ui/lib/instrumentation.ts
+++ b/packages/spotlight/src/ui/lib/instrumentation.ts
@@ -22,6 +22,9 @@ export function getIntegrations(instrumentPageLoad = true): Integration[] {
     }),
     Sentry.replayIntegration(),
     Sentry.browserProfilingIntegration(),
+    Sentry.consoleLoggingIntegration({
+      levels: ["log", "info", "warn", "error", "debug"],
+    }),
   ];
 
   const hash = document.location.hash.slice(1);


packages/spotlight/src/instrument.ts
```diff
@@ -54,6 +54,7 @@ if (spotlightEnv && instancePort !== 0) {
 
 const sentry = init({
   dsn: "https://51bcd92dba1128934afd1c5726c84442@o1.ingest.us.sentry.io/4508404727283713",
+  enabled: process.env.NODE_ENV !== "development",
```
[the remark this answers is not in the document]
betegon: this was actually intentional but it's true that we setup the environment to `development` when there's no `NODE_ENV`, so i'll take it.

packages/spotlight/src/electron/main/index.ts
```diff
@@ -100,6 +100,7 @@ app.on("ready", () => {
 
 Sentry.init({
   dsn: "https://192df1a78878de014eb416a99ff70269@o1.ingest.sentry.io/4506400311934976",
+  enabled: !!process.env.NODE_ENV && process.env.NODE_ENV !== "development",
```
BYK: Why not `process.env.NODE_ENV && process.env.NODE_ENV !== "development"`?

packages/spotlight/src/ui/lib/instrumentation.ts
```diff
@@ -40,6 +40,9 @@ export default function initSentry() {
 
   // Web UI initialization
   const sentryClient = Sentry.init({
+    // Only enable Sentry when NODE_ENV is explicitly set and not "development"
+    // (treat undefined as development to avoid accidental telemetry)
+    enabled: !!process.env.NODE_ENV && process.env.NODE_ENV !== "development",
```
BYK: Same as above

packages/spotlight/src/instrument.ts
```diff
@@ -54,6 +54,7 @@ if (spotlightEnv && instancePort !== 0) {
 
 const sentry = init({
   dsn: "https://51bcd92dba1128934afd1c5726c84442@o1.ingest.us.sentry.io/4508404727283713",
+  enabled: !!process.env.NODE_ENV && process.env.NODE_ENV !== "development",
```
BYK: Wow, we should probably generate a common Sentry SDK init config first and then just amend it. Repeating this 3 times is silly.

packages/spotlight/src/electron/main/index.ts
betegon: because it's inferred as `"" | boolean | undefined` but enabled should be just `boolean | undefined` 🤷

BYK: Ah 😓

BYK: You can alternatively do `Boolean(process.env.NODE_ENV)` btw

betegon: oh, let's do the casting. thanks, much more explicit

packages/spotlight/src/instrument.ts
betegon: doneee

packages/spotlight/src/instrument.ts
```diff
@@ -53,13 +54,9 @@ if (spotlightEnv && instancePort !== 0) {
...
 const sentry = init({
+  ...sentryBaseConfig,
   dsn: "https://51bcd92dba1128934afd1c5726c84442@o1.ingest.us.sentry.io/4508404727283713",
-  enabled: !!process.env.NODE_ENV && process.env.NODE_ENV !== "development",
-  environment: process.env.NODE_ENV || "development",
-  release: `spotlight@${process.env.npm_package_version}`,
   debug: Boolean(process.env.SENTRY_DEBUG),
-
-  tracesSampleRate: 1,
   enableLogs: true,
```
BYK: The following 2 lines should also go into the base config, right?

BYK (approved):

packages/spotlight/src/instrument.ts
betegon: we only had logs enabled in the server, but just added them to the other platforms.

the disable spotlight only makes sense on the server, as it's for not sending spotlight errors to users by pushing to localhost:8969/stream or whatever local url it is.

packages/spotlight/src/ui/lib/instrumentation.ts
```diff
@@ -22,6 +22,9 @@ export function getIntegrations(instrumentPageLoad = true): Integration[] {
     }),
     Sentry.replayIntegration(),
     Sentry.browserProfilingIntegration(),
+    Sentry.consoleLoggingIntegration({
```
BYK, line 25: Pretty sure this will collide with enableLogs

betegon: wdym? the base config enables logs and here we setup this integration, and the same for the other instrumentations.

I followed https://docs.sentry.io/platforms/javascript/guides/react/logs/#console-logging-integration

BYK: Ugh you're right. I thought that integration was turned on automatically

PR #1149, opened 2025-12-01
build: upload sourcemaps to sentry

MathurAditya724:


packages/spotlight/vite.node.config.ts
```diff
@@ -1,58 +1,78 @@
...
+  if (!isDev) {
+    env = process.env as Record<string, string>;
+  }
+
+  return {
+    plugins: [
+      dtsPlugin,
+      sentryVitePlugin({
+        org: env.MAIN_VITE_SENTRY_ORG,
+        project: env.MAIN_VITE_UI_SENTRY_PROJECT,
```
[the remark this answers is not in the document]
MathurAditya724: we use the same project for both CLI and UI so this is fine

.github/workflows/build.yml
```diff
@@ -83,6 +83,18 @@ jobs:
...
+        env:
+          MAIN_VITE_SENTRY_ORG: ${{ vars.MAIN_VITE_SENTRY_ORG }}
+          MAIN_VITE_SENTRY_PROJECT: ${{ vars.MAIN_VITE_SENTRY_PROJECT }} # Electron App
+          MAIN_VITE_UI_SENTRY_PROJECT: ${{ vars.MAIN_VITE_UI_SENTRY_PROJECT }} # Spotlight UI
+          MAIN_VITE_SENTRY_AUTH_TOKEN: ${{ secrets.MAIN_VITE_SENTRY_AUTH_TOKEN }}
+        run: |
+          echo "MAIN_VITE_SENTRY_ORG=$MAIN_VITE_SENTRY_ORG" >> $GITHUB_ENV
+          echo "MAIN_VITE_SENTRY_PROJECT=$MAIN_VITE_SENTRY_PROJECT" >> $GITHUB_ENV
+          echo "MAIN_VITE_UI_SENTRY_PROJECT=$MAIN_VITE_UI_SENTRY_PROJECT" >> $GITHUB_ENV
+          echo "MAIN_VITE_SENTRY_AUTH_TOKEN=$MAIN_VITE_SENTRY_AUTH_TOKEN" >> $GITHUB_ENV
```
BYK, line 97: You know you can do the `env` definition at job level and avoid this silliness, right? RIGHT? 👀 🤣

packages/spotlight/vite.ui.config.ts
```diff
@@ -1,28 +1,49 @@
...
+
+  return {
+    plugins: [
+      ...reactPlugins,
+      sentryVitePlugin({
+        org: env.MAIN_VITE_SENTRY_ORG,
+        project: env.MAIN_VITE_UI_SENTRY_PROJECT,
+        authToken: env.MAIN_VITE_SENTRY_AUTH_TOKEN,
+        release: {
+          name: process.env.npm_package_version,
```
BYK: We should probably extract these common things out into `vite.config.base`

BYK (approved):

.github/workflows/build.yml
MathurAditya724: I tried that initially but it was not getting picked up by vite, not sure why, So I sticked with this ugly thing. I will try to improve this in another PR

PR #1173, opened 2025-12-02
Fix(Electron): Electron window dragging and traffic light positioning on macOS

betegon:
We had two issues  on macOS Electron app:

- Window not draggable
- The close/minimize/maximize buttons were positioned on top of the Spotlight logo in the sidebar

We were checking `IN_DESKTOP_ENV` which was never set anywhere, so the conditional `mt-8` margin were we set the draggable area and the semaphore was place wasn't applied.

I believe this was caused when we merged all packages into the spotlight one.

https://github.com/user-attachments/assets/ceedd707-7c14-471d-a824-1fe31024a528

### DIFF
--- a/packages/spotlight/src/ui/telemetry/components/TelemetryView.tsx
+++ b/packages/spotlight/src/ui/telemetry/components/TelemetryView.tsx
@@ -43,8 +43,8 @@ export default function TelemetryView({
   return (
     <div className="from-primary-900 to-primary-950 flex h-full overflow-hidden bg-gradient-to-br from-0% to-20% font-sans text-white">
       <TelemetrySidebar errorCount={errorCount} traceCount={traceCount} logCount={logCount} isOnline={isOnline} />
-      {/* pt-8 adds top padding in Electron to account for the 40px drag bar */}
-      <div className={cn("flex-1 overflow-auto", IS_ELECTRON && "pt-8")}>
+      {/* pt-10 (40px) adds top padding in Electron to account for the 40px drag bar */}
+      <div className={cn("flex-1 overflow-auto", IS_ELECTRON && "pt-10")}>
         <Routes>
           <Route path="not-found" element={<p>Not Found - How'd you manage to get here?</p>} key="not-found" />
           <Route path="traces/*" element={<TracesTab />} key="traces" />
--- a/packages/spotlight/src/ui/telemetry/components/TelemetrySidebar.tsx
+++ b/packages/spotlight/src/ui/telemetry/components/TelemetrySidebar.tsx
@@ -81,8 +81,8 @@ export default function TelemetrySidebar({ errorCount, traceCount, logCount, isO
       style={{ width: "240px", minWidth: "240px" }}
       aria-label="Navigation"
     >
-      {/* mt-8 adds top margin in Electron to account for the 40px drag bar */}
-      <header className={cn("p-4", IS_ELECTRON && "mt-8")}>
+      {/* mt-10 (40px) adds top margin in Electron to account for the 40px drag bar */}
+      <header className={cn("p-4", IS_ELECTRON && "mt-10")}>
         <div className="text-primary-200 flex flex-col gap-x-2">
           <div className="inline-flex items-center gap-x-2">
             <Logo height={24} width={24} />


.changeset/bumpy-mugs-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spotlightjs/spotlight": minor
```
BYK:
```suggestion
"@spotlightjs/spotlight": patch
```

packages/spotlight/src/electron/main/index.ts
```diff
@@ -175,39 +176,41 @@ const createWindow = () => {
...
+            </div>
+            <h1 class="error header">Oops! An error occurred.</h1>
+            <p class="error description">Press Cmd + R to reload the app.</p>
+          \`;
+          document.body.appendChild(errorScreen);
+        }
+        
+        // Re-create elements if body is replaced
+        new MutationObserver(() => {
+          if (!document.getElementById('electron-top-drag-bar')) {
```
BYK: Why would this happen? Needs docs.

packages/spotlight/src/ui/lib/Router.tsx
```diff
@@ -11,10 +12,9 @@ interface RouterProps {
  * - HashRouter for Electron apps (file:// protocol)
  */
 export function Router({ children }: RouterProps) {
-  // Detect if we're running in Electron (file:// protocol)
-  const isElectron = typeof window !== "undefined" && window.location.protocol === "file:";
+  const inElectron = isElectron();
 
-  if (isElectron) {
+  if (inElectron) {
     return <HashRouter>{children}</HashRouter>;
```
BYK, line 21: Do we still need this?

packages/spotlight/src/ui/telemetry/components/TelemetryView.tsx
```diff
@@ -39,12 +40,12 @@ export default function TelemetryView({
...
   const logCount = store.getLogs().length;
 
-  const isElectron = typeof process !== "undefined" && process.versions?.electron;
+  const inElectron = isElectron();
 
   return (
     <div className="from-primary-900 to-primary-950 flex h-full overflow-hidden bg-gradient-to-br from-0% to-20% font-sans text-white">
       <TelemetrySidebar errorCount={errorCount} traceCount={traceCount} logCount={logCount} isOnline={isOnline} />
-      <div className={cn("flex-1 overflow-auto", isElectron && "pt-8")}>
+      <div className={cn("flex-1 overflow-auto", inElectron && "pt-8")}>
```
BYK: I wonder why we need this (I mean the old code not your change)

packages/spotlight/src/bootstrap.ts
```diff
@@ -1,6 +1,16 @@
 // Bootstrap script to load the appropriate entry point based on the environment
-// With nodeIntegration: true, process is available in Electron renderer
-if (typeof process !== "undefined" && process.versions?.electron) {
+// Check multiple methods to detect if the app is running in Electron even on dev builds
+const isElectronEnv =
```
BYK: I'd say either this or the `packages/spotlight/src/ui/lib/isElectron.ts` helper needs to go as they are redundant. I think this one is better as it just runs once and we can set `IS_DESKTOP_ENV=1` instead of `__SPOTLIGHT_IS_ELECTRON__`.

Actually, even better, I think we can set this in Vite config.

packages/spotlight/src/bootstrap.ts
```diff
@@ -1,6 +1,9 @@
 // Bootstrap script to load the appropriate entry point based on the environment
-// With nodeIntegration: true, process is available in Electron renderer
-if (typeof process !== "undefined" && process.versions?.electron) {
+// __IS_ELECTRON__ is set to true at build time by vite.electron.config.ts
+
+declare const __IS_ELECTRON__: boolean | undefined;
```
MathurAditya724: Do we need this line? As we have already declared this in `vite-env.d.ts`

BYK: Cursor claimed this file is run before/outside of vite 🤷🏻

BYK: Removed 😀

MathurAditya724 (approved):

Issue #1171, opened 2025-12-02
CORS restrictions missing configuration for additional allowed origins


rassie:
## Problem

Commit d9ed848 (#1138) introduced CORS origin restrictions to the sidecar for security. However, the implementation hardcodes allowed origins with no way to configure additional ones.

The only allowed origins are:
- `localhost` / `127.0.0.1` / `[::1]` (any port/protocol)
- `https://spotlightjs.com` and subdomains (HTTPS only, default port)

## Broken Use Cases

This breaks legitimate development scenarios:

- **Custom local domains** - `myapp.local`, `dev.myapp.test` (common with /etc/hosts)
- **Docker containers** with custom hostnames
- **Tunneling services** - ngrok, localtunnel, Cloudflare tunnels (webhook testing, mobile dev)
- **VMs** with different hostnames
- **LAN access** from another device (testing on phone/tablet)
- **Corporate internal domains** like `dev.company.internal`
- **WSL** with custom networking

## Proposed Solution

Add an `allowedOrigins` option to `SideCarOptions`:

```typescript
export type SideCarOptions = {
  // ...existing options
  
  /**
   * Additional origins to allow for CORS requests.
   * Useful for custom local domains, tunnels, etc.
   */
  allowedOrigins?: string[];
};
```

The `isAllowedOrigin` function should check this list in addition to the hardcoded defaults.

## References

- Security commit: d9ed848e4239ef4d56ebdc21296e864a0f740cf9
- Original security issue: #1137

rassie:
> I'd not want to make people configure this if possible.

I'm absolutely fine with the current state being the default. I just need a way, however complex it may be, to add my local development domain to the list. And that domain is not built upon development TLDs unfortunately, for various reasons.

PR #1175, opened 2025-12-02
fix(sidecar): Allow any DNS pointing to localhost in CORS

BYK:
Refs #1171.

### DIFF
--- a/packages/spotlight/src/server/utils/cors.ts
+++ b/packages/spotlight/src/server/utils/cors.ts
@@ -325,6 +325,7 @@ export function getDnsCacheSize(): number {
  * Validates if an origin should be allowed to access the Sidecar.
  *
  * Allowed origins:
+ * - localhost (RFC 6761 reserved name, always trusted)
  * - Any origin whose hostname resolves to this machine's IPs (with TTL >= 1h)
  * - https://spotlightjs.com and subdomains (HTTPS only, default port)
  *
@@ -341,6 +342,13 @@ export async function isAllowedOrigin(origin: string): Promise<boolean> {
   try {
     const url = new URL(origin);
     const hostname = url.hostname.toLowerCase();
+
+    // Fast path: localhost is always allowed (RFC 6761 reserved name)
+    // This is safe because "localhost" is guaranteed to resolve to loopback
+    // and cannot be registered as a public domain name.
+    if (hostname === "localhost") {
+      return true;
+    }
 
     // Fast path: spotlightjs.com domains (no DNS lookup needed)
     if (isSpotlightOrigin(url, hostname)) {
--- a/packages/website/src/content/docs/docs/sidecar/index.mdx
+++ b/packages/website/src/content/docs/docs/sidecar/index.mdx
@@ -267,6 +267,7 @@ Run one sidecar and point all your services to it for unified debugging across y
 
 The sidecar is designed exclusively for local development. It only accepts connections from:
 
+- **localhost**: Always allowed (RFC 6761 reserved name)
 - **This machine's IPs**: Any hostname that resolves to an IP assigned to this machine, including:
   - Loopback addresses (`127.0.0.1`, `::1`)
   - Local network IPs (e.g., `192.168.x.x`, `10.x.x.x`)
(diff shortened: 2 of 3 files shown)


packages/spotlight/src/server/utils/cors.ts
```diff
@@ -1,15 +1,273 @@
...
+  let minTtl: number | undefined;
+
+  // Run all resolution methods in parallel for efficiency
+  const [resolve4Result, resolve6Result, lookupResult] = await Promise.allSettled([
+    resolve4(hostname, { ttl: true }),
+    resolve6(hostname, { ttl: true }),
+    lookup(hostname, { all: true }),
+  ]);
+
+  // Process IPv4 DNS results (with TTL)
```
BYK, line 170: The `Math.max()` along with `minTtl` logic in the entire file seems to be incorrect. We want to _reject_ DNS records with a TTL < MIN_TTL_MS as those records are unsafe. Forcing our own longer TTL sounds actually wrong.

That said we should probably make `MIN_TTL_MS` 1 hour

packages/spotlight/src/server/utils/cors.ts
```diff
@@ -1,15 +1,273 @@
...
+  machineIPsCache = null;
+  machineIPsCacheTime = 0;
+}
+
+/**
+ * Get the current cache size. Useful for testing and monitoring.
+ */
+export function getDnsCacheSize(): number {
+  return dnsCache.size;
+}
```
BYK, line 322: If this is not used anywhere, let's just remove it.

packages/spotlight/src/server/utils/cors.ts
```diff
@@ -1,15 +1,273 @@
...
+ *
+ * Threat Model:
+ * - An attacker controlling evil.com could try to rebind it to 127.0.0.1
+ * - Our 6h minimum cache means the attacker must keep the user engaged for 6+ hours
+ * - Even then, rebinding to 127.0.0.1 would make their site inaccessible
+ * - This makes DNS rebinding attacks impractical against Spotlight
+ *
+ * For machine IPs (non-loopback), the risk is even lower because:
+ * - External access requires explicit port forwarding/exposure
+ * - Most machines are behind NAT, blocking inbound connections by default
```
BYK: This comment seems to be a repetition from above. Just refer to the parts above this place instead of repeating content.

packages/spotlight/src/server/utils/cors.ts
```diff
@@ -18,18 +276,52 @@ export function isAllowedOrigin(origin: string): boolean {
...
+    // Check cache first
+    const cached = dnsCache.get(hostname);
+    if (cached && cached.expiresAt > Date.now()) {
+      return cached.isLocal;
     }
 
     // Check for spotlightjs.com domains - must be HTTPS with default port
-    if (url.protocol !== "https:" || (url.port !== "" && url.port !== "443")) {
+    // These are always allowed regardless of what they resolve to
+    if (hostname === "spotlightjs.com" || hostname.endsWith(".spotlightjs.com")) {
```
BYK: This looks like a cheaper check so why not move it earlier than the DNS look ups?

packages/spotlight/src/server/utils/cors.ts
```diff
@@ -18,18 +276,52 @@ export function isAllowedOrigin(origin: string): boolean {
...
+      });
+
+      return isLocal;
+    } catch {
+      // DNS resolution failed - cache the failure for a short time
+      // to avoid hammering DNS servers
+      dnsCache.set(hostname, {
+        isLocal: false,
+        expiresAt: Date.now() + FAILURE_TTL_MS,
+      });
```
BYK: Feels like this whole logic needs to be a helper function for clarity?

packages/website/src/content/docs/docs/sidecar/index.mdx
```diff
@@ -261,6 +261,65 @@ Run one sidecar and point all your services to it for unified debugging across y
...
+
+**How DNS rebinding works:**
+1. An attacker controls `evil.com` with a short DNS TTL
+2. User visits `evil.com`, which initially resolves to the attacker's server
+3. The attacker's page loads malicious JavaScript
+4. The attacker rebinds `evil.com` to `127.0.0.1` (or another machine IP)
+5. JavaScript makes requests to `evil.com` (now pointing to the machine)
+
+**Our mitigation:**
+- DNS resolution results are cached for a **minimum of 6 hours**
```
BYK: This part needs updating per my earlier comments (re reject any DNS record with a TTL less than 6 hours -- or 1 hour if you update it to be 1 hour)

PR #1179, opened 2025-12-03
docs(website):  CLI run documentation with Docker Compose auto-detection

betegon:


packages/website/src/content/docs/docs/cli/run.mdx
```diff
@@ -46,6 +46,47 @@ This makes it incredibly easy to get started - just add Spotlight to your existi
 }
 ```
 
+## Auto-Detection with Docker Compose
+
+Spotlight also automatically detects Docker Compose projects. When you run `spotlight run` without a command, it looks for these files in the current directory:
+
```
BYK, line 52: I think we also run when we detect that env var below. We should also detail "spotlight injection" a bit: we pass relevant en vars to services, add the host definition etc

BYK (approved):
🚀

packages/website/src/content/docs/docs/cli/run.mdx
betegon: i wrote that last bit but felt like too much info. I'll add it back but I as a user don't think i'd really care. 

i'd care if I have used other methods were i had to make that effort to make spotlight work but if  I just use spotlight run, i just want it to work 😆

betegon: check it out now @BYK

PR #1181, opened 2025-12-04
fix the npx thing

### LINKED ISSUE #1180: npx @spotlightjs/spotlight is broken
Run `npx @spotlightjs/spotlight`:

```
/home/byk/.npm/_npx/caa3a7cc76eeca4a/node_modules/.bin/spotlight: 1: Syntax error: ";" unexpected
```

MathurAditya724:
closes #1180

### DIFF
--- a/.changeset/fuzzy-years-admire.md
+++ b/.changeset/fuzzy-years-admire.md
@@ -2,4 +2,4 @@
 "@spotlightjs/spotlight": patch
 ---
 
-minor fixes
+Fix `npx @spotlightjs/spotlight` fail


.changeset/fuzzy-years-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spotlightjs/spotlight": patch
+---
+
+minor fixes
```
BYK:
```suggestion
Fix `npx @spotlightjs/spotlight` fail
```

packages/spotlight/vite.node.config.ts
```diff
@@ -1,15 +1,35 @@
 import { resolve } from "node:path";
 import { sentryVitePlugin } from "@sentry/vite-plugin";
+import type { Plugin } from "vite";
 import { defineConfig } from "vite";
 import { aliases, defineProduction, dtsPlugin, sentryPluginOptions } from "./vite.config.base";
 
+// Custom plugin to add shebang to the CLI entry point
+// This runs after all other transformations to ensure the shebang is the first line
+const shebangPlugin = (): Plugin => ({
```
BYK, line 9: Lol, so this was the missing bit?

packages/spotlight/vite.node.config.ts
```diff
@@ -1,15 +1,35 @@
...
+  name: "shebang",
+  apply: "build",
+  enforce: "post",
+  generateBundle(_options, bundle) {
+    for (const [fileName, chunk] of Object.entries(bundle)) {
+      if (fileName === "run.js" && chunk.type === "chunk") {
```
BYK, line 15: Just a concern, we also have `cli/run.js` would that be a problem?

packages/spotlight/vite.node.config.ts
MathurAditya724: Yup, as initially we were compiling using tsc, so this was added already

packages/spotlight/vite.node.config.ts
MathurAditya724: The npx command uses the `./src/run.ts` as described in the package.json. And this name comes from the vite config - `vite.node.config.ts`.

So this shouldn't be a problem

Issue #1197, opened 2025-12-09
Detect docker compose on spotlight run explicitly


betegon:
we should support `spotlight run docker compose up` and `docker-compose` too.

PR #1199, opened 2025-12-09
feat(CLI): Support docker compose explicitly

betegon:
solves #1197 .

follows a very similar approach as our current implementation for running docker-compose when it's automatically detected.

note: gotta revisit the code still

### DIFF
--- a/packages/spotlight/src/server/cli/run.ts
+++ b/packages/spotlight/src/server/cli/run.ts
@@ -12,7 +12,7 @@ import { logger } from "../logger.ts";
 import type { SentryLogEvent } from "../parser/types.ts";
 import type { CLIHandlerOptions } from "../types/cli.ts";
 import { detectDockerCompose, parseExplicitDockerCompose, prepareDockerComposeRun } from "../utils/docker-compose.ts";
-import { getSpotlightURL } from "../utils/extras.ts";
+import { getSpotlightURL, openInBrowser } from "../utils/extras.ts";
 import { EventContainer, getBuffer } from "../utils/index.ts";
 import tail, { type OnItemCallback } from "./tail.ts";
 
@@ -114,6 +114,7 @@ export default async function run({
   filesToServe,
   format,
   allowedOrigins,
+  open,
 }: CLIHandlerOptions) {
   let relayStdioAsLogs = true;
 
@@ -166,6 +167,10 @@ export default async function run({
   // or started in a weird manner (like over a unix socket)
   const actualServerPort = (serverInstance.address() as AddressInfo).port;
   const spotlightUrl = getSpotlightURL(actualServerPort, LOCALHOST_HOST);
+
+  if (open) {
+    openInBrowser(actualServerPort);
+  }
   let shell = false;
   let stdin: string | undefined = undefined;
   const env = {
--- a/.changeset/cli-open-flag.md
+++ b/.changeset/cli-open-flag.md
@@ -1,0 +1,5 @@
+---
+"@spotlightjs/spotlight": minor
+---
+
+Add `--open` / `-o` CLI flag to automatically open the Spotlight dashboard in your default browser when starting the sidecar
--- a/packages/spotlight/src/server/cli.ts
+++ b/packages/spotlight/src/server/cli.ts
@@ -44,6 +44,11 @@ const PARSE_ARGS_CONFIG = {
       type: "boolean",
       default: false,
     },
+    open: {
+      type: "boolean",
+      short: "o",
+      default: false,
+    },
     help: {
       type: "boolean",
       short: "h",
@@ -58,6 +63,7 @@ export type CLIArgs = {
   port: number;
   debug: boolean;
   help: boolean;
+  open: boolean;
   format: FormatterType;
   cmd: string | undefined;
   cmdArgs: string[];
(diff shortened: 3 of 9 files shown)


packages/spotlight/src/server/cli/run.ts
```diff
@@ -240,6 +244,25 @@ export default async function run({
...
       env.PATH = path.resolve("./node_modules/.bin") + path.delimiter + env.PATH;
     }
+  } else {
+    // Handle explicit docker compose commands (e.g., "spotlight run docker compose up")
+    const explicitDockerConfig = parseExplicitDockerComposeUp(cmdArgs);
+    if (explicitDockerConfig) {
+      logger.info(
+        `Detected explicit Docker Compose command with ${explicitDockerConfig.serviceNames.length} service(s): ${explicitDockerConfig.serviceNames.join(", ")}`,
+      );
+      metrics.count("cli.run.explicit_docker", 1, {
```
BYK:
```suggestion
      metrics.count("cli.run.docker_compose.explicit", 1, {
```

packages/spotlight/src/server/cli/run.ts
```diff
@@ -240,6 +244,25 @@ export default async function run({
...
+        attributes: { services: explicitDockerConfig.serviceNames.length },
+      });
+      // Use host.docker.internal for backend services to access the host machine
+      env.SENTRY_SPOTLIGHT = getSpotlightURL(actualServerPort, DOCKER_HOST);
+      const command = buildDockerComposeCommand(explicitDockerConfig);
+      cmdArgs = command.cmdArgs;
+      stdin = command.stdin;
+      // Always unset COMPOSE_FILE to avoid conflicts with explicit -f flags
+      // biome-ignore lint/performance/noDelete: need to remove env var entirely
+      delete env.COMPOSE_FILE;
```
BYK: Just like earlier, now this block of code seems to be repeated 3 times. I think this PR is the place to abstract this out.

packages/spotlight/src/server/utils/docker-compose.ts
```diff
@@ -151,44 +223,71 @@ export function detectDockerCompose(): DockerComposeConfig | null {
...
+    argsStartIndex = 1;
+  } else if (cmdArgs[0] === "docker" && cmdArgs[1] === "compose") {
+    command = ["docker", "compose"];
+    argsStartIndex = 2;
+  } else {
+    return null;
   }
 
-  const serviceNames = Array.from(serviceNamesSet);
+  const remainingArgs = cmdArgs.slice(argsStartIndex);
```
BYK: I'd say anything after this line is unnecessary. We don't care if it is the `up` command or `exec` command or `run` command. All we care is the user is running `doker compose` or `docker-compose` through Spotlight run and we should just inject 😀

packages/spotlight/src/server/utils/docker-compose.ts
```diff
@@ -151,44 +223,71 @@ export function detectDockerCompose(): DockerComposeConfig | null {
...
+  const upIndex = remainingArgs.indexOf("up");
+  if (upIndex === -1) {
+    logger.debug("Not a 'docker compose up' command, skipping Spotlight injection");
+    return null;
+  }
+
+  const argsBeforeUp = remainingArgs.slice(0, upIndex);
+  const userFiles = parseComposeFileFlags(argsBeforeUp);
```
BYK: Well this like makes sense for the correct config detection so we should keep it (contrary to my comment above)

BYK (approved):

packages/spotlight/src/server/utils/docker-compose.ts
```diff
@@ -116,13 +196,36 @@ export function buildDockerComposeCommand(config: DockerComposeConfig): {
...
+  config: DockerComposeConfig,
+  serverPort: number,
+  env: Record<string, string | undefined>,
+): { cmdArgs: string[]; stdin: string } {
+  env.SENTRY_SPOTLIGHT = `http://${DOCKER_HOST_INTERNAL}:${serverPort}/stream`;
+  const command = buildDockerComposeCommand(config);
+  // biome-ignore lint/performance/noDelete: need to remove env var entirely
+  delete env.COMPOSE_FILE;
+  return command;
+}
```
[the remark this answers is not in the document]
betegon: not really, frontend runs on browser so the url should stay localhost.

packages/spotlight/src/server/utils/docker-compose.ts
```diff
@@ -149,27 +170,59 @@ function collectServices(composeFiles: string[]): string[] {
...
+const FLAGS_WITH_VALUES = ["-f", "--file", "-p", "--project-name", "--project-directory", "--env-file", "--profile"];
+
 /**
- * Parse -f/--file flags from command arguments
+ * Extract -f/--file flags from global options (before the subcommand).
+ * Once we hit the first non-flag argument (the subcommand), everything passes through.
+ * This prevents confusing `logs -f` (follow) with `-f file.yml` (compose file).
  */
-function parseComposeFileFlags(args: string[]): string[] {
+function parseComposeFlags(args: string[]): { files: string[]; remainingArgs: string[] } {
```
BYK, line 183: You know what you can use here: https://nodejs.org/api/util.html#utilparseargsconfig

PR #1212, opened 2025-12-18
fix: opening external links in default browser

MathurAditya724:


packages/spotlight/src/electron/main/index.ts
```diff
@@ -166,6 +166,12 @@ const createWindow = () => {
     win = null;
   });
 
+  // Open external links in the default browser
+  win.webContents.setWindowOpenHandler(details => {
+    shell.openExternal(details.url);
+    return { action: "deny" };
+  });
```
betegon, line 175: we could send the failure to sentry:

```typescript
win.webContents.setWindowOpenHandler(details => {
  shell.openExternal(details.url).catch(error => {
    Sentry.captureException(error);
  });

  return { action: "deny" };
});
```

PR #1213, opened 2025-12-18
fix: updated the empty pages of traces and envelopes

MathurAditya724:
image

image

### DIFF
--- a/packages/spotlight/src/ui/telemetry/components/insights/envelopes/EnvelopeList.tsx
+++ b/packages/spotlight/src/ui/telemetry/components/insights/envelopes/EnvelopeList.tsx
@@ -80,24 +80,14 @@ export default function EnvelopeList() {
         </EmptyDescription>
       </EmptyHeader>
       <EmptyContent>
-        <div className="flex gap-4">
-          <a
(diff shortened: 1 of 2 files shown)


packages/spotlight/src/ui/telemetry/components/insights/envelopes/EnvelopeList.tsx
```diff
@@ -70,5 +71,34 @@ export default function EnvelopeList() {
...
+      </EmptyHeader>
+      <EmptyContent>
+        <div className="flex gap-4">
+          <a
+            href="https://docs.sentry.io/"
+            target="_blank"
+            rel="noopener noreferrer"
+            className="underline hover:text-primary-400"
+          >
+            Sentry Docs
```
betegon: i'd be pretty clear here, just one CTA to spotlight docs. we'll redirect you to sentry docs at some point in the spotlight docs if necesary, but having the two CTA could get people to click on sentry docs rather than the spotlight ones.

We should have docs good enough to point you to sentry if necessary.

betegon (changes requested):

PR #1222, opened 2025-12-24
feat: better error message display

MathurAditya724:
Preview of how this will look, this will be shown on both web and electron app

https://github.com/user-attachments/assets/34bab7f8-7776-49bd-ba66-ce77b5134190


betegon (commented):
this looks good, could you pls attach an image of how it looks like to de description?

PR #1227, opened 2026-01-02
docs: add section for the spotlight sdk

MathurAditya724:


packages/website/src/content/docs/docs/sdk/index.mdx
```diff
@@ -0,0 +1,222 @@
...
+## API Reference
+
+### `createSpotlightBuffer()`
+
+Creates a new message buffer to store incoming telemetry events.
+
+```typescript
+import { createSpotlightBuffer } from '@spotlightjs/spotlight/sdk';
+
+const buffer = createSpotlightBuffer();
```
betegon, line 35: how can user know the args for `createSpotlightBuffer`? like the buffer side you mention below. at least we could mention the name of that one in the text below.

PR #1320, opened 2026-06-10
Fix path stacking in sidebar and detail tabs (React Router v7 splat routes)

### LINKED ISSUE #1319: Sidebar navigation stacks URL path
### Environment

What version are you running? Etc.
4.11.5
### Steps to Reproduce

1. Open spotlight on the web
2. Click around the side navbar

### Expected Result

Page correctly navigates

### Actual Result

the URL path stacks on each other, breaking the navigation to other sections

zhiyan114:
Resolves #1319.

### Root cause

The sidebar links were relative (`<Link to="errors">`) and are rendered inside a **splat route** (`/telemetry/*`). They started stacking the URL path because the app silently moved from React Router **v6 to v7**:

- `packages/spotlight/package.json` declares `react-router-dom: ^6.30.2`, but the root `pnpm.overrides` entry `react-router-dom@>=6.0.0 <7.0.0: ">=6.30.4"` has an **unbounded value**, so pnpm rewrote the spec to `>=6.30.4` and resolved it to **7.17.0**.
- In v7 the `v7_relativeSplatPath` behavior is the default: a relative link/redirect rendered inside a splat route resolves against the **full current location, including the matched splat segment**. So from `/telemetry/traces/<id>` the sidebar's `to="errors"` resolved to `/telemetry/traces/<id>/errors` instead of `/telemetry/errors`.

The same mechanism affected more than the sidebar — the detail **tab bars** (`to="./context"`) and the fallback `<Navigate>` redirects stacked too (e.g. `/telemetry/traces/<id>/context/context`).

### Changes

Make the v7 install intentional and convert every splat-relative navigation to an absolute path:

- **Deps:** declare `react-router-dom: ^7.17.0` (declared == installed) and remove the dead/buggy `react-router`/`react-router-dom` overrides from the root `package.json`.
- **Sidebar** (`TelemetrySidebar.tsx`): absolute `/telemetry/<section>` links.
- **TelemetryTabs:** add an absolute `basePath` prop and pass it from the Trace, Span and Event detail tab bars (which render under dynamic splat routes).
- **Fallback redirects:** make the `<Navigate>` defaults absolute in `insights`, `TraceDetails` and `SpanDetails`.
- **Single source of truth:** add a module-level `TELEMETRY_BASE_URL` constant instead of repeating the `/telemetry` prefix across components.
- **Type safety:** `SpanDetails` now builds its base path from the route params (`traceId`/`spanId`, guaranteed for the matched route) instead of the span's *optional* `trace_id`, avoiding a possible `/telemetry/traces/undefined/...` URL.
- **Test:** `TelemetryTabs.test.tsx` asserts links stay absolute/stable with `basePath`, and explicitly demonstrates the stacking bug when relative `nested` links are used (so it would fail if the fix regressed).

### Verification

- `vitest run`: 188 passing (incl. the new regression tests).
- Typecheck: no new errors introduced.
- Manual: navigating the sidebar and detail tabs from deep URLs no longer stacks the path.


BYK:
Thanks for the fix, @zhiyan114! 🙏

I pushed a follow-up commit because the sidebar was one symptom of a wider issue. The actual root cause is that our root `pnpm.overrides` had an **unbounded** value (`react-router-dom@>=6.0.0 <7.0.0: ">=6.30.4"`), which silently resolved the declared `^6.30.2` all the way up to **react-router-dom 7.17.0**. Under v7, relative links rendered inside splat routes resolve against the full current location (including the splat segment), so the detail **tab bars** and a few fallback `<Navigate>` redirects were stacking the path too (e.g. `/telemetry/traces/<id>/context/context`).

The follow-up makes the v7 install intentional (declares `^7.17.0`, drops the dead overrides) and converts the remaining splat-relative navigation to absolute paths (`TelemetryTabs` `basePath`, the fallback redirects), plus a small regression test. Your absolute-path approach for the sidebar is kept as-is.

I've updated the PR description with the full details. 188 tests pass and typecheck is clean.

PR #1355, opened 2026-09-15
fix(deps): bump @microlink/react-json-view to fix invalid element type crash

rvanvelzen:
## Summary
- `JsonViewer` (used by trace/event context views, span DB query descriptions, envelope details, logs) can crash the whole app with React error #130 ("Element type is invalid... but got: object").
- Root cause: `@microlink/react-json-view@1.23.x`'s CJS/UMD bundle ends up double-wrapped as `{ default: Component }` under Vite's dependency interop (both `vite build` and the dev server's esbuild pre-bundling), so the default import resolves to an object instead of the component.
- The package fixed this themselves in `1.31.9` by adding a proper `index.mjs` ESM entry point and a conditional `exports` map (getsentry/spotlight is unaffected by this class of interop bug once resolving to that entry). This PR just bumps the dependency to pick up their fix; no application code changes needed.

## Test plan
- [x] Reproduced the crash live against real trace data using the previous pinned version (1.23.4 range, resolved 1.26.1) in both `vite --config vite.dev.config.ts` and the production `vite build --config vite.ui.config.ts` bundle.
- [x] After the bump (resolves to 1.31.34), rebuilt the production bundle and loaded 10+ real trace/context views (including large nested payloads) with no error #130 and no console errors from `JsonViewer`.
- [x] `tsc --noEmit` clean for the affected files; existing `AnsiText` unit tests in the same directory still pass.


BYK (approved):

BYK:
Thanks a lot!

